PR #2, opened 2017-07-20
mempool wait for the appearance of all transactions

Tolsi:


kushti:
Some comments in ErgoMemPool would be useful! At least, type signature in "waitedForAssembly: TrieMap[Seq[ModifierId], Promise[Seq[AnyoneCanSpendTransaction]]]" is non-trivial: why a seq is used as a key?

PR #71, opened 2017-10-27
use require instead of assert

### LINKED ISSUE #48: Check for correct usage of assert(), require(), enshuring()

Tolsi:
Closes #48


src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
```diff
@@ -195,7 +193,7 @@ object UtxoState {
           paranoidChecks = true
         ).get
 
-      assert(persistentProver.digest.sameElements(storage.version.get))
+      require(persistentProver.digest.sameElements(storage.version.get))
```
kushti: Why not "ensuring"? For post-conditions ensuring is a more natural choice.

Tolsi: ```def ensuring(cond: A => Boolean): A = { assert(cond(self)); self }```

plus this check is required or just for debug? I think that all our checks are mandatory by default, then we can change them to assert

PR #95, opened 2017-12-06
transactionsInBlock setting in testing{} section

kushti:


catena2w:
I propose to extract MiningSettings(mining: Boolean, offlineGeneration: Boolean, transactionsInBlock: Int)

PR #122, opened 2018-01-26
Switch to sync-rework scorex version

catena2w:


src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -70,35 +68,44 @@ abstract class ErgoNodeViewHolder[StateType <: ErgoState[StateType]](settings: E
...
-            log.info(s"State and History are inconsistent on startup. Applied missed modifier ${fb.encodedId}")
-            s
-          case Failure(e) =>
-            //TODO catch errors here
-            log.error(s"Failed to apply missed modifier ${fb.encodedId}. Try to resync from genesis", e)
-            ErgoApp.forceStopApplication(500)
-            state
+  private def restoreConsistentState(state: StateType, history: ErgoHistory): StateType = {
+    if (history.bestFullBlockIdOpt.isEmpty) {
+      state
```
kushti: So if no full blocks in history, state is consistent? What is the logic behind that?

PR #130, opened 2018-02-01
Sync refactoring

catena2w:


src/main/scala/org/ergoplatform/nodeView/history/MissedCache.scala
```diff
@@ -0,0 +1,7 @@
+package org.ergoplatform.nodeView.history
+
+
+
+class MissedCache {
+
```
daron666: Do we really need this one?

Issue #137, opened 2018-02-06
Call by name logger


catena2w:
May be copy-paste from https://github.com/wavesplatform/Waves/blob/master/src/main/scala/scorex/utils/ScorexLogging.scala#L9

daron666:
Or we can use https://github.com/typesafehub/scala-logging where all this code is implemented using macroses.

PR #143, opened 2018-02-07
Typesafe logger has been added

daron666:


daron666:
fixes #137

kushti:
We do have Typesafe Logger being used by Scorex, why do we need for this PR then?

PR #168, opened 2018-02-14
Fix type derivation for Difficulty. Make it implicit.

### LINKED ISSUE #159: Sometimes API returns float for difficulty
{'appVersion': '0.2.1', 'stateRoot': 'Po6gJZ5bYZ4amBUxGqt8xz2Sa89gKn1NcE2cqT5h6p9wD', 'peersCount': 3, 'headersHeight': 18316, 'unconfirmedCount': 0, 'bestHeaderId': 'Bzb3kH4V7VeRZErJB5Mp5kGmhxWQQCnY2saF4ZVBSZ9', 'bestFullHeaderId': '2cRWUtya6g41CDcXXhsfGPpeu7tmPQziVNUUj8oTgNj4', **'difficulty': 10.0**, 'fullHeight': 18172, 'isMining': False, 'stateType': 'utxo', 'name': 'UAj1rAf', 'stateVersion': 'QZNGARCoMLV73VfrvuidJFN7AcjnURCUvRJoMPebj1J', 'votes': 'UAj1rAf', 'previousFullHeaderId': '2ho4ZbqCyomNUSHAox7HgccPdk4caADneYVN8ieUPHkq'}

Not sure why this happens. Think we should fix that, difficulty is always represented by integer. Also, write unit tests for that.

daron666:
fixes #159


goral09:
Don't forget to remove https://github.com/ergoplatform/ergo/blob/9e592a888b0dcf1b0185a8fc0963e2bd8aae36cd/src/main/scala/org/ergoplatform/api/routes/InfoRoute.scala#L48

src/main/scala/org/ergoplatform/api/routes/InfoRoute.scala
```diff
@@ -73,4 +77,60 @@ object InfoRoute {
...
+                   stateType: String,
+                   stateVersion: Option[String],
+                   isMining: Boolean,
+                   votes: String,
+                   peersCount: Int
+                 )
+
+  object Info {
+
+    def make(nodeId: Array[Byte],
```
goral09: `def apply` will also work and will allow for nicer syntax. `Info(…)` instead of `Info.make(…)`

src/test/scala/org/ergoplatform/api/routes/InfoRoutesSpec.scala
```diff
@@ -22,7 +24,8 @@ class InfoRoutesSpec extends FlatSpec
   it should "return info" in {
     Get("/info") ~> route ~> check {
       status shouldBe StatusCodes.OK
-      InfoRoute.makeInfoJson(nodeId, minerInfo, connectedPeers.length, readers, "digest").toString shouldEqual responseAs[String]
+      val resp = responseAs[String]
+      Info.make(nodeId, minerInfo, connectedPeers.length, readers, "digest").asJson.toString shouldEqual resp
```
goral09: This doesn't verify that the fix has been made. It will verify that the same output was produced using two different functions which under the hood use the same machinery. 

Proper test would be to check that `difficulty` field is not a `Double` in any case. I would suggest to use `scalacheck` for generating the sample data and check for `.` existence.

daron666: Yeah. You are right. Will do it.

daron666: Also I'm thinking that now as we convert `Long` to `JSON`, and this conversion is determined, there is no need to check it. Cause it's circe responsibility to check this case.

goral09:
You're right. But it was also `circe` responsibility to not screw up when `getOrElse` was used and it did. Now it shouldn't be a problem since we're using `Long`. It's your call where to cut corners - I understand that writiing better test will take unproportionally more time.

Sent from [ProtonMail](https://protonmail.ch), encrypted email based in Switzerland.

-------- Original Message --------

src/main/scala/org/ergoplatform/api/routes/InfoRoute.scala
```diff
@@ -39,38 +43,65 @@ case class InfoRoute(readersHolder: ActorRef,
...
+      Info(
+        Algos.encode(nodeId),
+        Version.VersionString,
+        bestHeader.map(_.height).getOrElse(0),
+        bestFullBlock.map(_.header.height).getOrElse(0),
+        bestHeader.map(_.encodedId),
+        bestFullBlock.map(_.header.encodedId),
+        bestFullBlock.map(_.header.parentId).map(Base58.encode),
+        stateRoot,
+        bestFullBlock.map(_.header.requiredDifficulty).getOrElse(BigInt(0)).toLong,
```
catena2w: Is it safe to do ".toLong" here? Difficulty may be much bigger then long

daron666: I didn't know that. So how we can render this kind of numbers with circe? Maybe it will be better to expect all kinds of number in response? By default `BigInt(10).asJson` will be rendered as `1e1`.

goral09: rendering it as `String` ? it's only `info` endpoint - shouldn't be used as source of data for any important computations. and even if so - it's not difficult to parse it to `BigInt` on client side.

catena2w:
Fixed in https://github.com/ergoplatform/ergo/pull/170

PR #170, opened 2018-02-14
Correct node name in InfoRoute

catena2w:


src/main/scala/org/ergoplatform/api/routes/InfoRoute.scala
```diff
@@ -19,14 +19,18 @@ import scala.concurrent.Future
...
+                     nodeId: Array[Byte])
                     (implicit val context: ActorRefFactory) extends ErgoBaseApiRoute {
   override val route = info
 
+  override val settings: RESTApiSettings = ergoSettings.scorexSettings.restApi
+
   private def getConnectedPeers: Future[Int] = (peerManager ? PeerManager.GetConnectedPeers).mapTo[Seq[Handshake]].map(_.size)
 
-  private def getStateType: String = if (digest) "digest" else "utxo"
+  private def getStateType: String = if (ergoSettings.nodeSettings.ADState) "digest" else "utxo"
```
goral09: It's always better to encode (if possible) possible states in types.

catena2w: Created issue https://github.com/ergoplatform/ergo/issues/174

PR #184, opened 2018-02-19
Introduced StateType ADT. Resolves #174

### LINKED ISSUE #174: Rename boolen parameters in types
Instead of `ergo.node.ADState = false` param it's better have state types, like `ergo.node.stateType = "utxo"` and parse it to static type structure like
```
sealed trait StateType
case object UtxoState extends StateType
case object DigestState extends StateType
```
Check other similar parameters for such a refactoring
Don't forget to fix documentation.

mike-aksarin:
Introduced StateType ADT. Resolves #174


src/main/resources/node1/application.conf
```diff
@@ -5,7 +5,7 @@ ergo {
   # Settings for node view holder regime. See papers.yellow.ModifiersProcessing.md
   node {
     # Keep state root hash only and validate transactions via ADProofs
-    ADState = false
+    stateType = "utxo"
```
goral09: If the set of possible states known beforehand I would consider adding proper comment here. Otherwise people will find out about it only when app crashes.

src/main/scala/org/ergoplatform/api/routes/InfoRoute.scala
```diff
@@ -71,7 +72,7 @@ object InfoRoute {
       "stateRoot" -> stateRoot.asJson,
       "difficulty" -> bestFullBlock.map(_.header.requiredDifficulty).getOrElse(BigInt(0)).toString(10).asJson,
       "unconfirmedCount" -> unconfirmedCount.asJson,
-      "stateType" -> stateType.asJson,
+      "stateType" -> stateType.toString.asJson,
```
goral09: You have an `Encoder` for `StateType` so `stateType.asJson` should work?

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -142,11 +142,14 @@ private[nodeView] class UtxoErgoNodeViewHolder(settings: ErgoSettings, timeProvi
...
 
 object ErgoNodeViewHolder {
-  def createActor(system: ActorSystem, settings: ErgoSettings, timeProvider: NetworkTimeProvider): ActorRef = {
-    if (settings.nodeSettings.ADState) {
-      system.actorOf(Props.create(classOf[DigestErgoNodeViewHolder], settings, timeProvider))
-    } else {
-      system.actorOf(Props.create(classOf[UtxoErgoNodeViewHolder], settings, timeProvider))
+
+  def viewHolderClass(settings: ErgoSettings): Class[_ <: ErgoNodeViewHolder[_]] =
+    settings.nodeSettings.stateType match  {
```
goral09: This looks _hacky_ . Cleaner approach would be to define a type class
```
trait StateNodePropsProvider[S <: StateType] { def props[S]: Props } 

implicit object DigestProps extends StateNodePropsProvider[Digest.type] { 
  override def props[Digest.type] : Props = … 
}

def viewHolderProps[S <: Statetype](st: settings.nodeSettings.stateType)(implicit propsProvider: StateNodePropsProvider[S]) = propsProvider.props(st)
…
```
disclaimer: I didn't compile the code ;)

kushti: Better to follow this approach already being used in Scorex https://github.com/ScorexFoundation/Scorex/issues/162 ( also it is worth to create Ergo Wiki pages in this github on code conventions and desirable patterns to use).

goral09: I implemented this approach in my first PR. This won't change the implementation - it will just move the code to different place.

src/main/scala/org/ergoplatform/api/routes/InfoRoute.scala
mike-aksarin: yes, this would work, you're really sharp. I left `toString` here just for the better safety but `circe` does this job good enough.

src/main/resources/node1/application.conf
mike-aksarin: good idea!

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
mike-aksarin: You're right, at least exposing `viewHolderClass` method was hacky here. I rewrote this code to adopt  `<SomeActor>Ref.apply` approach from Scorex. See: https://github.com/ScorexFoundation/Scorex/pull/169/files 

Please also note that we are not going to expose these `UtxoErgoNodeViewHolder` and `DigestErgoNodeViewHolder` classes here as they are package-private. `ErgoNodeViewHolderRef` should be use instead to create props or reference to both types of ErgoNodeViewHolders.

goral09: This looks much better :+1:  but still allows for sth like this:
```
case StateType.Digest => Props.create(classOf[UtxoErgoNodeViewHolder], settings, timeProvider)
```

anyway, you have to make a _leap of faith_ somewhere - although I would argue that the later the better. I understand that you have to deliver. Anyway, this is much better :+1:

mike-aksarin: Ah, now I got your idea! I've just added some type safety for the `NodeViewHolder` instantiation, though the code become a bit more hard to comprehend. Though I'm still in doubts which of the two variants was better

goral09: @mike-aksarin I will have a look and get back to you. It's normal that the code that is more type-safe becomes less straightforward to comprehend at first. This is matter of practice and hiding unnecessary complexity.

goral09: @mike-aksarin I removed the code you added that put type constraints and I removed `ErgoNodeViewProps` entirely like that : 
```

/** This class guarantees to its inheritors the creation of correct instance of [[ErgoNodeViewHolder]]
  *  for the given instance of [[StateType]]
  */
object DigestNodeViewProps {
  def apply(settings: ErgoSettings, timeProvider: NetworkTimeProvider, digestType: StateType.DigestType): Props =
    Props.create(classOf[DigestNodeViewHolder], settings, timeProvider)
}

object UtxoNodeViewProps {
  def apply(settings: ErgoSettings, timeProvider: NetworkTimeProvider, digestType: StateType.UtxoType): Props =
    Props.create(classOf[UtxoNodeViewHolder], settings, timeProvider)
}
```

and it still compiles so this approach doesn't give you any benefit.

PR #185, opened 2018-02-19
Clean sbt cache before assembly

andyceo:


goral09:
Before calling `assembly` I would make double check that it's also `reload` and `update`.

kushti:
@goral09 why these commands are needed?

goral09:
@kushti it `reload`s the project definition in `build.sbt` and `update` s the dependencies.

kushti:
@goral09 we're using sbt-lock plugin so dependency versions are locked.

kushti:
@andyceo please add "reload"

andyceo:
Done.

PR #217, opened 2018-03-14
Bench tuning

daron666:


catena2w:
Why do we need to change `src/test/resources/bench/modifiers.txt? It's 21 Mb file that makes our git very heavy..

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -190,12 +190,13 @@ trait HeadersProcessor extends ScorexLogging {
       Failure(new Error("Parent header is marked as semantically invalid"))
     } else {
       Success()
-    }.recoverWith { case thr =>
```
catena2w: What is the purpose of this change?

daron666: I removed it, cause this change is out of scope of current task. I think that this `recoverWith` will be only applied to `Success` and hence never will be called.

Here is link to small example that demonstrates that `recover` in `validate` won't work, rather than in `validateF` [Scastie](https://scastie.scala-lang.org/uFFsJoxfQqu6isDhZinrjg)

PR #226, opened 2018-03-22
Shutdown hook has been added to prevent corrupted state after restart

### LINKED ISSUE #155: Application shutdown
Check that node shutdown correctly. For now it throws an Exception, not sure that it closes in consistent state

daron666:
fixes #155

### DIFF
--- a/src/main/scala/org/ergoplatform/ErgoApp.scala
+++ b/src/main/scala/org/ergoplatform/ErgoApp.scala
@@ -1,6 +1,6 @@
 package org.ergoplatform
 
-import akka.actor.ActorRef
+import akka.actor.{ActorRef, ActorSystem, PoisonPill}
 import org.ergoplatform.api.routes._
 import org.ergoplatform.local.ErgoMiner.StartMining
 import org.ergoplatform.local.TransactionGenerator.StartGeneration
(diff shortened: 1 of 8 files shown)


src/main/scala/org/ergoplatform/ErgoApp.scala
```diff
@@ -65,15 +67,30 @@ class ErgoApp(args: Seq[String]) extends Application {
...
 
+  val actorsToStop = Seq(minerRef,
+    peerManagerRef,
+    networkControllerRef,
+    readersHolderRef,
+    nodeViewSynchronizer,
+    localInterface,
+    nodeViewHolderRef
+  )
+  sys.addShutdownHook(ErgoApp.shutdown(actorSystem, actorsToStop))
```
mike-aksarin: I think we also need to ensure that shutdown procedure is correct for the node terminations due to unhandled exceptions, e.g. in Actor threads. Coordinated shutdown occurs in this case, but I'm not sure that we close all the actors gracefully.

mike-aksarin:
Looks great! Maybe we could  also ensure that shutdown procedure is correct as well for the node terminations due to *unhandled exceptions*, e.g. in Actor threads? Coordinated shutdown occurs in this case, but I'm not sure that we close all the actors gracefully.

PR #246, opened 2018-04-18
Fix OpenAPI spec and methods

andyceo:
- Fixed OpenAPI method /blocks/{headerId}/transactions: now method return BlockTransactions object (was returning list of transactions)
- Numerous fixes for OpenAPI routes (see commits)


src/main/resources/api/openapi.yaml
```diff
@@ -194,66 +227,86 @@ components:
...
           minimum: 0
           example: 667
+          nullable: true
         bestFullHeaderId:
-          $ref: '#/components/schemas/ModifierId'
+          type: string
+          description: Can be 'null' if no full block is applied since node launch
+          nullable: true
+          allOf:
+            - $ref: '#/components/schemas/ModifierId'
```
mike-aksarin: I guess we can introduce a new type like `ModifierIdOption`, which is nullable, and use this type instead of referencing `allOf` ModifierId in all these places

andyceo: Not sure this is good for this particular case, because we can not use any sibling elements with $ref. See https://swagger.io/docs/specification/using-ref/

So, if we need unique descriptions here, we need to repeat property definitions too.

PR #261, opened 2018-04-27
Fix mining specification

catena2w:


src/main/scala/org/ergoplatform/api/routes/TransactionsApiRoute.scala
```diff
@@ -26,10 +26,10 @@ case class TransactionsApiRoute(readersHolder: ActorRef, nodeViewActorRef: Actor
 
   override val settings: RESTApiSettings = restApiSettings
 
-  private def getMemPool: Future[Option[ErgoMemPoolReader]] = (readersHolder ? GetReaders).mapTo[Readers].map(_.m)
+  private def getMemPool: Future[ErgoMemPoolReader] = (readersHolder ? GetReaders).mapTo[Readers].map(_.m)
```
mike-aksarin: looks like we should wait the whole timeout if MemPool was not ready yet at the moment of request.

src/main/scala/org/ergoplatform/nodeView/ErgoReadersHolder.scala
```diff
@@ -32,7 +32,10 @@ class ErgoReadersHolder(viewHolderRef: ActorRef) extends Actor with ScorexLoggin
       mempoolReaderOpt = Some(reader)
 
     case GetReaders =>
-      sender ! Readers(historyReaderOpt, stateReaderOpt, mempoolReaderOpt)
+      (historyReaderOpt, stateReaderOpt, mempoolReaderOpt) match {
+        case (Some(h), Some(s), Some(m)) => sender ! Readers(h, s, m)
+        case _ =>
```
mike-aksarin: if we use `?` pattern here to get the readers, then the caller will wait the whole timeout, right?

src/main/scala/org/ergoplatform/nodeView/ErgoReadersHolder.scala
```diff
@@ -49,7 +52,7 @@ object ErgoReadersHolder {
 
   case object GetReaders
 
-  case class Readers(h: Option[ErgoHistoryReader], s: Option[ErgoStateReader], m: Option[ErgoMemPoolReader])
+  case class Readers(h: ErgoHistoryReader, s: ErgoStateReader, m: ErgoMemPoolReader)
```
mike-aksarin: this is nice improvement, i like it

src/main/scala/org/ergoplatform/nodeView/ErgoReadersHolder.scala
mike-aksarin: Maybe we need a comment here that we never have Nones here after successful node start

PR #277, opened 2018-05-16
Api cleanup

mike-aksarin:
Use scorex ApiResponse facility and drop the alternative implementation of the same functionality in ergo. 

This also includes #274 transaction POST


kushti:
@mike-aksarin please update or close

PR #283, opened 2018-05-23
Testnet1

catena2w:
Breaking changes for testnet 1


src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -68,7 +67,7 @@ class ErgoMiner(ergoSettings: ErgoSettings,
       miningThreads.foreach(_ ! candidateOpt.get)
     case StartMining if candidateOpt.isEmpty =>
       requestCandidate
-      context.system.scheduler.scheduleOnce(1.seconds, self, StartMining)
+      context.system.scheduler.scheduleOnce(1.seconds, self, StartMining)(context.system.dispatcher)
```
mike-aksarin: could be shortened to `context.dispatcher`.

same for below calls

PR #307, opened 2018-06-15
Yello Paper Updates

kushti:


catena2w:
We discussed `R1` for colored coins, why did you decided to change it to `R3`?

catena2w:
will be fixed in a separate PR for issue https://github.com/ergoplatform/ergo/issues/364

PR #310, opened 2018-06-19
Intelligent modifiers cache

kushti:


src/main/scala/org/ergoplatform/nodeView/ErgoModifiersCache.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  override def findCandidateKey(history: ErgoHistory): Option[K] = {
+    val headersHeight = history.headersHeight
+
+    val maxIterations = 10
+
+    @tailrec
+    def cacheIteration(called: Int): Option[K] = {
+      val expectedFullBlockParts = history
+        .headerIdsAtHeight(history.fullBlockHeight + 1)
```
catena2w: What if we have a fork? E.g current chain is `A1  ->  A2` (height 2)  and we have blocks B2 and B3 in cache. Looks like we'll never try to apply block B2, because it's height is equals to history.fullBlockHeight. 
A1  ->  A2
      \-> B2-> B3

src/test/scala/org/ergoplatform/nodeView/ErgoModifiersCacheSpecification.scala
```diff
@@ -0,0 +1,99 @@
+package org.ergoplatform.nodeView
+
+import org.ergoplatform.modifiers.history.{ADProofs, BlockTransactions, Header}
+import org.ergoplatform.nodeView.state.StateType
+import org.ergoplatform.utils.{ErgoPropertyTest, HistorySpecification}
+import scorex.crypto.hash.Blake2b256
+
+import scala.collection.mutable
+
+class ErgoModifiersCacheSpecification extends ErgoPropertyTest with HistorySpecification {
```
catena2w: Please write a test for situation, when there are different forks.

build.sbt
```diff
@@ -14,7 +14,7 @@ resolvers ++= Seq("Sonatype Releases" at "https://oss.sonatype.org/content/repos
   "Typesafe maven releases" at "http://repo.typesafe.com/typesafe/maven-releases/",
   "Sonatype Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots/")
 
-val scorexVersion = "effd499c-SNAPSHOT"
+val scorexVersion = "b2835a7b-SNAPSHOT"
```
catena2w: Update Scorex to latest master

Issue #328, opened 2018-06-28
bestFullBlock and bestHeader are not linearized 


kushti:
Potentially bestFullBlock* and bestHeader* variables in ErgoHistory could reference different chains(due to forks), which is counter-intuitive and causing problems in ErgoHistory clients (e.g. nextModifiersToDownload in ToDownloadsProcessor and findCandidate in ErgoModifiersCache, but we need to carefully check all the places where bestFullBlock.header is handled to be sure it is not considered as a part of a best-headers chain).

Possible solutions are: 
1) include "commonBlock" variable to the ErgoHistory interface, which shows a common block of two potentially different chains(one with most work, another with most work and full blocks only). Then client applications can check whether forking happens (by checking commonBlock == fullBlock). 

2) make the only one best chain rule, then bestFullBlock and bestHeader will be linearized. Switching to a better chain will happen only if it contains enough full blocks (full blocks in the better chain have more cumulative weight than the full block of the current chain).

catena2w:
We decided to add `bestLinearHeader` link, implemented at https://github.com/ergoplatform/ergo/pull/333

catena2w:
duplicate of https://github.com/ergoplatform/ergo/issues/331

PR #330, opened 2018-06-28
Common block in /info API route

catena2w:
Variant 1 fix of https://github.com/ergoplatform/ergo/issues/328

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
@@ -70,13 +70,11 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
     * Common block between bestHeader and bestFullBlock
     */
   def commonBlockIdOpt: Option[ModifierId] = {
-    bestFullBlockOpt match {
-      case Some(fb) =>
+    bestFullBlockOpt.flatMap { fb =>
         headerChainBack(config.blocksToKeep,
           fb.header,
           h => headerIdsAtHeight(h.height).headOption.forall(_ sameElements h.id))
           .headOption.map(_.id)
-      case None => None
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -66,6 +66,20 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
     */
   def headersHeight: Int = bestHeaderIdOpt.flatMap(id => heightOf(id)).getOrElse(-1)
 
+  /**
+    * Common block between bestHeader and bestFullBlock
+    */
+  def commonBlockIdOpt: Option[ModifierId] = {
+    bestFullBlockOpt match {
```
mike-aksarin: maybe `bestFullBlockOpt flatMap { fb =>` will look better here?

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -66,6 +66,18 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
     */
   def headersHeight: Int = bestHeaderIdOpt.flatMap(id => heightOf(id)).getOrElse(-1)
 
+  /**
+    * Common block between bestHeader and bestFullBlock
+    */
+  def commonBlockIdOpt: Option[ModifierId] = {
+    bestFullBlockOpt.flatMap { fb =>
+        headerChainBack(config.blocksToKeep,
```
kushti: Why this limit? A common header could be deeper than that.
An independent question is what could be lowest limit on config.blocksToKeep (I propose to set it to be config.keepVersions at least).

kushti: commonBlockIdOpt implementation is utterly inefficient. For every call a lot of disk reads will be done potentially. Isn't it better to store and update a variable storing the reference to the common header?

catena2w:
should be useless after https://github.com/ergoplatform/ergo/pull/333 merge

PR #336, opened 2018-07-03
Testnet 1.5

kushti:


.gitignore
```diff
@@ -16,10 +16,12 @@ target
 
 # local application.conf
 local-application.conf
-my-application.conf
+*-local.conf
+my*.conf
 
 # logs
 *.log
+*.log.gz
```
catena2w: May be it's better to ignore all `.gz` files to avoid binary files in git repo

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -15,25 +18,46 @@ import sigmastate.serialization.{Serializer => SSerializer}
...
                            override val outputCandidates: IndexedSeq[ErgoBoxCandidate])
   extends Transaction with ErgoLikeTransactionTemplate[Input] with MempoolModifier
-     with ModifierValidator with ScorexLogging {
+    with ModifierValidator with ScorexLogging {
 
   override type IdType = ModifierId
 
   override lazy val id: ModifierId = ModifierId @@ Blake2b256.hash(messageToSign)
 
+  private def fillAssetsMap(boxes: IndexedSeq[ErgoBoxCandidate],
```
catena2w: Please provide comments to this class methods

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -15,25 +18,46 @@ import sigmastate.serialization.{Serializer => SSerializer}
...
+                            map: mutable.Map[ByteArrayWrapper, Long]) = Try {
+    boxes.foreach { box =>
+      require(box.additionalTokens.size <= ErgoBox.MaxTokens, "Output contains too many assets")
+      box.additionalTokens.foreach { case (assetId, amount) =>
+        require(amount >= 0, s"negative asset amount for ${Base16.encode(assetId)}")
```
catena2w: Do not use `Base16` directly, use `Algos.encode` or `encoder.encode`

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -46,48 +70,69 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
       }
       accCost + scriptCost
     }
-    txCost
-  }
+
+    lazy val inputSum = boxesToSpend.map(_.value).reduce(Math.addExact(_, _))
+    lazy val outputSum = outputCandidates.map(_.value).reduce(Math.addExact(_, _))
+
+    def checkAssetPreservationRules = {
```
catena2w: May be better to have `Try` return type here

kushti: @mike-aksarin is developing a wrapper in Validator for Try() and an arbitrary code block (to be wrapped in Try inside)

PR #343, opened 2018-07-05
Do not request modifiers, that where already delivered. fix #288

### LINKED ISSUE #288: Received modifier that is already in history
During initial sync there are a lot of messages like
```
Received modifier GmuJExuGbMcMouKAsgSrGBavGXauefqaBTnjhKr46afo that is already in history
```

catena2w:


src/main/scala/org/ergoplatform/network/ErgoDeliveryTracker.scala
```diff
@@ -27,9 +27,11 @@ class ErgoDeliveryTracker(system: ActorSystem,
   def isExpecting: Boolean = expecting.nonEmpty
 
   /**
-    * @return ids we're going to download
+    * TODO temporary fiz until https://github.com/ergoplatform/ergo/issues/342 will be implemented
```
kushti: fiz => fix

kushti:
@catena2w approved, but please made fixes according to my comments. Also , this fix is not complete, as elements from "delivered" are being deleted on regular basis (so the updated  check only works before a scheduled call to clean the "delivered" set).

PR #347, opened 2018-07-06
Fix ErgoDeliveryTracker behavior

catena2w:


src/main/scala/org/ergoplatform/network/ErgoDeliveryTracker.scala
```diff
@@ -30,8 +30,8 @@ class ErgoDeliveryTracker(system: ActorSystem,
     * TODO temporary fix until https://github.com/ergoplatform/ergo/issues/342 will be implemented
     * @return ids we're going to download or already have downloaded
     */
-  def expectingAndDelivered: Iterable[ModifierId] =
-    ModifierId @@ delivered.keys.map(_.array) ++ ModifierId @@ expectingFromRandom.keys.map(_.array)
+  def expectingAndDelivered: (Iterable[ModifierId], Iterable[ModifierId]) =
```
mike-aksarin: we probably should rename this to `deliveredAndExpected`

PR #348, opened 2018-07-06
Status checker

catena2w:


catena2w:
Will be implemented in scorex

Issue #311, opened 2018-06-19
Fix TransactionGenerator


kushti:

catena2w:
fixed in https://github.com/ergoplatform/ergo/pull/461

Issue #315, opened 2018-06-20
Implement wallet


catena2w:
May be it's better to implement custom tokens first (https://github.com/ergoplatform/ergo/issues/314).
Required for https://github.com/ergoplatform/ergo/issues/311

catena2w:
Done in https://github.com/ergoplatform/ergo/pull/422

PR #362, opened 2018-07-16
Wallet prototype implementation

kushti:


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  private def extractFromBlock(fb: ErgoFullBlock): Int = {
+    height = fb.header.height
+    lastBlockUtxoRootHash = fb.header.stateRoot
+    fb.transactions.count(tx => scan(tx, Some(height)))
+  }
+
+  override def receive: Receive = {
```
mike-aksarin: cyclomatic complexity is reported here. I think it worth splitting the method to several handlers as we usually do

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,197 @@
...
+        context.system.scheduler.scheduleOnce(10.seconds)(self ! Resolve)
+      }
+
+    case ScanOnchain(fullBlock) =>
+      val txsFound = extractFromBlock(fullBlock)
+      (1 to txsFound).foreach(_ => self ! Resolve)
+
+    //todo: update utxo root hash
+    case Rollback(heightTo) =>
+      height.until(heightTo, -1).foreach{h =>
```
mike-aksarin: formatting issue at  `foreach{h` – spaces are missing

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,197 @@
...
+        val toRemove = Registry.confirmedAt(h)
+        toRemove.foreach{boxId =>
```
mike-aksarin: formating issue `foreach{boxId`

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,197 @@
...
+          Registry.removeFromRegistry(boxId).foreach{tb =>
```
mike-aksarin: formatting issue `foreach{tb`

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  def removeFromRegistry(boxId: ByteArrayWrapper) = registry.remove(boxId)
+
+  private val confirmedIndex = mutable.TreeMap[Height, Seq[ByteArrayWrapper]]()
+
+  def putToConfirmedIndex(height: Height, boxId: ByteArrayWrapper) = synchronized {
+    confirmedIndex.put(height, confirmedIndex.getOrElse(height, Seq.empty) :+ boxId)
+  }
+
+  def confirmedAt(height: Height): Seq[ByteArrayWrapper] = confirmedIndex.getOrElse(height, Seq.empty)
```
mike-aksarin: we should synchronize here as well, or make `confirmedIndex ` a concurrent data structure like `TrieMap`

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  private var _confirmedBalance: Long = 0
```
mike-aksarin: this should be `volatile`, or atomic, or all reads should be `synchronized`  as well as writes

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
...
+  private val _confirmedAssetBalances: mutable.Map[ByteArrayWrapper, Long] = mutable.Map()
```
mike-aksarin: we need a concurrent data structure here, or all reads should be synchronized as well as writes

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  private var _unconfirmedBalance: Long = 0
+  private val _unconfirmedAssetBalances: mutable.Map[ByteArrayWrapper, Long] = mutable.Map()
```
mike-aksarin: we need a concurrent data structure here, or all reads should be synchronized as well as writes

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
...
+  def putToConfirmedIndex(height: Height, boxId: ByteArrayWrapper) = synchronized {
+    confirmedIndex.put(height, confirmedIndex.getOrElse(height, Seq.empty) :+ boxId)
+  }
+
+  def confirmedAt(height: Height): Seq[ByteArrayWrapper] = confirmedIndex.getOrElse(height, Seq.empty)
+
+  private var _confirmedBalance: Long = 0
+  private val _confirmedAssetBalances: mutable.Map[ByteArrayWrapper, Long] = mutable.Map()
+
+  private var _unconfirmedBalance: Long = 0
```
mike-aksarin: this should be `volatile`, or atomic, or all reads should be `synchronized` as well as writes

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
...
+  private val _unconfirmedAssetBalances: mutable.Map[ByteArrayWrapper, Long] = mutable.Map()
+
+  def confirmedBalance = _confirmedBalance
+  def confirmedAssetBalances = _confirmedAssetBalances
```
mike-aksarin: we shouldn't give away a mutable data structure I guess

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  def unconfirmedBalance = _unconfirmedBalance
+  def unconfirmedAssetBalances = _unconfirmedAssetBalances
```
mike-aksarin: same as above: we probably shouldn't give away a mutable data structure

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,131 @@
+package org.ergoplatform.nodeView.wallet
+
+import io.iohk.iodb.ByteArrayWrapper
+import org.ergoplatform.nodeView.history.ErgoHistory.Height
+
+import scala.collection.mutable
+
+object Registry {
+
+  private val registry = mutable.Map[ByteArrayWrapper, TrackedBox]()
```
mike-aksarin: same as bellow: we need a concurrent data structure here, or all reads should be synchronized as well as writes

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  private val confirmedIndex = mutable.TreeMap[Height, Seq[ByteArrayWrapper]]()
```
mike-aksarin: here as well: we need a concurrent data structure, or all reads and writes should be synchronized

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
```diff
@@ -0,0 +1,213 @@
...
+import scala.util.Try
+
+
+/**
+  * An address is a short string corresponding to some script used to protect a box. Unlike (string-encoded) binary
+  * representation of a script, an address has some useful characteristics:
+  *
+  * * Integrity of an address could be checked., as it is incorporating a checksum.
+  * * A prefix of address is showing network and an address type.
+  * * An address is using an encoding (namely, Base58) which is avoiding similarly l0Oking characters, friendly to
```
mike-aksarin: do we really use base58 still?

also are the double ** stars – something intended?

src/main/scala/org/ergoplatform/nodeView/wallet/Registry.scala
```diff
@@ -0,0 +1,138 @@
+package org.ergoplatform.nodeView.wallet
+
+import io.iohk.iodb.ByteArrayWrapper
+import org.ergoplatform.nodeView.history.ErgoHistory.Height
+
+import scala.collection.mutable
+
+class Registry {
```
mike-aksarin: We can remove all the synchronisation from Registry if we plan to have it only as private Actor field.

Otherwise we need to provide additional synchronisation for mutable field reads, not only for writes.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,202 @@
...
+    case ScanOnchain(fullBlock) =>
+      val txsFound = extractFromBlock(fullBlock)
+      (1 to txsFound).foreach(_ => self ! Resolve)
+
+    //todo: update utxo root hash
+    case Rollback(heightTo) =>
+      height.until(heightTo, -1).foreach { h =>
+        val toRemove = registry.confirmedAt(h)
+        toRemove.foreach { boxId =>
+          registry.removeFromRegistry(boxId).foreach { tb =>
```
mike-aksarin: We probably need to call also  `box.deregister(registry)` here to decrease the balances from registered boxes

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,202 @@
...
+
+    case Resolve =>
+      resolveUncertainty()
+      //todo: avoid magic number, use non-default executor? check that resolve is not scheduled already
+      if (registry.uncertainBoxes.nonEmpty) {
+        context.system.scheduler.scheduleOnce(10.seconds)(self ! Resolve)
+      }
+
+    case ScanOnchain(fullBlock) =>
+      val txsFound = extractFromBlock(fullBlock)
```
mike-aksarin: This counts transactions, containing watched boxes, but not the boxes themselves, which could be more numerous. Maybe this is ok

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,202 @@
...
+  private def extractFromBlock(fb: ErgoFullBlock): Int = {
+    height = fb.header.height
+    lastBlockUtxoRootHash = fb.header.stateRoot
+    fb.transactions.count(tx => scan(tx, Some(height)))
+  }
+
+  def scanLogic: Receive = {
+    case ScanOffchain(tx) =>
+      if (scan(tx, None)) {
+        self ! Resolve
```
mike-aksarin: This sends only one  `Resolve` per transaction, though there could be several watched boxes in it. Maybe this is ok.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+    case Resolve =>
+      resolveUncertainty()
+      //todo: avoid magic number, use non-default executor? check that resolve is not scheduled already
+      if (registry.uncertainBoxes.nonEmpty) {
+        context.system.scheduler.scheduleOnce(10.seconds)(self ! Resolve)
```
mike-aksarin: Here we send twice more `Resolve` messages than was sent after scanning transactions

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -0,0 +1,213 @@
...
+import scala.collection.Map
+import scala.collection.mutable
+import scala.util.{Failure, Random, Success}
+import scala.concurrent.ExecutionContext.Implicits.global
+
+
+case class BalancesSnapshot(height: Height, balance: Long, assetBalances: Map[ByteArrayWrapper, Long])
+
+
+class ErgoWalletActor(settings: ErgoSettings) extends Actor with ScorexLogging {
```
catena2w: Please add comments, what are the goals of this class. 
Why do we need separate actor for Wallet?

catena2w:
closes https://github.com/ergoplatform/ergo/issues/311 and https://github.com/ergoplatform/ergo/issues/315

src/main/scala/org/ergoplatform/local/TransactionGenerator.scala
```diff
@@ -1,77 +1,108 @@
...
+
+  private var transactionsPerBlock = 0
+  private var currentFullHeight = 0
 
-  @SuppressWarnings(Array("TraversableHead"))
   override def receive: Receive = {
     case StartGeneration =>
       if (!isStarted) {
-        context.system.scheduler.schedule(1500.millis, 1500.millis)(self ! FetchBoxes)(context.system.dispatcher)
+        viewHolder ! GetDataFromCurrentView[ErgoHistory, UtxoState, ErgoWallet, ErgoMemPool, Unit] { v =>
```
catena2w: Please add logs that generation was started, etc.

src/main/resources/api/openapi.yaml
```diff
@@ -1030,3 +1076,140 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/BalancesSnapshot'
+        default:
+          description: Error
+          content:
+            text/plain; charset=UTF-8:
+              schema:
+                type: string
+
+  /wallet/transaction:
```
catena2w: We already have this method in transactions endpoint, I don't think we need to similar methods, at least one should be removed

src/main/scala/org/ergoplatform/api/WalletApiRoute.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  override val route: Route = (pathPrefix("wallet") & withCors) {
+    balancesRoute ~
+      unconfirmedBalanceRoute ~
+      sendTransactionRoute ~
+      generateTransactionRoute ~
+      generateAndSendTransactionRoute
+  }
+
+  def sendTransactionRoute: Route = (path("transaction") & post & entity(as[ErgoTransaction])) { tx =>
```
catena2w: It is copy-paste of https://github.com/ergoplatform/ergo/blob/master/src/main/scala/org/ergoplatform/api/TransactionsApiRoute.scala#L30

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoProvingInterpreter.scala
```diff
@@ -0,0 +1,91 @@
...
+import scorex.crypto.hash.Blake2b256
+import sigmastate.AvlTreeData
+import sigmastate.interpreter.{ContextExtension, CostedProverResult, ProverInterpreter}
+import scala.util.{Failure, Success, Try}
+
+
+/**
+  * A class which is holding secrets and signing transactions.
+  *
+  * Currently it just generates few secrets from a seed and sign a transaction (against input boxes to spend and
```
catena2w: How to define number of secrets (what is `few` here)?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoProvingInterpreter.scala
```diff
@@ -0,0 +1,91 @@
...
+           boxesToSpend: IndexedSeq[ErgoBox],
+           stateContext: ErgoStateContext): Try[ErgoTransaction] = Try {
+
+    require(unsignedTx.inputs.length == boxesToSpend.length)
+
+    unsignedTx.inputs.zip(boxesToSpend).foldLeft(Try(IndexedSeq[Input]() -> 0L)) {
+      case (inputsCostTry, (unsignedInput, inputBox)) =>
+        require(util.Arrays.equals(unsignedInput.boxId, inputBox.id))
+
+        inputsCostTry match {
```
catena2w: Looks like this match can be replaced by `flatMap`

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWallet.scala
```diff
@@ -1,36 +1,75 @@
...
+import scala.util.{Failure, Success, Try}
+import org.ergoplatform.local.TransactionGenerator.StartGeneration
+import org.ergoplatform.local.TransactionGeneratorRef
+import org.ergoplatform.nodeView.history.ErgoHistoryReader
+
+
 
-class ErgoWallet extends Vault[ErgoTransaction, ErgoPersistentModifier, ErgoWallet]
-  with ScorexLogging {
+class ErgoWallet(nodeViewHolderRef: ActorRef,
```
catena2w: Please write comments with the purposes if this class

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWallet.scala
```diff
@@ -1,36 +1,75 @@
...
+                 historyReader: ErgoHistoryReader,
+                 settings: ErgoSettings)(implicit val actorSystem: ActorSystem)
+  extends Vault[ErgoTransaction, ErgoPersistentModifier, ErgoWallet] with ErgoWalletReader with ScorexLogging {
 
+  override lazy val actor: ActorRef = actorSystem.actorOf(Props(classOf[ErgoWalletActor], settings))
 
+  if (settings.testingSettings.transactionGeneration) {
```
catena2w: Not clean, why transaction generator is a child of the wallet

src/main/scala/org/ergoplatform/nodeView/wallet/TrackedBox.scala
```diff
@@ -0,0 +1,250 @@
...
+  def certainty: BoxCertainty
+
+  final def spent: Boolean = spendingStatus.spent
+  final def onchain: Boolean = onchainStatus.onchain
+  final def certain: Boolean = certainty.certain
+
+  /**
+    * Transaction created the box
+    */
+  def creationTx: ErgoTransaction
```
catena2w: For consistent view, transaction that spent the box should be somewhere here

src/main/scala/org/ergoplatform/nodeView/wallet/TrackedBox.scala
```diff
@@ -0,0 +1,250 @@
...
+  def value: Long = box.value
+
+  lazy val assets: Map[ModifierId, Long] = box.additionalTokens.map { case (id, amt) =>
+    bytesToId(id) -> amt
+  }.toMap
+
+  /**
+    * Register this box in a wallet storage
+    */
+  def register(storage: WalletStorage): Unit = storage.putToRegistry(this)
```
catena2w: Why such a syntax? Our regular syntax is `storage.put(object)` rather than `object.register(storage)`

src/main/scala/org/ergoplatform/nodeView/wallet/TrackedBoxSerializer.scala
```diff
@@ -0,0 +1,214 @@
...
+      }
+    }
+  }
+
+}
+
+class SpentOffchainBoxSerializer(txLookup: TransactionLookup) extends TypedBoxSerializer[SpentOffchainBox] {
+
+  def write(spentOffchainBox: SpentOffchainBox, w: ByteWriter): Unit = {
+    import spentOffchainBox._
```
catena2w: this import looks quite strange. I think it would be much cleaner to write `spentOffchainBox.creationTx` and so on for class fields

src/main/scala/org/ergoplatform/nodeView/wallet/TrackedBoxSerializer.scala
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+
+}
+
+object TrackedBoxSerializer {
+  type TransactionLookup = ModifierId => Option[ErgoTransaction]
+}
+
+class UnspentOffchainBoxSerializer(txLookup: TransactionLookup) extends TypedBoxSerializer[UnspentOffchainBox] {
```
catena2w: Do we need this 4 separate case classes (both for serializers and boxes)? I think single class `TypedBoxSerializer` with parameters `(spendingStatus, onchainStatus)` would be enough and will reduce code size.
These classes implementations are very similar, read/write fields are almost copy-paste of each other

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWallet.scala
catena2w: `nodeViewHolderRef` and `historyReader` looks quite bad here.
`Wallet` should not send messages to `NodeViewHolder`,  `NodeViewHolder` should call `Wallet` when needed.
And interaction between `Wallet` and `History` should be done by `NodeViewHolder`, their direct communication make code tight coupled

src/main/scala/org/ergoplatform/api/WalletApiRoute.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.ergoplatform.settings.ErgoSettings
+import scorex.core.NodeViewHolder.ReceivableMessages.LocallyGeneratedTransaction
+import scorex.core.api.http.ApiError.BadRequest
+import scorex.core.api.http.ApiResponse
+import scorex.core.settings.RESTApiSettings
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+case class WalletApiRoute(readersHolder: ActorRef, nodeViewActorRef: ActorRef, ergoSettings: ErgoSettings)
```
catena2w: Looks like there is no way to get addresses of out wallet. Let's add a method that return sequence of addresses of our wallet, otherwise it is quite useless as soon as we do not know, how to request money

src/main/resources/api/openapi.yaml
```diff
@@ -97,6 +92,30 @@ components:
           type: integer
           format: int32
 
+    PaymentRequest:
```
catena2w: Let's make assets and registers optional, most of users do not need them.

src/main/resources/api/openapi.yaml
```diff
@@ -1030,3 +1076,140 @@ paths:
...
+              schema:
+                type: string
+        default:
+          description: Error
+          content:
+            text/plain; charset=UTF-8:
+              schema:
+                type: string
+
+  /wallet/transaction/payment:
```
catena2w: Example request
```
[
  {
    "address": "string",
    "value": 0,
    "assets": [
      {
        "tokenId": "4ab9da11fc216660e974842cc3b7705e62ebb9e0bf5ff78e53f9cd40abadd117",
        "amount": 1000
      }
    ],
    "registers": {
      "R4": "7c"
    }
  }
]
```
does not work here with correct amount and address

mike-aksarin: this was because of register value (earlier it worked), fixed

src/main/scala/org/ergoplatform/api/WalletApiRoute.scala
mike-aksarin: introduced `wallet/addresses` API method

src/main/scala/org/ergoplatform/nodeView/wallet/TrackedBoxSerializer.scala
mike-aksarin: This was to make code looking native to Box and to avoid "train wreck" call chaining like `spentOffchainBox.creationTx.id`. 

But ok, I'll to remove this

src/main/scala/org/ergoplatform/nodeView/wallet/TrackedBoxSerializer.scala
mike-aksarin: Using one type for all box types reduces code size but also reduces type safety when using boxes and lacks code clarity when creating them. 

I'm about to finish code rearrangement for this

src/main/scala/org/ergoplatform/nodeView/wallet/TrackedBox.scala
kushti: Spending transaction exists only for the spent box,  transaction which creates the box is presenting always.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  override def hashCode(): Int = Ints.fromByteArray(pubkeyBytes.takeRight(4))
+}
+
+object P2PKAddress {
+  val addressTypePrefix: Byte = 1: Byte
+
+  def apply(pubkey: ProveDlog): P2PKAddress = {
+    val bs = pubkey.h.getEncoded(true)
```
greenhat: `val bs = ValueSerializer.serialize(pubkey)` fits better since we decode it with `DataSerializer`.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
```diff
@@ -0,0 +1,227 @@
...
+
+case class ErgoAddressEncoder(settings: ErgoSettings) {
+
+  import ErgoAddressEncoder._
+
+  val ChecksumLength = 4
+
+  private val networkPrefix = settings.chainSettings.addressPrefix
+
+  def toString(address: ErgoAddress): String = {
```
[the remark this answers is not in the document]
kushti: .toString implemented

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
```diff
@@ -0,0 +1,227 @@
...
+    val withNetworkByte = (networkPrefix + address.addressTypePrefix).toByte +: address.contentBytes
+
+    val checksum = hash256(withNetworkByte).take(ChecksumLength)
+    Base58.encode(withNetworkByte ++ checksum)
+  }
+
+  def fromString(addrStr: String): Try[ErgoAddress] = Base58.decode(addrStr).flatMap { bytes =>
```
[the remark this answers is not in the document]
kushti: what about exceptions thrown within map() then?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
```diff
@@ -0,0 +1,240 @@
...
+      }
+
+      val bs = withoutChecksum.tail
+
+      addressType match {
+        case P2PKAddress.addressTypePrefix =>
+          val buf = ByteBuffer.wrap(bs)
+          val r = new ByteBufferReader(buf).mark()
+          val ge = DataSerializer.deserialize[SGroupElement.type](SGroupElement, r)
+          val pd = ProveDlog(ge)
```
greenhat: @kushti After using `ValueSerializer` for P2PK serialization (per my suggestion) we need to deserialize ProveDlog here with `val pd = ValueSerializer.deserialize(r).asInstanceOf[ProveDlog]` since `ProveDlogSerializer` prefixes with it's op code byte. If we want to remove this extra byte we should switch back to serializing the underlying group element. Sorry I missed that in my initial suggestion.

src/main/scala/org/ergoplatform/nodeView/wallet/DefaultBoxSelector.scala
```diff
@@ -0,0 +1,94 @@
...
+  * Default implementation of the box selector. It simply picks boxes till sum of their monetary values
+  * meets target Ergo balance, then it checks which assets are not fulfilled and adds boxes till target
+  * asset values are met.
+  */
+object DefaultBoxSelector extends BoxSelector {
+
+  private def formChangeBoxes(changeBalance: Long,
+                              changeBoxesAssets: Seq[mutable.Map[ModifierId, Long]]): Option[Seq[(Long, Map[ModifierId, Long])]] = {
+    //at least 1 ergo token should be assigned per a created box
+    if (changeBoxesAssets.size > changeBalance) {
```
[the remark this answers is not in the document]
mike-aksarin: I think, there is a difference only for endless collections like `Stream`.

Also the default implementation of `Seq. lengthCompare` is O(N), but it should be overridden in sequences with known length like `Vector`

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
```diff
@@ -0,0 +1,227 @@
...
+  *
+  * For an address type, we form content bytes as follows:
+  *
+  * P2PK - serialized (compressed) public key
+  * P2sH - 160 bit of the script
+  * P2S  - serialized script
+  *
+  * Address examples for testnet:
+  *
+  * 3   - P2PK (3WvsT2Gm4EpsM9Pg18PdY6XyhNNMqXDsvJTbbf6ihLvAmSb7u5RN)
```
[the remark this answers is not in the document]
kushti: Need to think, but better to do explicit check of the first byte AFTER decoding

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
kushti: @greenhat  fixed!

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoAddress.scala
```diff
@@ -0,0 +1,227 @@
...
+    val withNetworkByte = (networkPrefix + address.addressTypePrefix).toByte +: address.contentBytes
+
+    val checksum = hash256(withNetworkByte).take(ChecksumLength)
+    Base58.encode(withNetworkByte ++ checksum)
+  }
+
+  def fromString(addrStr: String): Try[ErgoAddress] = Base58.decode(addrStr).flatMap { bytes =>
+    Try {
+      val headByte = bytes.head
+      require(headByte >= networkPrefix)
```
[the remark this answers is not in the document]
kushti: headByte = networkPrefix + addressType, not networkPrefix. Testnet address will pass this test in the mainnet but fail few lines after in a pattern matching

PR #365, opened 2018-07-16
Known peers it test

VictorMikheev:


src/it/scala/org/ergoplatform/it/Docker.scala
```diff
@@ -152,7 +154,7 @@ class Docker(suiteConfig: Config = ConfigFactory.empty, tag: String = "ergo_inte
     val networkingConfig = ContainerConfig.NetworkingConfig
       .create(Map(networkName -> endpointConfigFor(ip)).asJava)
 
-    val knownPeersSetting = seedAddress.fold("")(sa => s" -Dscorex.network.knownPeers.0=$sa")
+    val knownPeersSetting = knownPeers(nodes, nodeConfig)
```
mike-aksarin: we need `knownPeers ` function in all integration tests and it always almost the same. I think it worth adding default implementation for this method at least. There is no need for making it abstract here

src/it/scala/org/ergoplatform/it/FourNodesSuite.scala
```diff
@@ -8,6 +9,15 @@ class FourNodesSuite extends IntegrationSuite {
 
   private val nodesCount = 4
 
+  def knownPeers(nodes: Map[String, Node], nodeConfig: Config) = {
+    val node01 = nodes.map(_._2).find(_.settings.scorexSettings.network.nodeName == "node01")
```
mike-aksarin: i don't think this code became better after this change

src/it/scala/org/ergoplatform/it/ThreeNodesSuite.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def knownPeers(nodes: Map[String, Node], nodeConfig: Config) = {
+    val node01 = nodes.map(_._2).find(_.settings.scorexSettings.network.nodeName == "node01")
+    val node02 = nodes.map(_._2).find(_.settings.scorexSettings.network.nodeName == "node02")
+
+    val ls = List(node01, node02).flatten.map(_.settings.scorexSettings.network.nodeName)
+
+    if (nodeConfig.getString("scorex.network.nodeName") == "node02" && node01.isDefined) {
+      val n = node01.get
+      s" -Dscorex.network.knownPeers.0=${n.nodeInfo.networkIpAddress}:${n.nodeInfo.containerNetworkPort}"
```
mike-aksarin: it worth introducing helper method in `Docker` class for the known peers setting to avoid ugly code repitition

src/it/scala/org/ergoplatform/it/KnownNodesSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  val blocksCount = 5
+  val forkDepth = blocksCount
+
+  s"The third node knows first node" in {
+
+    val node01 = nodes.find(_.settings.scorexSettings.network.nodeName == "node01")
+    val node03 = nodes.find(_.settings.scorexSettings.network.nodeName == "node03")
+
+    val result = node03.map(_.waitForPeers(2).map { peers =>
```
mike-aksarin: `waitForPeers` doesn't include `knownPeers` currently. You can try `/peers/all` call to get all the known peers, or establish link between nodes with `/peers/connect` API, not `knownPeers` setting

catena2w:
Recreated from local branch https://github.com/ergoplatform/ergo/pull/371

PR #366, opened 2018-07-16
Use cache in HistoryStorage

catena2w:


lock.sbt
```diff
@@ -5,8 +5,12 @@ dependencyOverrides in ThisBuild ++= Seq(
...
   "ch.qos.logback" % "logback-core" % "1.3.0-alpha4",
   "com.chuusai" % "shapeless_2.12" % "2.3.2",
-  "com.google.code.findbugs" % "jsr305" % "3.0.2",
-  "com.google.guava" % "guava" % "21.0",
+  "com.github.cb372" % "scalacache-core_2.12" % "0.24.2",
+  "com.github.cb372" % "scalacache-guava_2.12" % "0.24.2",
+  "com.google.code.findbugs" % "jsr305" % "1.3.9",
+  "com.google.errorprone" % "error_prone_annotations" % "2.1.3",
+  "com.google.guava" % "guava" % "25.1-jre",
+  "com.google.j2objc" % "j2objc-annotations" % "1.1",
```
kushti: Ain't we going to add a lot of dependencies to just use some trivial FP sugar? I would like to propose to use Guava directly.

PR #372, opened 2018-07-19
Add test case for checking that ergo miner filter out double spends.

### LINKED ISSUE #356: Check that a miner is creating a valid block candidate
There's a suspicion that a  miner is creating an invalid block candidate if e.g. there are double-spending transactions in the mempool.

daron666:
Fixes #356 

I found that we already have some kind of double-spending protection, so I decided to write some test cases to check everything works as expected, also I was trying to create test case using alive node view holder node but got no success, cause tx validation always fails with strange error `None.get`.


daron666:
Also fixes #373

daron666:
@kushti @catena2w here is the test case (3rd in ErgoMinerSpec) that shows how miner could mine blocks with double spend tx, no filtering or invalidation has been proc by second coinbase tx.

How it works
1 Create coinbase tx with random proposition
2 Add to the mempool
3 Start mine
4 First block has been mined successfully with both coinbase txs.

Also this handmade tx wasn't evicted from mempool after being mined.

Please don't merge this pr, It's still in progress.

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +69,79 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+      offlineGeneration = true,
+      verifyTransactions = true)
+    val chainSettings = defaultSettings.chainSettings.copy(blockInterval = 2.seconds)
+    val ergoSettings = defaultSettings.copy(nodeSettings = nodeSettings, chainSettings = chainSettings)
+
+    val nodeViewHolderRef: ActorRef = ErgoNodeViewRef(ergoSettings, timeProvider, emission)
+    val readersHolderRef: ActorRef = ErgoReadersHolderRef(nodeViewHolderRef)
+    val minerRef: ActorRef = ErgoMinerRef(ergoSettings, nodeViewHolderRef, readersHolderRef, timeProvider, emission)
+
+    val state: UtxoState = await(
```
catena2w: You can send GetReaders message and receive all readers

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +69,79 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+      (nodeViewHolderRef ? GetDataFromCurrentView[ErgoHistory, UtxoState, ErgoWallet, ErgoMemPool, UtxoState]
+        (v => v.state)).mapTo[UtxoState]
+    )
+
+    val minerProp = DLogProverInput(BigIntegers
```
catena2w: Looks like it is not used. But it is unclear, why do you need 2 propositions

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +69,79 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+          logger.error(v.pool.unconfirmed.toString)
+          v.pool.unconfirmed.values.size
+        }).mapTo[Int]
+    )
+
+    unconfirmedSize shouldBe 1
+
+    minerRef ! StartMining
+
+    expectNoMessage(6 seconds)
```
catena2w: Why do you expect, that block is mined here?

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +69,79 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+    )
+
+    val minerProp = DLogProverInput(BigIntegers
+      .fromUnsignedByteArray(ergoSettings.scorexSettings.network.nodeName.getBytes())
+    ).publicImage
+
+    val fakeProp = DLogProverInput(BigIntegers.fromUnsignedByteArray("ttttttttt".getBytes())).publicImage
+
+    val emissionBox = state.emissionBoxOpt.get
+    val tx = ErgoMiner.createCoinbase(emissionBox, state.stateContext.height, Seq.empty, fakeProp, emission)
```
catena2w: Where is the second transaction in the pool? It is assumed to be generated by the miner?

catena2w:
@daron666 I tried this test locally, no blocks where generated, it failed in the last line with a error `1 was not equal to 0` (which is correct - since no blocks where mined, transaction in the pool is still there).

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +69,79 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+
+    nodeViewHolderRef ! LocallyGeneratedTransaction[ErgoTransaction](tx)
+    expectNoMessage(1 seconds)
+
+    val unconfirmedSize = await(
+      (nodeViewHolderRef ? GetDataFromCurrentView[ErgoHistory, UtxoState, ErgoWallet, ErgoMemPool, Int]
+        {v =>
+          logger.error(v.pool.unconfirmed.toString)
```
catena2w: Why error log level here (and why do we need this log at all)?

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
catena2w: When I increase this value to `expectNoMessage(20.seconds)`, test passed.

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +71,90 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+
+    val minerRef: ActorRef = ErgoMinerRef(ergoSettings, nodeViewHolderRef, readersHolderRef, timeProvider, emission, TrueLeaf)
+    expectNoMessage(1 second)
+    val r: Readers = Await.result((readersHolderRef ? GetReaders).mapTo[Readers], 10 seconds)
+
+    val state: UtxoStateReader = r.s.asInstanceOf[UtxoStateReader]
+    val history: ErgoHistoryReader = r.h
+    val startBlock = history.bestHeaderOpt
+
+    val newBLockDuration = 30 seconds
```
catena2w: variable name is formatted incorrectly

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +71,90 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+    val inputs = validErgoTransactionGenTemplate(0, -1, 100, 100).sample.get._1
+    val (l, r) = inputs.splitAt(50)
+    val tx_1 = validTransactionGen(l).sample.get
+    val tx_2 = validTransactionGen(r :+ l.last).sample.get
+
+    ErgoMiner.fixTxsConflicts(Seq(tx_1, tx_2, tx)) should contain theSameElementsAs Seq(tx_1, tx)
+    ErgoMiner.fixTxsConflicts(Seq(tx_2, tx_1, tx)) should contain theSameElementsAs Seq(tx_2, tx)
+  }
+
+  ignore should "create own coinbase transaction, if there is already a transaction, that spends emission box" in {
```
catena2w: Let's implement it

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +71,90 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+    // TODO
+  }
+
+  it should "work correctly with 2 coinbase txs in pool" in {
+    val tmpDir = createTempDir
+
+    type msgType = SemanticallySuccessfulModifier[_]
+    val msgClass = classOf[msgType]
```
catena2w: Variable name is useless. Let's use name that describes a variable content

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -61,6 +71,90 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+    val outputs2 = IndexedSeq(new ErgoBoxCandidate(boxToDoubleSpend.value, prop2))
+    val tx2 = new ErgoTransaction(IndexedSeq(input), outputs2)
+
+    nodeViewHolderRef ! LocallyGeneratedTransaction[ErgoTransaction](tx1)
+    nodeViewHolderRef ! LocallyGeneratedTransaction[ErgoTransaction](tx2)
+    expectNoMessage(1 seconds)
+
+    r.m.unconfirmed.size shouldBe 2
+
+    testProbe.expectMsgClass(newBLockDuration, msgClass)
```
catena2w: Why waiting for 3 blocks here?

daron666: @catena2w cause miner already have new candidates while we are making this tx, 3 blocks is `for sure` duration for tx to be mined.

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -49,7 +50,7 @@ class ErgoMiner(ergoSettings: ErgoSettings,
   private var candidateOpt: Option[CandidateBlock] = None
   private val miningThreads: mutable.Buffer[ActorRef] = new ArrayBuffer[ActorRef]()
 
-  private val minerProp: Value[SBoolean.type] = {
+  private val minerProp: Value[SBoolean.type] = minerPropOpt.getOrElse{
```
catena2w: Code formatting looks incorrect

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -276,6 +279,15 @@ object ErgoMinerRef {
            (implicit context: ActorRefFactory): ActorRef =
     context.actorOf(props(ergoSettings, viewHolderRef, readersHolderRef, timeProvider, emission))
 
+  def apply(ergoSettings: ErgoSettings,
```
catena2w: Why 2 apply methods instead of one with minerProp being Option?

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -2,30 +2,41 @@ package org.ergoplatform.mining
...
-class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTestHelpers {
+class ErgoMinerSpec extends FlatSpec with ErgoTestHelpers with ValidBlocksGenerators {
 
   val defaultAwaitDuration: FiniteDuration = 5.seconds
   implicit val timeout: Timeout = Timeout(defaultAwaitDuration)
 
   def await[A](f: Future[A]): A = Await.result[A](f, defaultAwaitDuration)
 
-  it should "not freeze while generating candidate block with large amount of txs" in {
+  it should "not freeze while generating candidate block with large amount of txs" in new TestKit(ActorSystem()){
```
catena2w: Why new testkit here?
Also code formatting looks incorrect

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -60,7 +71,135 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+
+    val emissionBox = r.s.asInstanceOf[UtxoStateReader].emissionBoxOpt.get
+
+    val prop1 = DLogProverInput(BigIntegers.fromUnsignedByteArray("test1".getBytes())).publicImage
+
+    val input = Input(emissionBox.id, ProverResult(Array.emptyByteArray, ContextExtension.empty))
+    val outputs = IndexedSeq(new ErgoBoxCandidate(1000000L, prop1))
+    val tx = new ErgoTransaction(IndexedSeq(input), outputs)
+
+    nodeViewHolderRef ! LocallyGeneratedTransaction[ErgoTransaction](tx)
```
catena2w: Please add a check that it was added to mempool

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -60,7 +71,135 @@ class ErgoMinerSpec extends TestKit(ActorSystem()) with FlatSpecLike with ErgoTe
...
+    val inputs = validErgoTransactionGenTemplate(0, -1, 100, 100).sample.get._1
+    val (l, r) = inputs.splitAt(50)
+    val tx_1 = validTransactionGen(l).sample.get
+    val tx_2 = validTransactionGen(r :+ l.last).sample.get
+
+    ErgoMiner.fixTxsConflicts(Seq(tx_1, tx_2, tx)) should contain theSameElementsAs Seq(tx_1, tx)
+    ErgoMiner.fixTxsConflicts(Seq(tx_2, tx_1, tx)) should contain theSameElementsAs Seq(tx_2, tx)
+  }
+
+  it should "create own coinbase transaction, if there is already a transaction, that spends emission box" in new TestKit(ActorSystem()){
```
catena2w: When I looks to Miner code, I can't understand, how it works. Can you describe, why this test passed?

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
daron666: @catena2w 
Cause we need to guarantee isolation for each case. Sometimes tests being failed with strange errors if we use one system for every case.

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
daron666: There is a mandatory that only miner can spent emission box, so every inputs of emission box outside of coinbase filtering out from block.

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
daron666: @catena2w It won't go to mempool, it won't pass validation.

daron666:
@catena2w done.

PR #386, opened 2018-07-30
Scorex with modifiers status keeper

### LINKED ISSUE #342: More stable and fast synchronization process
Modifiers can be in different conditions, until they will be in a history:
- To be expected to exist somewhere (when we got a block, but not it's block sections)
- To be requested from someone
- To be put in ModifiersCache
- To be applied to a history

We should support this state machine and us this state machine in our NodeViewSynchronizer.

catena2w:
close #342.
Use Scorex from unmerged PR https://github.com/ScorexFoundation/Scorex/pull/280


src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -40,52 +42,57 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
       }
+    }
   }
 
   /**
-    * Broadcast inv on successful Header and BlockTransactions application
-    * Do not broadcast Inv messages during initial synchronization (the rest of the network should already have all
-    * this messages)
-    *
+    * If new enough semantically valid ErgoFullBlock was applied, send inv for block header and all it's sections
```
kushti: it's => its

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/FullBlockSectionProcessor.scala
```diff
@@ -70,19 +70,22 @@ trait FullBlockSectionProcessor extends BlockSectionProcessor with FullBlockProc
...
           fatal(s"Header ${header.encodedId} for modifier ${m.encodedId} is semantically invalid")
         }
-        .validate(header.height >= minimalHeight) {
-          fatal(s"Too old modifier ${m.encodedId}: ${header.height} < $minimalHeight")
+        .validate(!isHeadersChainSynced || header.height >= minimalHeight) {
+          error(s"Too old modifier ${m.encodedId}: ${header.height} < $minimalHeight")
+        }
+        .validate(isHeadersChainSynced) {
+          error("Headers chain is not synced yet")
+        }
```
kushti: From the two conditions above, validation will always fail.

src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
```diff
@@ -66,32 +66,34 @@ class UtxoState(override val persistentProver: PersistentBatchAVLProver[Digest32
...
-                                       height: Height) = Try {
-
-    val createdOutputs = transactions.flatMap(_.outputs).map(o => (ByteArrayWrapper(o.id), o)).toMap
-    val totalCost = transactions.map { tx =>
-      tx.statelessValidity.get
-      val boxesToSpend = tx.inputs.map(_.boxId).map { id =>
-        createdOutputs.get(ByteArrayWrapper(id)).orElse(boxById(id)) match {
-          case Some(box) => box
-          case None => throw new Error(s"Box with id ${Algos.encode(id)} not found")
+                                       height: Height) = persistentProver.synchronized {
```
kushti: UtxoState is not thread-safe, and thus should encapsulated in NodeViewHolder instance (which is an actor). Doing synchronization on writes then does not make sense.

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
```diff
@@ -34,16 +34,21 @@ class HistoryStorage(indexStore: Store, objectsStore: ObjectsStore, config: Cach
...
+              modifiersCache.put(id, pm)
               Some(pm)
             case None => None
           }
         }
     }
   }
 
-  def getIndex(id: ByteArrayWrapper): Option[ByteArrayWrapper] = indexStore.get(id)
+  def getIndex(id: ByteArrayWrapper): Option[ByteArrayWrapper] = Option(indexCache.getIfPresent(id)).orElse {
```
kushti: What does the term "index" mean here?

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -94,7 +94,7 @@ class ErgoMiner(ergoSettings: ErgoSettings,
   }
 
   private def shouldStartMine(b: ErgoFullBlock): Boolean = {
-    ergoSettings.nodeSettings.mining && b.header.timestamp >= startTime
+    ergoSettings.nodeSettings.mining && b.header.isNew(timeProvider, ergoSettings.chainSettings.blockInterval)
```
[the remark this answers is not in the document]
catena2w: `startTme` removed.
Start mining condition is now defined by the same function we use to estimate, that headers chain is synced, etc. Old condition was bad for long synchronization process (before current optimizations, synchronization took several hours, but anyway synchronization will take more time with increased chain length) - mining was started on quite old blocks.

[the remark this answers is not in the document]
catena2w: Do you know the way how to estimate that current block is the best one more precisely?

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
catena2w: A regular database index.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/FullBlockSectionProcessor.scala
catena2w: It does not fail, if `isHeadersChainSynced == true && header.height >= minimalHeight`. Rewrote to this single condition

src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
catena2w: This synchronization is done together with `proofsForTransactions` at https://github.com/ergoplatform/ergo/pull/386/files#diff-804b1a8751b1f609f0501f81261be3deR73. We do not need to lock whole `NodeViewHolder` during proof generation, but proof generation during block application may lead to incorrect proofs, and thus miner will generate incorrect block locally.

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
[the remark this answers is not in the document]
kushti: Let's use double the interval, with just the interval a miner will wait for longer potentially.

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
kushti: Then indexCache as a "cache of a regular database index" is not very descriptive. Please use more descriptive name or write  comments.

kushti:
@catena2w approved but please address the comments. Also, I would highly recommend to document validation rules touched in this PR (in a separate PR), they are not covered in the specification currently!

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
catena2w: Wrote a description for the whole class

catena2w:
Created issue https://github.com/ergoplatform/ergo/issues/410 for validation documentation

PR #402, opened 2018-08-08
Minor code fixes

builat:
* mutability removed from parseBytes method
* if block refactored in validate method.

### DIFF
--- a/src/main/scala/org/ergoplatform/modifiers/history/PoPoWProof.scala
+++ b/src/main/scala/org/ergoplatform/modifiers/history/PoPoWProof.scala
@@ -41,29 +41,54 @@ object PoPoWProof {
 @SuppressWarnings(Array("TraversableHead", "CollectionIndexOnNonIndexedSeq"))
 class PoPoWProofUtils(powScheme: PowScheme) {
 
+  val checkList: Seq[(PoPoWProof, BigInt) => Option[String]] = Seq(
+    (proof, _) =>
+      if (proof.suffix.lengthCompare(proof.k) != 0) Some(s"Incorrect suffix ${proof.suffix.length} != ${proof.k}")
+      else None,
+
+    (proof, _) =>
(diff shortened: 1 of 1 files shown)


kushti:
@builat please take a look to ModifierValidator class, is it possible to reuse it, instead of building a Seq?

builat:
@kushti well, if I understand Modifier Validator class, then definitely in possible to reuse it, but, anyway we should declare all the validation conditions mentioned in declared seq.

catena2w:
@builat please use `ModifierValidator` and `ValidationResult` here, you may see an example here - https://github.com/ergoplatform/ergo/blob/master/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala#L327.
P.S. `PoPoWProof` class is a bit outdated, so now it's style differs from other code, but it's better to make it similar to other modifiers validation.

builat:
@catena2w not sure about ".toTry" method. But to be compliant with tests put it there.

src/main/scala/org/ergoplatform/modifiers/history/PoPoWProof.scala
```diff
@@ -39,32 +41,39 @@ object PoPoWProof {
...
+      .validate(proof.m >= 1) {
+        error(s"m should positive, ${proof.m} given")
+      }
+      .validate(proof.suffix.head.interlinks(proof.i) == proof.innerchain.last.id) {
+        error(s"Incorrect link form suffix to innerchain in $proof")
+      }
+      .validate(proof.innerchain.length >= proof.m) {
+        error(s"Innerchain length is not enough in $proof")
+      }
+      .validate(!proof.innerchain.forall(h => powScheme.realDifficulty(h) < innerDifficulty)) {
```
kushti: Seems like condition is not inversed here. Please check here and further.

PR #403, opened 2018-08-09
Add Genesis block id to node info

### LINKED ISSUE #292: Indicate on what net node is in NodeInfo
We need some introspection about what test- or mainnet node is currently on.

Return this info in some field, based on current used Genesis block.

daron666:
Fixes #292

### DIFF
--- a/src/main/scala/org/ergoplatform/local/ErgoStatsCollector.scala
+++ b/src/main/scala/org/ergoplatform/local/ErgoStatsCollector.scala
@@ -34,9 +34,11 @@ class ErgoStatsCollector(viewHolderRef: ActorRef,
     context.system.scheduler.schedule(10.second, 10.second)(peerManager ! GetConnectedPeers)(context.system.dispatcher)
   }
 
+  var genesisBlockIdOpt: Option[String] = None
+
   var nodeInfo = NodeInfo(settings.scorexSettings.network.nodeName, Version.VersionString, 0, 0, None,
     settings.nodeSettings.stateType, None, isMining = settings.nodeSettings.mining, None, None, None, None,
-    timeProvider.time())
+    timeProvider.time(), genesisBlockIdOpt)
 
   override def receive: Receive = onConnectedPeers orElse getNodeInfo orElse onMempoolChanged orElse
     onHistoryChanged orElse onSemanticallySuccessfulModification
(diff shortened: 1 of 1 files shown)


src/main/scala/org/ergoplatform/local/ErgoStatsCollector.scala
```diff
@@ -18,28 +22,46 @@ import scorex.core.utils.{NetworkTimeProvider, ScorexLogging}
...
+  implicit val timeout = Timeout(5 seconds)
+
   override def preStart(): Unit = {
     context.system.eventStream.subscribe(self, classOf[ChangedHistory[_]])
     context.system.eventStream.subscribe(self, classOf[ChangedMempool[_]])
     context.system.eventStream.subscribe(self, classOf[SemanticallySuccessfulModifier[_]])
-    context.system.scheduler.schedule(10.second, 10.second)(peerManager ! GetConnectedPeers)(context.system.dispatcher)
+    context.system.scheduler.schedule(10.second, 10.second)(peerManager ! GetConnectedPeers)
+
+    (readersHolder ? GetDataFromHistory[Init](r => Init(r))).mapTo[Init].pipeTo(self)
```
catena2w: Readers holder already have a `GetReaders` message, why do you need one more?

src/main/scala/org/ergoplatform/local/ErgoStatsCollector.scala
```diff
@@ -17,29 +20,46 @@ import scorex.core.network.peer.PeerManager.ReceivableMessages.GetConnectedPeers
...
+  implicit val timeout = Timeout(5 seconds)
+
   override def preStart(): Unit = {
     context.system.eventStream.subscribe(self, classOf[ChangedHistory[_]])
     context.system.eventStream.subscribe(self, classOf[ChangedMempool[_]])
     context.system.eventStream.subscribe(self, classOf[SemanticallySuccessfulModifier[_]])
-    context.system.scheduler.schedule(10.second, 10.second)(peerManager ! GetConnectedPeers)(context.system.dispatcher)
+    context.system.scheduler.schedule(10.second, 10.second)(peerManager ! GetConnectedPeers)
+
+    (readersHolder ? GetReaders).mapTo[Readers].pipeTo(self)
```
catena2w: I think readersHolder ! GetReaders should be enough, no need to wait the response

src/main/scala/org/ergoplatform/local/ErgoStatsCollector.scala
```diff
@@ -21,25 +22,37 @@ import scala.concurrent.duration._
...
-    timeProvider.time())
+    timeProvider.time(), None)
 
   override def receive: Receive = onConnectedPeers orElse getNodeInfo orElse onMempoolChanged orElse
-    onHistoryChanged orElse onSemanticallySuccessfulModification
+    onHistoryChanged orElse onSemanticallySuccessfulModification orElse init
+
+  private def init: Receive = {
+    case Readers(h, _, _) =>
+      nodeInfo = nodeInfo.copy(bestFullBlockOpt = h.bestFullBlockOpt,
```
catena2w: You can also use state reader and fill stateRoot and stateVerion here

PR #408, opened 2018-08-14
Add genesis id into config

### LINKED ISSUE #317: Wrong chain application check
- [x] Add optional genesis block to config to prevent synchronization stops
- [x] Write tests for incorrect blocks application (valid header, invalid block sections + valid full block, invalid transactions). Ensure, that application works correct after this attempts

daron666:
Fixes #317

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -34,6 +34,9 @@ ergo {
 
     # Number of state snapshot diffs to keep. Defines maximum rollback depth
     keepVersions = 200
+
+    # Optional genesis id block. Defines id of a block that should be applied firstly. Uncomment to use
+    # genesisId = "some_encoded_id"
   }
 
   testing {
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -34,9 +34,6 @@ ergo {
 
     # Number of state snapshot diffs to keep. Defines maximum rollback depth
     keepVersions = 200
-
-    # Optional genesis id block. Defines id of a block that should be applied firstly. Uncomment to use
-    # genesisId = "some_encoded_id"
   }
 
   testing {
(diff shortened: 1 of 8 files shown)


src/main/resources/application.conf
```diff
@@ -34,6 +34,9 @@ ergo {
 
     # Number of state snapshot diffs to keep. Defines maximum rollback depth
     keepVersions = 200
+
+    # Optional genesis id block. Defines id of a block that should be applied firstly. Uncomment to use
```
catena2w: Description: "Defines an id of the genesis block. Other genesis blocks will be considered invalid."
Put it to `chain` section and add id of current testnet genesis block to comment

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -314,6 +314,10 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
         .validateEqualIds(header.parentId, Header.GenesisParentId) { detail =>
           fatal(s"Genesis block should have genesis parent id. $detail")
         }
+        .validate(config.genesisId.fold(true) { _ == Algos.encode(header.id) }) {
```
catena2w: `forall` should be better here

src/test/scala/org/ergoplatform/nodeView/ErgoNodeViewHolderSpecification.scala
```diff
@@ -355,6 +359,102 @@ class ErgoNodeViewHolderSpecification extends ErgoNodeViewHolderTestHelpers {
     }
   }
 
+  private val t12 = TestCase("Do not apply txs with wrong header id") { fixture =>
```
catena2w: How does this relates to current PR?

daron666: Was in issue description. Add genesis id and some test cases.

image

PR #417, opened 2018-08-21
Fix for concurrent block application and proofs generation. 

### LINKED ISSUE #409: Proof creation at a time of block application
When our node is mining, it may create proofs for block transactions at the same time, as another block being applied. This leads to incorrect local block generation and node crash

catena2w:
close #409


src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
```diff
@@ -47,7 +49,7 @@ class UtxoState(override val persistentProver: PersistentBatchAVLProver[Digest32
 
   override val maxRollbackDepth = 10
 
-  override def rollbackTo(version: VersionTag): Try[UtxoState] = {
+  override def rollbackTo(version: VersionTag): Try[UtxoState] = persistentProver.synchronized {
```
mike-aksarin: To fully comply JMM, read access to `persistentProver` should be synchronized as well. This should be true for `rollbackVersions`, `boxById`, `randomBox` methods.

I'd prefer making State an actor to enable actor memory model instead

PR #421, opened 2018-08-24
Chain generator

catena2w:
Includes PR with extension

### DIFF
--- a/src/test/scala/org/ergoplatform/tools/ChainGenerator.scala
+++ b/src/test/scala/org/ergoplatform/tools/ChainGenerator.scala
@@ -10,10 +10,10 @@ import org.ergoplatform.nodeView.state.{BoxHolder, StateType, UtxoState}
 import org.ergoplatform.settings._
 import org.ergoplatform.utils.{ErgoTestHelpers, ValidBlocksGenerators}
 import scorex.core.utils.ScorexLogging
-import scorex.utils.Random
 
 import scala.annotation.tailrec
 import scala.concurrent.duration._
+import scala.util.Random
 
 /**
   * Application object for chain generation.
@@ -25,6 +25,7 @@ object ChainGenerator extends App with ValidBlocksGenerators with ErgoTestHelper
 
   val n: Char = 96
   val k: Char = 5
+  val txsSize: Int = 100 * 1024
   val pow = new EquihashPowScheme(n, k)
   val blockInterval = 2.minute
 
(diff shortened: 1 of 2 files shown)


src/test/scala/org/ergoplatform/tools/ChainGenerator.scala
```diff
@@ -0,0 +1,96 @@
...
+    val fullHistorySettings: ErgoSettings = ErgoSettings(dir.getAbsolutePath, chainSettings, settings.testingSettings,
+      nodeSettings, CacheSettings.default, settings.scorexSettings)
+    ErgoHistory.readOrGenerate(fullHistorySettings, timeProvider)
+  }
+  if (history.bestHeaderOpt.nonEmpty) {
+    log.error(s"History at ${dir.getAbsoluteFile} already exists. Exiting.")
+    System.exit(11)
+  }
+
+  var (state, boxHolder) = createUtxoState()
```
mike-aksarin: we can make `state` and `boxHolder` vals and parameters for `loop` function

src/test/scala/org/ergoplatform/tools/ChainGenerator.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def loop(last: Option[Header], acc: Seq[ErgoFullBlock]): Seq[ErgoFullBlock] = {
+    val time: Long = last.map(_.timestamp + blockInterval.toMillis).getOrElse(startTime)
+    if (time < timeProvider.time) {
+      val txsBoxes = validTransactionsFromBoxHolder(boxHolder)
+      val txs = txsBoxes._1
+      boxHolder = txsBoxes._2
+
+      @tailrec
+      def generate(candidate: CandidateBlock): ErgoFullBlock = {
```
mike-aksarin: this could be a top-level function as it looks not to use outer function context

src/test/scala/org/ergoplatform/tools/ChainGenerator.scala
```diff
@@ -0,0 +1,96 @@
...
+    log.error(s"History at ${dir.getAbsoluteFile} already exists. Exiting.")
+    System.exit(11)
+  }
+
+  var (state, boxHolder) = createUtxoState()
+
+  def loop(last: Option[Header], acc: Seq[ErgoFullBlock]): Seq[ErgoFullBlock] = {
+    val time: Long = last.map(_.timestamp + blockInterval.toMillis).getOrElse(startTime)
+    if (time < timeProvider.time) {
+      val txsBoxes = validTransactionsFromBoxHolder(boxHolder)
```
mike-aksarin: if `boxHolder` will be a `loop` function param, then we can use decomposition pattern here:
```
val (txs, newBoxHolder) = validTransactionsFromBoxHolder(boxHolder)
```

src/test/scala/org/ergoplatform/tools/ChainGenerator.scala
```diff
@@ -0,0 +1,94 @@
...
+    System.exit(11)
+  }
+
+  val (state, boxHolder) = createUtxoState()
+  val chain = loop(state, boxHolder, None, Seq())
+  log.info(s"Chain of length ${chain.length} generated")
+  chain.foreach { block =>
+    history.append(block.header).get
+  }
+  chain.foreach { block =>
```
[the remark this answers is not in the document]
catena2w: Current rules requires to apply full headers chain first, and only then start to apply block sections

PR #440, opened 2018-09-12
Scorex version with updated network layer

catena2w:
relates to unmerged https://github.com/ScorexFoundation/Scorex/pull/306


src/main/resources/application.conf
```diff
@@ -108,10 +108,10 @@ ergo {
 }
 scorex {
   network {
-    bindAddress = "0.0.0.0:9006"
+    bindAddress = "0.0.0.0:9003"
     maxInvObjects = 400
-    nodeName = "ergo-testnet1"
-    knownPeers = ["178.128.162.150:9006", "78.46.93.239:9006", "209.97.136.204:9006", "209.97.138.187:9006", "209.97.134.210:9006", "88.198.13.202:9006"]
+    nodeName = "ergo-testnet2"
+    knownPeers = ["69.55.62.58:9003", "167.99.215.203:9003"]
```
catena2w: known peers should not be changed here

src/main/resources/nodeTestnet2/application.conf
```diff
@@ -0,0 +1,29 @@
+ergo {
```
catena2w: Don't think we need this file in git. Also it's a regular config file that does not provide any info related to testnet2

PR #441, opened 2018-09-12
Supow implementation

### LINKED ISSUE #218: Inefficient outsourceability

catena2w:
Close #218, #390


src/main/scala/org/ergoplatform/crypto/pow/NumericHash.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import scala.annotation.tailrec
+
+/**
+  * One way cryptographic hash function, that produces numbers in [0,p) range
+  */
+class NumericHash(val p: BigInt) {
+  assert(p.bigInteger.bitLength() <= 256, "We use 256 bit hash here")
+  // biggest number < 2^256 that is divisible by p without remainder
+  private val validRange: BigInt = (BigInt(2).pow(256) / p) * p
```
kushti: If p is a power of 2, validRange will always be 2^256: 

  val h = new NumericHash(BigInt(2).pow(2))
  println(h.validRange == BigInt(2).pow(256)) //true

catena2w:
will be implemented in a separate project https://github.com/ergoplatform/autoleakus

PR #446, opened 2018-09-13
Fix for isHeadersChainSynced race condition

mike-aksarin:


src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -36,11 +36,12 @@ trait ToDownloadProcessor extends ScorexLogging {
   /** return true if we estimate, that our chain is synced with the network. Start downloading full blocks after that
     */
   def isHeadersChainSynced: Boolean = pruningProcessor.isHeadersChainSynced
+  def shouldDownloadBlockAtHeight(height: Int): Boolean = pruningProcessor.shouldDownloadBlockAtHeight(height)
```
catena2w: Do we need this def? Can we make it private?

src/test/scala/org/ergoplatform/nodeView/history/BlockSectionValidationSpecification.scala
```diff
@@ -77,6 +77,7 @@ class BlockSectionValidationSpecification extends HistorySpecification {
     // should not be able to apply when blocks at this height are already pruned
     history.applicableTry(section) shouldBe 'success
     history.pruningProcessor.minimalFullBlockHeightVar = history.bestHeaderOpt.get.height + 1
+    history.pruningProcessor.isHeadersChainSyncedVar = true
     history.applicableTry(section) shouldBe 'failure
     history.pruningProcessor.minimalFullBlockHeightVar = 0
```
catena2w: This line is no longer required, as soon as you set it to 0 by default

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
mike-aksarin: The idea was to use it instead of `minimalFullBlockHeight`

src/test/scala/org/ergoplatform/nodeView/history/BlockSectionValidationSpecification.scala
mike-aksarin: We need it because just two lines above we assigned an incorrect value to `minimalFullBlockHeightVar`

PR #449, opened 2018-09-18
Fix todo in transactions generator

catena2w:


src/main/scala/org/ergoplatform/local/TransactionGenerator.scala
```diff
@@ -43,6 +45,7 @@ class TransactionGenerator(viewHolder: ActorRef,
 
       viewHolder ! GetDataFromCurrentView[ErgoHistory, UtxoState, ErgoWallet, ErgoMemPool, Unit] { v =>
         currentFullHeight = v.history.headersHeight
+        v.vault.publicKeys(0, 100).onComplete(_.foreach(pks => propositions = pks))
 
         context.system.eventStream.subscribe(self, classOf[SuccessfulTransaction[ErgoTransaction]])
```
mike-aksarin: did you want this to be inside `GetDataFromCurrentView` handling by `NodeViewHolder`?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -190,8 +190,8 @@ class ErgoWalletActor(settings: ErgoSettings) extends Actor with ScorexLogging {
         sender() ! BalancesSnapshot(height, registry.unconfirmedBalance, registry.unconfirmedAssetBalances)
       }
 
-    case ReadPublicKeys(from, until) =>
-      publicKeys.slice(from, until)
+    case r: ReadPublicKeys =>
```
mike-aksarin: ReadPublicKeys(from, until) looks better to me here, but this is the matter of taste

PR #471, opened 2018-09-28
Demurrage component 

kushti:


src/main/scala/org/ergoplatform/modifiers/ErgoFullBlock.scala
```diff
@@ -38,8 +38,8 @@ case class ErgoFullBlock(header: Header,
 
   override lazy val size: Int = header.size + blockTransactions.size + adProofs.map(_.size).getOrElse(0)
 
-  override lazy val serializer: Serializer[ErgoFullBlock] = throw new Error("Never try to serialize ErgoFullBlock")
-
+  override lazy val serializer: Serializer[ErgoFullBlock] =
```
catena2w: This message means, that it would be good to implement serialization of ErgoFullBlock, I'd like to mention here, that it is not implemented by design, not because of lack of time

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -127,13 +135,12 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
         fillAssetsMap(boxesToSpend, inAssets)
         lazy val newAssetId = ByteArrayWrapper(inputs.head.boxId)
         validation.validateSeq(outAssets) {
-          case (validation, (outAssetId, outAmount)) =>
+          case (vld, (outAssetId, outAmount)) =>
```
catena2w: vld is not meaningful variable name

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -96,27 +96,35 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
   /**
     * @return total computation cost
     */
-  def statefulValidity(boxesToSpend: IndexedSeq[ErgoBox], blockchainState: ErgoStateContext): Try[Long] = {
-    lazy val lastUtxoDigest = AvlTreeData(blockchainState.digest, ErgoBox.BoxId.size)
+  def statefulValidity(boxesToSpend: IndexedSeq[ErgoBox],
+                       blockchainState: ErgoStateContext,
+                       metadata: Metadata): Try[Long] = {
```
catena2w: I thought we decided to exclude metadata from validation?

src/main/scala/org/ergoplatform/nodeView/ErgoInterpreter.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.ergoplatform.ErgoLikeContext.Height
+import org.ergoplatform._
+import org.ergoplatform.settings.{Constants, Parameters}
+import sigmastate.SBoolean
+import sigmastate.Values.Value
+import sigmastate.interpreter.Interpreter.VerificationResult
+import scala.util.Try
+
+
+class ErgoInterpreter(override val maxCost: Long = Parameters.MaxBlockCost)
```
catena2w: Class comments describing it's purposes might be useful here

src/main/scala/org/ergoplatform/nodeView/ErgoInterpreter.scala
```diff
@@ -0,0 +1,61 @@
...
+    lazy val varId = Constants.StorageIndexVarId
+
+    //no proof provided and enough time since box creation to spend it
+    if (context.currentHeight - context.self.creationHeight >= Constants.StoragePeriod
+      && proof.length == 0
+      && context.extension.values.contains(varId)) {
+
+      Try {
+        val idx = context.extension.values(varId).value.asInstanceOf[Short]
+        val outputCandidate = context.spendingTransaction.outputCandidates(idx)
```
catena2w: What about the case, when box value is not enough to pay storage fee and box should be just removed without  new output creation?

catena2w: I think it should be `Option[BoxCandidate]` here, that is None if box is just destroyed

src/main/scala/org/ergoplatform/settings/Parameters.scala
```diff
@@ -0,0 +1,14 @@
+package org.ergoplatform.settings
+
+object Parameters {
```
catena2w: Some class comments might be useful here. I expect it is assumed to be class, and it's parameters should be changed by miner votes?

src/main/scala/org/ergoplatform/settings/Parameters.scala
```diff
@@ -0,0 +1,14 @@
...
+  // Max total computation cost of a block.
+  val MaxBlockCost: Long = 1000000
+
+  // Max size of transactions section of a block.
+  val MaxBlockSize: Int = 512 * 1024
+
+  // Cost of storing 1 byte per block, in nanoErgs
+  // with default value of 12 nanoErgs, storage cost for an (ordinary) output of 80 bytes would be ~1.01 Ergo per 4 years
+  // max should be about 24 probably
```
catena2w: should we have max at all? Why 24?

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -98,7 +98,14 @@ class ErgoMinerSpec extends FlatSpec with ErgoTestHelpers with ValidBlocksGenera
...
         val unsignedTx = new UnsignedErgoTransaction(inputs, outputs)
-        val tx = prover.sign(unsignedTx, IndexedSeq(boxToSend), ErgoStateContext(r.h.fullBlockHeight, r.s.rootHash)).get
+        val tx = prover.sign(
+          unsignedTx,
+          IndexedSeq(boxToSend),
+          ergoSettings.metadata,
+          ErgoStateContext(r.h.fullBlockHeight,
+          r.s.rootHash)).get
+
+        // workaround to put tx in mempool without too much logs.
```
catena2w: There is no workaround in your PR, so I expect this comment should not be here

src/main/scala/org/ergoplatform/modifiers/ErgoFullBlock.scala
kushti: "Never try" is not informative message then, as the first reasonable question about that is "why?", and there is no any answer around. Changed message a little bit.

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
kushti: There is an issue for that (https://github.com/ScorexFoundation/sigmastate-interpreter/issues/267), but dependency is still in the sigma code we use

src/main/scala/org/ergoplatform/nodeView/ErgoInterpreter.scala
kushti: Line 28: "(box.value - maxStorageFee <= 0)". I think Boolean return value type is enough for checkExpiredBox. In future, "context.extension.values.contains(varId)" on line 45 is not needed when storage fee exceeds box value, but for now (while lazy evaluation is not done) we can not avoid it.

src/main/scala/org/ergoplatform/settings/Parameters.scala
kushti: As miner is charging "K*B*(out.value - self.value)" , but K here is taken at the moment of charging (not creation). Thus greedy miners have an incentive probably to rip short-term profits by increasing K (and limits in increasing the K are not seen).

kushti: with K = 12, a box of 80 bytes per 4 years ~= 1 Ergo
K = 24 => ~= 2 Ergo

papers/yellow/economy.tex
```diff
@@ -3,38 +3,36 @@ \section{Economic survivability}
...
-and index, while $R2.last\_height$ should be set to $R2.last\_height$ of spending box
-plus $SP$, to ensure that subsidized period starts over.
+$current\_block\_height \ge box.creation\_height + SP$, see $SP$ value below), anyone~(presumably, a miner) can
+create a new box $out$ with the exactly the same content (including the guarding
+script) except the monetary value and $R3$ content. The monetary value is to
+reduced by $K \cdot B \cdot (out.creation\_height - self.creation\_height)$ maximum, where $B$ is the spent box~($self$)
+size and $K$ is the storage cost of one byte. Thus, the difference is to be paid to the miner.
+If box value is less that the storage fee, all the box content including tokens could be spent by the miner.
+
+For efficient lookup, we require a spending proof for an expired box to contain nothing but an index of an
```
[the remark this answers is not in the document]
kushti: The text is corrected based on concrete feedback received via private discussion.

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
kushti: fixed

src/main/scala/org/ergoplatform/settings/Parameters.scala
kushti: yes, a comment added

src/main/scala/org/ergoplatform/nodeView/ErgoInterpreter.scala
kushti: added

kushti:
@catena2w all the comments have been addressed, also, K now is fee for byte per storage period (not 1 block).

kushti:
@catena2w please make another round of reviewing

papers/yellow/economy.tex
```diff
@@ -3,38 +3,37 @@ \section{Economic survivability}
...
-plus $SP$, to ensure that subsidized period starts over.
+$current\_block\_height \ge box.creation\_height + SP$, see $SP$ value below), anyone~(presumably, a miner) can
+create a new box $out$ with the exactly the same content (including the guarding
+script) except the monetary value and $R3$ content. The monetary value is to
+reduced by $K \cdot B \cdot (out.creation\_height - self.creation\_height)$ maximum, where $B$ is the spent box~($self$)
+size and $K$ is the storage cost of one byte. Thus, the difference is to be paid to the miner.
+If box value is less that the storage fee, all the box content including tokens could be spent by the miner.
+
+For efficient lookup~(to find a proper output regarding an expired input without an iteration over transaction outputs),
+we require a spending proof for an expired box to be just a context extension which contains only an index of an
```
catena2w:
```suggestion
... which contains only a variable with id $127$ containing an index of an output..
```

catena2w: From this 2 sentences it is still not clear, what is variable. I suppose to replace to something like:
```
...context extension which contains only a variable with id $127$ containing an index of an output  which is trying to spend the box.
```

src/main/scala/org/ergoplatform/settings/Constants.scala
```diff
@@ -11,17 +11,26 @@ import scorex.core.{ModifierTypeId, NodeViewModifier}
...
-  // Max size of transactions section of a block. todo: move to config
-  val MaxBlockSize: Int = 512 * 1024
+
+  val BlocksPerHour = 30
+
+  //For how many blocks a box could be put into the state with no paying.
+  //4 years
+  val StoragePeriod: Int = 4 * 365 * 24 * BlocksPerHour
+
+  val StorageContractCost: Long = 50
```
catena2w: We're going to allow to change cost for all opcodes, so it should be placed to `Parameters`.

papers/yellow/economy.tex
```diff
@@ -3,38 +3,37 @@ \section{Economic survivability}
...
+output which is trying to spend the box. The variable identifier for the index in the extension is $127$.
+
 
 We propose the following concrete parameters:
 \begin{itemize}
     \item{} $K$ - cost of storage of 1 byte of data in a State for 1 block.
```
catena2w:
```suggestion
    \item{} $K$ - cost of storing 1 byte per Constants.StoragePeriod blocks, in nanoErgos.
```

papers/yellow/economy.tex
```diff
@@ -3,38 +3,37 @@ \section{Economic survivability}
...
-    Should be determined by miner votes, $10^{-6} (Ergo/byte)$ by default.
-    \item{} $SP$ - number of blocks, box can be stored in a State for free untouched.
-    Since Ergo have tokens as a first-class citizens, it might be possible that tokens value
-    sufficiently exceed Ergo value in a box, thus $SP$ should be big enough to protect users
-    that only keep tokens in their boxes.
-    However, $SP$ should not exceed emission length to reward miners.
-    It is predefined $SP = 1051200 = 4$ years.
-\end{itemize}
\ No newline at end of file
+    Should be determined by miner votes, $12 (nanoErg/byte)$ by default, max value is 24.
```
catena2w: `$12 (nanoErg/byte)$` is definitely incorrect now.

Issue #450, opened 2018-09-18
Asset issue


catena2w:
- [x] Create API route to issue an asset
- [x] Issue and transfer assets in tx generator

catena2w:
I think in API asset issue should be similar asset issue in Waves/Ethereum and contain some usual fields, like Amount, Name, Description, Decimals. These fields should be put to registers, simplifying assets identification.

oskin1:
Closed #488, #493, #495

PR #488, opened 2018-10-09
Asset issue

oskin1:


oskin1:
#450

src/main/resources/api/openapi.yaml
```diff
@@ -1245,3 +1279,71 @@ paths:
             text/plain; charset=UTF-8:
               schema:
                 type: string
+
+  /wallet/assets/issue/generate:
```
mike-aksarin: Maybe we could use just `/wallet/assets/generate`

src/main/resources/api/openapi.yaml
```diff
@@ -1245,3 +1279,71 @@ paths:
...
+        default:
+          description: Error
+          content:
+            text/plain; charset=UTF-8:
+              schema:
+                type: string
+
+  /wallet/assets/issue:
+    post:
+      summary: Generate and send a transaction from payment
```
mike-aksarin: should be something like `asset issue transaction`

src/main/scala/org/ergoplatform/api/WalletApiRoute.scala
```diff
@@ -32,41 +34,51 @@ case class WalletApiRoute(readersHolder: ActorRef, nodeViewActorRef: ActorRef, e
...
+      addressesR ~
+      generatePaymentTransactionR ~
+      generateAssetIssueTransactionR ~
+      sendPaymentTransactionR ~
+      sendAssetIssueTransactionR
   }
 
-  def balancesRoute: Route = (path("balances") & get) {
+  def generatePaymentTransactionR: Route = (path( "payment" / "generate") & post
+    & entity(as[Seq[PaymentRequest]])) { generateTransaction }
```
mike-aksarin: maybe round braces looks better here?

src/main/resources/api/openapi.yaml
oskin1: It is supposed that each of these routes relates to transaction creation - the differences are the action that will be taken after (send transaction or not) and the actual type of transaction ("asset issue" or "payment")

src/main/scala/org/ergoplatform/nodeView/wallet/requests/AssetIssueRequest.scala
```diff
@@ -0,0 +1,69 @@
...
+import org.ergoplatform.settings.ErgoSettings
+import org.ergoplatform.{ErgoBox, ErgoBoxCandidate}
+import scorex.core.transaction.box.Box.Amount
+import sigmastate.Values.{IntConstant, StringConstant}
+
+/**
+  * Request for issuing new asset of Ergo Asset Standard 1.
+  *
+  * EAS1 token data is stored in registers in the following way:
+  * R3 - ID and supply amount
```
mike-aksarin: R2 probably meant here

src/main/scala/org/ergoplatform/nodeView/wallet/requests/AssetIssueRequest.scala
```diff
@@ -0,0 +1,69 @@
...
+import org.ergoplatform.ErgoBox.{R4, R5, R6}
+import org.ergoplatform.api.ApiCodecs
+import org.ergoplatform.nodeView.wallet.{ErgoAddress, ErgoAddressEncoder}
+import org.ergoplatform.settings.ErgoSettings
+import org.ergoplatform.{ErgoBox, ErgoBoxCandidate}
+import scorex.core.transaction.box.Box.Amount
+import sigmastate.Values.{IntConstant, StringConstant}
+
+/**
+  * Request for issuing new asset of Ergo Asset Standard 1.
```
mike-aksarin: Do we have any agreement about specification naming and numbering?

src/main/resources/api/openapi.yaml
oskin1: It is supposed that each of these routes relates to transaction creation - the differences are the action that will be taken after (send transaction or not) and the actual type of transaction ("asset issue" or "payment")

src/main/scala/org/ergoplatform/nodeView/wallet/requests/AssetIssueRequest.scala
oskin1: @mike-aksarin, No, I think any other option could be discussed. This name is just to make the point of point clear.

Issue #413, opened 2018-08-17
Change API errors content type to json


mike-aksarin:
Currently API provide `application/json` content type for successful responses and default `text/plain` for error responses. We need to unify response content type. This task depends on  similar Scorex task: ScorexFoundation/Scorex#304

1) Move Scorex API errors to `application/json` (ScorexFoundation/Scorex#304) and update Scorex version in Ergo
2) Override default akka http error handling.
3) Update application yaml file.

oskin1:
#491

PR #489, opened 2018-10-10
`application/json` error response content-type

oskin1:
#413


oskin1:
Backward compatibility is broken after update to latest Scores version, that is why ergo-it tests fail.

Issue #497, opened 2018-10-18
TransactionApiRoute improvements


oskin1:
- [ ] Perform stateless validation of transaction posted through `/transactions` and provide user with validity status in response.
- [ ] Introduce `offset` parameter in `/transaction/unconfirmed`

oskin1:
Closed #500

PR #500, opened 2018-10-18
TransactionApiRoutes improvements. Tx validation. offset parameter.

oskin1:
#497


src/main/scala/org/ergoplatform/api/TransactionsApiRoute.scala
```diff
@@ -23,18 +24,21 @@ case class TransactionsApiRoute(readersHolder: ActorRef, nodeViewActorRef: Actor
 
   private def getMemPool: Future[ErgoMemPoolReader] = (readersHolder ? GetReaders).mapTo[Readers].map(_.m)
 
-  private def getUnconfirmedTransactions(limit: Int): Future[Json] = getMemPool.map { p =>
-    p.take(limit).toSeq.map(_.asJson).asJson
+  private def getUnconfirmedTransactions(limit: Int, offset: Int): Future[Json] = getMemPool.map { p =>
```
mike-aksarin: please swap `limit` and `offset` params

PR #502, opened 2018-10-19
PoW update

catena2w:


catena2w:
Replaced by https://github.com/ergoplatform/ergo/pull/548

Issue #507, opened 2018-10-22
Box selector should ignore useless inputs.


oskin1:
Box selector should ignore boxes with empty Ergo balance for non-empty Ergo target.

catena2w:
After last changes boxes with empty ergo balance are not allowed.

PR #509, opened 2018-10-22
 Useless boxes filtering in box selector.

oskin1:


oskin1:
#507

catena2w:
After last changes there are not boxes with empty Ergo balance, I  think let's leave BoxSelector as is for now

PR #515, opened 2018-10-23
Fix assets test and transaction validation test after merging to 1.7

mike-aksarin:


src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -126,7 +126,7 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
         validation
           .demandEqualArrays(box.id, input.boxId, "Box id doesn't match input")
           .demandSuccess(costTry, s"Invalid transaction $this")
-          .demand(isCostValid, s"Validation failed for input #$idx of tx $this: $costTry")
+          .demand(isCostValid, s"Proposition proof failed for input #$idx of tx $this: $costTry")
```
oskin1: It seems to me that previous variant was correct

PR #523, opened 2018-10-28
History consistency after node restart spec.

### LINKED ISSUE #326: contains() and fullBlockOpt in History could be inconsistent after node restart
Parts of a block presented by the fullBlockOpt could be in the history (as contains() shows), which is making node stuck.

oskin1:
Close #326

### DIFF
--- a/src/it/scala/org/ergoplatform/it/HistoryConsistencySpec.scala
+++ b/src/it/scala/org/ergoplatform/it/HistoryConsistencySpec.scala
@@ -3,16 +3,21 @@ package org.ergoplatform.it
 import java.io.File
 
 import com.typesafe.config.Config
+import io.netty.util.{HashedWheelTimer, Timer}
+import org.asynchttpclient.util.HttpConstants
 import org.ergoplatform.it.container.{IntegrationSuite, Node}
+import org.ergoplatform.it.util._
 import org.scalatest.{FreeSpec, OptionValues}
 
-import scala.concurrent.Await
+import scala.concurrent.{Await, Future}
 import scala.concurrent.duration._
 
 class HistoryConsistencySpec
   extends FreeSpec
     with IntegrationSuite
     with OptionValues {
+
+  val timer: Timer = new HashedWheelTimer()
 
   val shutdownAtHeight: Int = 5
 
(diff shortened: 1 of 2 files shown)


src/it/scala/org/ergoplatform/it/api/NodeApi.scala
```diff
@@ -84,6 +90,37 @@ trait NodeApi {
...
+  def waitForInconsistentHistory: Future[Unit] = {
+    waitFor[Boolean](_.historyIsInconsistent, bool => bool, 25.millis).map(_ => ())
+  }
+
+  /** To find out whether history is inconsistent at this moment we should:
+    * 1. Fetch current `HistoryInfo`;
+    * 2. Check whether `FullBlock` with `bestHeaderId` from `HistoryInfo` is already available;
+    * 3. Check that `bestFullBlockId` isn't updated yet in `HistoryInfo`;
+    * */
+  def historyIsInconsistent: Future[Boolean] = {
```
catena2w: This is the part of concrete it test, we don't need it in node API.

src/it/scala/org/ergoplatform/it/api/NodeApi.scala
```diff
@@ -84,6 +90,37 @@ trait NodeApi {
...
 
+  def waitForHeight(expectedHeight: Int, retryingInterval: FiniteDuration = 1.second): Future[Int] = {
+    waitFor[Int](_.height, h => h >= expectedHeight, retryingInterval)
+  }
+
+  def waitForStartup: Future[this.type] = get("/info").map(_ => this)
+
+  /** Tries to catch the moment when full block is already persisted in history,
+    * but indexes aren't updated yet. */
+  def waitForInconsistentHistory: Future[Unit] = {
```
catena2w: This is the part of concrete it test, we don't need it in node API.

src/it/scala/org/ergoplatform/it/api/NodeApi.scala
```diff
@@ -150,4 +189,14 @@ object NodeApi extends ScorexLogging {
 
   case class Status(status: String)
 
+  case class HistoryInfo(bestHeaderId: String, bestBlockId: String, bestHeaderHeight: Int, bestBlockHeight: Int)
```
catena2w: we already have `ErgoStatsCollector#NodeInfo`, better to reuse it

src/it/scala/org/ergoplatform/it/HistoryConsistencySpec.scala
```diff
@@ -0,0 +1,51 @@
...
+      .flatMap { _ =>
+        docker.forceStopNode(node.containerId)
+        val restartedNode = docker
+          .startNode(nodeConfig, specialVolumeOpt = Some((localVolume, remoteVolume))).get
+        restartedNode.historyInfo
+          .flatMap(hi => restartedNode.waitForHeight(hi.bestBlockHeight + 1, 100.millis))
+          .flatMap(_ => restartedNode.historyInfo)
+          .flatMap { hi =>
+            restartedNode.headerIdsByHeight(hi.bestBlockHeight)
+              .map(_.headOption.value shouldEqual hi.bestBlockId)
```
catena2w: As you wrote in https://github.com/ergoplatform/ergo/issues/525, it is now valid situation, that bestFullBlock differs from bestHeader. It is not clear, what ar eyou actually testing here?

src/it/scala/org/ergoplatform/it/api/NodeApi.scala
oskin1: @catena2w, `ErgoStatsCollector#NodeInfo` contains too much redundant data, that should be fetched not only from API, but from settings as well, so its not conveniently to use it here.

src/it/scala/org/ergoplatform/it/HistoryConsistencySpec.scala
oskin1: This code tests the case when the node is shut down unexpectedly at the moment header and all its sections are stored but history indexes aren't updated yet - here we make sure that after such occasion history data is consistent after node restart.

oskin1: @catena2w, Test, used for i525 reproduction is different from this one, it exploits extremely short mining delay to create different blocks at the same height.

Issue #526, opened 2018-10-29
Implement and execute some non-trivial smart contract in Ergo testnet


catena2w:
For the beginning take some simple examples, like `if(height < 100) payToPk1 else payToPk2` and execute them in current testnet. Ensure we have everything for this (in Wallet, API, etc.).
If everything will be fine - I think some (possibly video) tutorial would be useful

oskin1:
I locked 2.0012 ERG with this script:
```
{
    val thirdPartyPk = PK("tJPvQhF2s5fFE1DNH8p2gd2Bkk8HAgFVSetvD1xCDVA6MayDCHbj76")
    atLeast(2, Array(thirdPartyPk, myPubKey_1, myPubKey_3))
}
```
And then successfully spent it (https://testnet.ergoplatform.com/en/transactions/0c4a06976a9c2692e407d37466adac7ce703b5aee19c4e63c92be88358370897)

PR #535, opened 2018-11-01
API methods for working with scripts

oskin1:


oskin1:
#526

src/main/scala/org/ergoplatform/api/WalletApiRoute.scala
```diff
@@ -38,16 +42,38 @@ case class WalletApiRoute(readersHolder: ActorRef, nodeViewActorRef: ActorRef, e
...
   }
 
+  private val source: Directive1[String] = entity(as[Json]).flatMap { p =>
+    p.hcursor.downField("source").as[String]
+      .fold(_ => reject, s => provide(s))
+  }
+
+  private def addressResponse(address: ErgoAddress): Json = Json.obj("address" -> addressJsonEncoder(address))
+
+  private def compileSource(source: String, publicKeys: Seq[ProveDlog]): Try[Value[SBoolean.type]] = {
```
ergomorphic: @oskin1 why not allow env: Map[String, Any] in the place of publicKeys? Also what about ProveDHT?

oskin1: @ergomorphic, the idea is only to load available in wallet public keys into env, so why do we really need `env: Map[String, Any]` here?

oskin1: @ergomorphic, ok, I've fixed that.

src/test/scala/org/ergoplatform/utils/Stubs.scala
```diff
@@ -134,12 +135,13 @@ trait Stubs extends ErgoGenerators with ErgoTestHelpers with ChainGenerator with
...
     private implicit val addressEncoder: ErgoAddressEncoder = new ErgoAddressEncoder(settings.chainSettings.addressPrefix)
     private val prover = new ErgoProvingInterpreter(seed, 2)
-    private val trackedAddresses: mutable.Buffer[ErgoAddress] =
-      mutable.Buffer(prover.dlogPubkeys: _ *).map(P2PKAddress.apply)
-
+    private val trackedAddresses: Seq[P2PKAddress] = prover.dlogPubkeys.map(P2PKAddress.apply)
 
     def receive: Receive = {
 
+      case ReadPublicKeys(from, until) =>
```
catena2w: It is not used anywhere

oskin1: @catena2w, it is required for tests as wallet route now read public keys from wallet, so we need stub to mock wallet actor response.
https://github.com/ergoplatform/ergo/blob/1db706f65805cac501f0464fe8674670e83a46ef/src/main/scala/org/ergoplatform/api/WalletApiRoute.scala#L144

Issue #530, opened 2018-10-31
Different synchronization regimes


catena2w:
We have a lot of possible node regimes, we need to check, that they all works. Every subtask here requires to write it test for concrete regime and implement missing functionality. Separate tasks may be created for subtasks if needed.
- [x] Full-node regime. Should be implemented and tested.
- [x] Digest state with `blocksToKeep>0`. Should download all headers, initialize digest state at `headersHeight - blocksToKeep`,  download last `blocksToKeep` full blocks. Was implemented at testnet1.5, tests are needed. 
- [x] Utxo state with `blocksToKeep>0`. Same as a previous one, but with Utxo State. To initialize UTXO state, `UtxoSnapshot` is needed. This is not implemented in Ergo, however AVL tree serialization/deserialization is already implemented in scrypto (see https://github.com/input-output-hk/scrypto/blob/master/src/test/scala/scorex/crypto/authds/avltree/batch/serialization/AVLBatchSerializationSpecification.scala#L35).
- [x] Digest state with `PoPoWBootstrap = true` and `blocksToKeep>0`. Should download and check PoPoW proof (from multiple nodes), initialize Digest state at `headersHeight - min(blocksToKeep, popowProof.suffix.length)`, download remaining full blocks. https://github.com/ergoplatform/ergo/issues/211 should be implemented first
-  [x] Utxo state with `PoPoWBootstrap = true` and `blocksToKeep>0`. Same as a previous one, but with Utxo State.

PR #543, opened 2018-11-06
Digest sate node sync spec.

oskin1:


oskin1:
#530

src/it/scala/org/ergoplatform/it/PrunedDigestNodeSyncSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+      val targetHeight = sampleInfo.bestBlockHeightOpt.value
+      Async.await(digestNode.waitForHeight(targetHeight))
+      val digestNodeInfo = Async.await(digestNode.info)
+      digestNodeInfo shouldEqual sampleInfo
+      val digestNodePrunedBlockId = Async.await(digestNode.headerIdsByHeight(targetHeight - blocksToKeep - 1)).head
+      Async.await(digestNode.singleGet(s"/blocks/$digestNodePrunedBlockId")
+        .map(_.getStatusCode == HttpConstants.ResponseStatusCodes.OK_200)) shouldBe false
+    }
+
+    Await.result(result, 10.minutes)
```
catena2w: Why `Await.result` here instead of `Async.await`?

src/it/scala/org/ergoplatform/it/PrunedDigestNodeSyncSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+  val minerConfig: Config = nodeSeedConfigs.head
+    .withFallback(miningDelayConfig(10000))
+    .withFallback(specialDataDirConfig(remoteVolume))
+  val digestConfig: Config = digestStatePeerConfig
+    .withFallback(blockIntervalConfig(8000))
+    .withFallback(prunedHistoryConfig(blocksToKeep))
+    .withFallback(nonGeneratingPeerConfig)
+    .withFallback(nodeSeedConfigs(1))
+
+  // Testing scenario:
```
catena2w: This test scenario actually does not check, that node did not applied full blocks with `height < {targetHeight - blocksToKeep}`. So it is still possible, that node have downloaded all the blocks and removed blocks at `0-{targetHeight - blocksToKeep}` after that.
Test should check that Digest node initialized at block `{targetHeight - blocksToKeep}` without applying first blocks,

oskin1: @catena2w, checks added

src/it/scala/org/ergoplatform/it/PrunedDigestNodeSyncSpec.scala
oskin1: @catena2w , `Async.await` does not allow to specify timeout.

PR #548, opened 2018-11-08
Update PoW

catena2w:
New PoW algorithm implementation


src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -221,15 +167,17 @@ class ErgoMinerSpec extends FlatSpec with ErgoTestHelpers with ValidBlocksGenera
...
 
-    val height = r.h.bestFullBlockOpt.map(_.header.height).getOrElse(startHeight)
-    val boxToDoubleSpend: ErgoBox = r.h.bestFullBlockOpt.get.transactions.last.outputs.last
-    val input = Input(boxToDoubleSpend.id, ProverResult(Array.emptyByteArray, ContextExtension.empty))
+    val boxToDoubleSpend = r.h.bestFullBlockOpt.get.transactions.last.outputs.last
+    boxToDoubleSpend.proposition shouldBe defaultMinerSecret.publicImage
 
-    val outputs1 = IndexedSeq(new ErgoBoxCandidate(boxToDoubleSpend.value, prop1, creationHeight = height))
-    val tx1 = ErgoTransaction(IndexedSeq(input), outputs1)
+    val input = Input(boxToDoubleSpend.id, ProverResult(Array.emptyByteArray, ContextExtension.empty))
```
mike-aksarin: We also `have emptyProverResult` in `WalletTestOps`. We can move it to `NodeViewBaseOps`

catena2w:
@mike-aksarin fixed your comments except of the paper. The paper will be significantly redone a bit later.

papers/yellow/pow/ErgoPow.tex
```diff
@@ -0,0 +1,229 @@
...
+            \Let{$d$}{$(\sum_{j \in J}{H(i||M||pk||m||w)} \cdot x - sk).mod(q)$}
+            \If{$d <= b$}
+            \State \Return $(m,pk,w,nonce,d)$
+            \EndIf
+            \EndWhile
+        \end{algorithmic}
+    \end{algorithm}
+
+    Note that although the mining process utilizes private keys, solution itself
+    only contain public keys. Solution verification may be done by Alg.~\ref{alg:verify}.
```
oskin1: contains

papers/yellow/pow/ErgoPow.tex
```diff
@@ -0,0 +1,229 @@
...
+    %    to find an appropriate index set.
+    %
+    %    Protocol is quite efficient: in addition to payload, block header should
+    %    contain 2 public keys of size 32 bytes, number $d$ that is at most 32 bytes
+    %    (but will contain a lot of leading zeros in case of small target $b$) and a nonce,
+    %    which size might be small enough (say, 8 bytes). Thus for about 100 additional bytes
+    %    will be added to the block header.
+
+    %    Difficulty $h$ adjusts to the network computational power in such a way,
+    %    that correct solutions appears roughly once per 2 minutes. Non-optimized Scala
```
oskin1: appear

src/it/resources/parameters-template.txt
```diff
@@ -40,7 +40,7 @@ paths:
             "adProofsRoot": "a84f62a669fb3684308ea609af6cd831b939b70210307e143a05321ce8efeda2",
             "transactionsRoot": "cc32add0ada11b6a81f07b61a6c606b2277af9999b453538d32b3c409630bce1",
             "extensionHash": "0e5751c026e543b2e8ab2eb06099daa1d1e5df47778f7787faab45cdf12fe3a8",
-            "equihashSolutions": "00000914000125c300013c440001c51e00000fbb00002a960000edc40001baee000065d500006ff90000e43f00011ca500016f270001781d00017b150001d54d0000493500012c19000171eb000184f50000720d0000e0f80000bcd9000181bc00004dc40000ef250000d09300011c540000f1190001d6f60000ffaf0001104f",
+            "powSolutions": "???",
```
oskin1: Let's put it here instead of `???`
```
"powSolutions" : {
    "pk" : "0350e25cee8562697d55275c96bb01b34228f9bd68fd9933f2a25ff195526864f5",
    "w" : "032e3f5edb88f3cc7384bfedc892fc8dcb10d7a3bf3741d08a62cc701848d0932c",
    "n" : "0000000000000000",
    "d" : 603083798111021851164432213586916186738093029948325633833495911523787854249
},
```

src/main/scala/org/ergoplatform/mining/DefaultFakePowScheme.scala
```diff
@@ -0,0 +1,41 @@
...
+  * Fake Pow Scheme for tests.
+  * Fill solution with random values, all blocks are valid during validation
+  */
+object DefaultFakePowScheme extends AutolykosPowScheme(1, 1) {
+  override def validate(header: Header): Try[Unit] = Success(Unit)
+
+  override def prove(parentOpt: Option[Header],
+                     nBits: Long,
+                     stateRoot:
+                     ADDigest,
```
oskin1: Why break line here?

src/main/scala/org/ergoplatform/mining/AutolykosPowScheme.scala
```diff
@@ -0,0 +1,212 @@
...
+    (parentId, version, interlinks, height)
+  }
+
+  /**
+    * Hash function that takes `m` and `nonceBytes` and returns a list of size `k` with numbers in
+    * [0,`N`)
+    */
+  private def genIndexes(seed: Array[Byte]): Seq[Int] = {
+    val hash = Blake2b512(Bytes.concat(seed))
+    hash.grouped(3).take(k).toSeq.map(b => Math.abs(Ints.fromByteArray(0.toByte +: b) % N))
```
kushti: 1. Math.abs is not needed, your integer is always positive by construction
2. 3 bytes is about 16.7M, so N = 10^8 = 100M does not make any sense.
3. There is no need to use Blake2b512. If you need for 512 bits, you can use two calls with different prefixes (e.g. 0 and 1), it will be about similar performance, but no additional primitive needed. But I think it could be enough to use 256 bits, so bytes (0-2) for the first index, bytes (1-3) for the second etc.

src/main/scala/org/ergoplatform/mining/mining.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  type PrivateKey = BigInt
+
+  val PublicKeyLength: Byte = 33
+
+  val group: BcDlogFp[EcPointType] = CryptoConstants.dlogGroup
+
+  val q: BigInt = group.q
+
+  private val hashFn: NumericHash = new NumericHash(q)
```
kushti: Blake2b512 is not needed, in case 25519 q ~= 252 bits, so you can just set four most significant bits of 256-bits hash to zero.

src/main/scala/org/ergoplatform/mining/AutolykosPowScheme.scala
```diff
@@ -0,0 +1,212 @@
...
+    val s = header.powSolution
+
+    require(s.d < b || s.d > (q - b), s"Incorrect d=${s.d} for b=$b")
+    require(s.pk.getCurve == group.curve && !s.pk.isInfinity, "pk is incorrect")
+    require(s.w.getCurve == group.curve && !s.w.isInfinity, "w is incorrect")
+
+    val p1 = pkToBytes(s.pk)
+    val p2 = pkToBytes(s.w)
+    val f = genIndexes(Bytes.concat(msg, s.n)).map(ib => genElement(msg, p1, p2, Ints.toByteArray(ib))).sum.mod(q)
+    val left = s.w.multiply(f.bigInteger).add(s.pk.negate())
```
kushti: better to add s.pk to the right thus saving one negate operation

src/main/scala/org/ergoplatform/mining/AutolykosSolution.scala
```diff
@@ -0,0 +1,72 @@
...
+  * Solution of Autolykos PoW puzzle
+  *
+  * @param pk - miner public key. Should be used to collect block rewards
+  * @param w  - one-time public key. Prevents revealing of miners secret
+  * @param n  - nonce
+  * @param d  - distance between pseudo-random number, corresponding to nonce `n` and a secret,
+  *           corresponding to `pk`. The lower `d` is, the harder it was to find this solution.
+  */
+case class AutolykosSolution(pk: ECPoint, w: ECPoint, n: Array[Byte], d: BigInt) extends BytesSerializable {
+  assert(!pk.isInfinity && !w.isInfinity, s"Infinity points are not allowed ${pk.isInfinity}, ${w.isInfinity}")
```
kushti: It is possible to shut down any node in the network now by generating a header with an infinity point.

Assertions should not be without catching in objects we can get over network (and not prevously checked).

A possible solution is to check infinity point in deserialization as well, or avoid assertion here (thus relying only on validate(_: Header)).

src/main/scala/org/ergoplatform/modifiers/history/Header.scala
```diff
@@ -202,7 +175,7 @@ object HeaderSerializer extends Serializer[Header] {
 
     @tailrec
     def parseInterlinks(index: Int, endIndex: Int, acc: Seq[ModifierId]): Seq[ModifierId] = if (endIndex > index) {
-      val repeatN: Int = bytes.slice(index, index + 1).head
+      val repeatN: Int = 0xff & bytes(index)
```
kushti: why not just bytes(index) ?

src/main/scala/org/ergoplatform/mining/AutolykosSolution.scala
catena2w: It is not possible to shut down a node, as soon as we already checks this during deserialization -infinity points are encoded in 1 byte, while we take PublicKeyLength.
Removed this assertion anyway

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -225,51 +236,55 @@ class ErgoMiner(ergoSettings: ErgoSettings,
...
+  def collectRewards(emissionBoxOpt: Option[ErgoBox],
                      currentHeight: Int,
-                     feeBoxes: Seq[ErgoBox],
-                     minerProp: Value[SBoolean.type],
-                     emission: EmissionRules): ErgoTransaction = {
-    feeBoxes.foreach(b => assert(b.proposition == TrueLeaf, s"Trying to create coinbase from protected fee box $b"))
+                     txs: Seq[ErgoTransaction],
+                     minerPk: ProveDlog,
+                     emission: EmissionRules): Seq[ErgoTransaction] = {
+    val feeBoxes: Seq[ErgoBox] = ErgoState.boxChanges(txs)._2.filter(_.proposition == Constants.FeeProposition)
```
kushti: Better to compare _.propositionBytes (the comparison would be faster)

catena2w: Is it always faster? If yes - I propose to redefine `equals` in `sigmastate.Values.Value`.

src/main/scala/org/ergoplatform/mining/mining.scala
catena2w: replaced underlying hash to Blake2b256.

src/main/scala/org/ergoplatform/modifiers/history/Header.scala
catena2w: `bytes(index)` will return signed byte, and this function will go to infinite loop if someone will eventually find a block with more then 127 leading zero bits.

src/main/scala/org/ergoplatform/mining/AutolykosPowScheme.scala
catena2w: fixed

src/main/scala/org/ergoplatform/modifiers/history/Header.scala
kushti: require(repeatN > 0) is not needed then, as your repeatN is non-negative by construction, or this line to be changed to just require(repeatN != 0)

src/main/scala/org/ergoplatform/mining/AutolykosPowScheme.scala
```diff
@@ -0,0 +1,220 @@
...
+  * Mining process is implemented in inefficient way and should not be used in real environment.
+  *
+  * @see papers/yellow/pow/ErgoPow.tex for full description
+  * @param k - number of elements in one solution
+  * @param n - power of number of elements in a list
+  */
+class AutolykosPowScheme(val k: Int, val n: Int) extends ScorexLogging {
+
+  assert(k <= 32, "k > 32 is not allowed due to genIndexes function")
+  assert(n < 32, "n >= 32 is not allowed")
```
kushti: As max value for an object of type Int is 2^31-1, the proper limit is n < 31

src/main/scala/org/ergoplatform/mining/mining.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  def hash(in: Array[Byte]): BigInt = hashFn.hash(in)
+
+  def genPk(s: PrivateKey): ECPoint = group.exponentiate(group.generator, s.bigInteger)
+
+  def randomSecret(): PrivateKey = DLogProverInput.random().w
+
+  def lg(x: Int): Int = (Math.log(x) / Math.log(2)).toInt.ensuring(s => Math.pow(2, s) == x)
```
kushti: not used

Issue #251, opened 2018-04-23
Integration test for the node with Digest state synchronization


mike-aksarin:
Add integration test when one of the nodes has state="digest" and check how it synchronizes

oskin1:
Closed #543

PR #549, opened 2018-11-08
Pruned full node synchronization specification

oskin1:
#251 
#530 (3rd item)


oskin1:
Depends on https://github.com/ergoplatform/ergo/pull/543

PR #574, opened 2018-12-19
Chain generator fixed

### LINKED ISSUE #573: Fix ChainGenerator
`org.ergoplatform.tools.ChainGenerator` does not work after new PoW implementation, because transaction generator from tests generates incorrect transactions (it tries to spend transactions from emission box in the same block.
ChainGenerator should generate correct blockchain with generated transactions with genesis block at `startTimestamp`.

oskin1:
Close #573


oskin1:
ChainGenerator is out of date, so I propose to re-implement it using real node components (miner, txGen, NVH) in order to avoid copying logic form main app.

oskin1:
Duplicates #608

PR #586, opened 2018-12-26
Update sigmastate (Height is Int)

greenhat:
References:
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/340
https://github.com/ScorexFoundation/sigmastate-interpreter/issues/335

Todo:
- [x] switch `box.creationInfo._1` to Int (height when block got included) in sigmastate and remove `Downcast`;
- [x] find and fix one extra (compared to `voting`) failing test;
- [x] switch to sigmastate's snapshot from `master`;

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
@@ -17,7 +17,7 @@ import scorex.crypto.authds.{ADDigest, ADKey}
 import scorex.util.encode.Base16
 import scorex.util.{ModifierId, ScorexLogging, bytesToId}
 import sigmastate.SCollection.SByteArray
-import sigmastate.Values.{IntArrayConstant, IntConstant, LongConstant, SigmaPropValue, Value}
+import sigmastate.Values.{IntArrayConstant, IntConstant, SigmaPropValue, Value}
 import sigmastate.{Values, _}
 import sigmastate.lang.Terms._
 import sigmastate.serialization.ErgoTreeSerializer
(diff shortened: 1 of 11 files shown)


src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -105,21 +108,21 @@ object ErgoState extends ScorexLogging {
     val rewardOut = ByIndex(Outputs, IntConstant(0))
     val minerOut = ByIndex(Outputs, IntConstant(1))
 
-    val epoch = Plus(LongConstant(1), Divide(Minus(Height, LongConstant(s.fixedRatePeriod)), LongConstant(s.epochLength)))
-    val coinsToIssue = If(LT(Height, LongConstant(s.fixedRatePeriod)),
+    val epoch = Plus(IntConstant(1), Divide(Minus(Height, IntConstant(s.fixedRatePeriod.toInt)), IntConstant(s.epochLength)))
+    val coinsToIssue = If(LT(Height, IntConstant(s.fixedRatePeriod.toInt)),
       s.fixedRate,
-      Minus(s.fixedRate, Multiply(s.oneEpochReduction, epoch))
+      Minus(s.fixedRate.toInt, Multiply(s.oneEpochReduction.toInt, epoch))
```
catena2w: `fixedRate` and `oneEpochReduction` should remain Long

ergomorphic: @catena2w you mean the operations should be performed with long values? Because header is now Int.

catena2w: height is in, but this does not affect fixedRate and oneEpochReduction, they are still long

greenhat: @ergomorphic @catena2w Done. I've upcasted `epoch` to Long.

greenhat:
@kushti I've rebased and updated genesis state hash for it:test. Only two of them are failing, but they're the same two tests that also fail in `voting` branch.

PR #600, opened 2019-01-16
Foundation box

### LINKED ISSUE #559: Founders box
Implement the founders reward box and a corresponding script.
- Update script of emissionBox, such that miners reward is 67.5 when it is higher now
- Add foundersBox to genesis state, that allows to take the remaining part of the emission box (so total emission does not change)
- Besides of rules, describing amount of Ergo, allowed to take at corresponding height, coins should also be protected by additional rules, described below
- I propose to issue a special token X (e.g. in quantity 100) and add a section to protection script, that allow to take the reward, only if at least 51% of this token is  involved in the transaction (this will require to create the third box to genesis state with this token)
- Script should also allow to change this additional rule, if 51% of token holders are involved. It is better to implement branching condition with MAST.

catena2w:
Close #559.
Relies on unmerged https://github.com/ScorexFoundation/sigmastate-interpreter/pull/360 PR


src/main/scala/org/ergoplatform/nodeView/state/StateConstants.scala
```diff
@@ -13,9 +13,9 @@ import scorex.crypto.authds.ADDigest
   * @param settings          - node settings
   */
 case class StateConstants(nodeViewHolderRef: Option[ActorRef], settings: ErgoSettings) {
-  lazy val emission: EmissionRules = settings.emission
-  lazy val genesisEmissionBox: ErgoBox = ErgoState.genesisEmissionBox(emission)
+  lazy val emission: EmissionRules = settings.chainSettings.emission
+  lazy val genesisEmissionBox: ErgoBox = ErgoState.genesisBoxes(settings.chainSettings).head
```
aslesarenko: @catena2w looks like we creating `genesisBoxes` more than once. Here and in UTXO genesis state.

src/main/scala/org/ergoplatform/settings/ChainSettings.scala
```diff
@@ -17,4 +22,16 @@ case class ChainSettings(protocolVersion: Byte,
...
+                         foundersPubkeys: Seq[String],
+                         genesisStateDigestHex: String,
+                         genesisId: Option[ModifierId] = None) {
+
+  val genesisStateDigest: ADDigest = Base16.decode(genesisStateDigestHex) match {
+    case Success(b) => ADDigest @@ b
+    case _ => throw new Error(s"Failed to parse genesisStateDigestHex = $genesisStateDigestHex")
+  }
+
+  val emission = new EmissionRules(monetary)
```
aslesarenko: Suggest better name `emissionRules`

src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
```diff
@@ -3,23 +3,87 @@ package org.ergoplatform.nodeView.state
...
+      .map(str => pkFromBytes(Base16.decode(str).get))
+      .map(pk => ProveDlog(pk))
+    settingsPks.count(p => defaultProver.dlogPubkeys.contains(p)) shouldBe 2
+
+    forAll(defaultHeaderGen) { header =>
+      val rewardPk = (new DLogProverInput(BigInt(header.height).bigInteger)).publicImage
+
+      val t = validTransactionsFromBoxHolder(bh, new Random(height))
+      val txs = t._1
+      bh = t._2
```
aslesarenko: @catena2w What is this?

catena2w: Is there a better way to do this? `validTransactionsFromBoxHolder` return tuple, we should update `bh` variable to the second element of the tuple

src/main/scala/org/ergoplatform/nodeView/state/StateConstants.scala
catena2w: fixed

src/main/scala/org/ergoplatform/settings/ChainSettings.scala
catena2w: fixed

PR #606, opened 2019-01-18
ErgoMiner refactoring.

oskin1:


src/main/scala/org/ergoplatform/modifiers/history/Extension.scala
```diff
@@ -44,6 +44,11 @@ case class ExtensionCandidate(fields: Seq[(Array[Byte], Array[Byte])]) {
   def toExtension(headerId: ModifierId): Extension = Extension(headerId, fields)
 }
 
+object ExtensionCandidate {
+  // todo fill with interlinks and other useful values after nodes update
+  def empty: ExtensionCandidate = ExtensionCandidate(Seq.empty)
```
catena2w: we won't have empty candidates after interlinks moved to it, so I don't think the method make sense

oskin1: @catena2w, fixed

PR #608, opened 2019-01-18
ChainGenerator fixed.

### LINKED ISSUE #573: Fix ChainGenerator

oskin1:
Close #573


oskin1:
> Approved, but I'd like to see performance results - how much will it take to generate 50k blocks?

~ 107 min

PR #615, opened 2019-01-27
Update for fix-i372 in sigmastate

aslesarenko:
[#372 Related sigma issue](https://github.com/ScorexFoundation/sigmastate-interpreter/issues/372)


aslesarenko:
These changes are already in v2.0

PR #616, opened 2019-01-28
External miner interface

### LINKED ISSUE #612: API for external miner
- Replace `/blocks/candidateBlock` with API method that return `m := Blake2b256(hdr.bytesWithoutPow)` and public key.
- Add method, that receives solution (w, nonce, d), construct block with all the data kept inside ErgoMiner and sends it to NodeViewHolder

oskin1:
Close #612


catena2w:
Switched status to wip. Let's wait until GPU miner implementation and merge after that

src/main/resources/application.conf
```diff
@@ -25,6 +25,12 @@ ergo {
     # Is the node is doing mining
     mining = false
 
+    # Use external miner, native miner is used if set to `false`
+    useExternalMiner = false
```
catena2w: Let's have external miner by default

src/main/resources/application.conf
```diff
@@ -25,6 +25,12 @@ ergo {
...
+
+    # Public key mining rewards will be dedicated to
+    miningPubKeyHex = null
```
catena2w: Can you do here pubkey or P2PK address (both should work)? I think it's more convenient for users to think in terms of addresses

oskin1: Done

PR #626, opened 2019-02-07
Switching to the Secp256k1 curve 

kushti:


src/main/scala/org/ergoplatform/mining/mining.scala
```diff
@@ -24,8 +25,7 @@ package object mining {
 
   def randomSecret(): PrivateKey = DLogProverInput.random().w
 
-  def pkToBytes(pk: EcPointType): Array[Byte] = GroupElementSerializer.toBytes(pk)
+  def groupElemToBytes(pk: EcPointType): Array[Byte] = GroupElementSerializer.toBytes(pk)
```
oskin1: `pk` is not meaningful parameter name any more - maybe `ge` or `elem` would be more suitable?

kushti: fixed!

PR #628, opened 2019-02-12
Interlinks moved to extension

### LINKED ISSUE #452: Move interlinks to extension

oskin1:
Close #452


src/main/scala/org/ergoplatform/modifiers/history/PoPowAlgos.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+    * Unpacks interlinks from key-value format of extension.
+    */
+  @inline def unpackInterlinks(fields: Seq[(Array[Byte], Array[Byte])]): Seq[ModifierId] = {
+    fields
+      .filter(_._1.headOption.contains(InterlinksFieldsPrefix))
+      .foldLeft(Seq.empty[ModifierId]) { case (acc, (_, v)) =>
+        assert(v.lengthCompare(Constants.ModifierIdSize + 1) == 0)
```
kushti: Using assert (again, better require to avoid confusions) with no catching to catch events which should never happen (if the implementation is done properly) is okay, however, when external data is being parsed:
1. All the checks should be non-elidable
2. Possible faults should be clearly visible to the function callers, i.e. the function should return Try[...]

Also, please write a test for parsing wrong interlinks bytes.

PR #643, opened 2019-03-01
Add data inputs to transaction

### LINKED ISSUE #585: Library inputs
Add a new sequence of inputs to transaction, that will be available from Ergo Script, but won't be spent and their proposition won't be spent

catena2w:
close #585


catena2w:
Ready, but depends on unmerged https://github.com/ScorexFoundation/sigmastate-interpreter/pull/411

PR #655, opened 2019-03-14
Abort earlier during too costly block or transaction validation, output asset value could be positive only

kushti:
This PR contains one consensus-critical changes:
 - it is required for all the outputs now to have positive value for each asset (currently, it is enough to have non-negative only).

Also, now the client is aborting earlier if a transaction or a block is too costly:
1. If transaction is too costly before even looking into its assets (so it has too many inputs, data inputs, or outputs), verification aborts immediately.
2. If transaction with assets is too costly, verification aborts immediately after checking the assets.
3. If cumulative transaction verification cost becomes too big after an input, verification aborts immediately after revealing the fact.
4. A transaction validation is now receiving accumulated cost of previous transaction, and adding cost of the transaction to this base. Thus block validation potentially aborts earlier.

Please note that currently transaction validation is not being aborted if its cost is too big, it is possible to get Success(cost), where cost is above limit, as a result.  This potentially leads to DDoS attacks.


src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -109,60 +112,80 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
-        costTry.recover { case t => t.printStackTrace() }
-
-        lazy val (isCostValid, scriptCost) = costTry.getOrElse((false, 0L))
-        validation
-          .demandEqualArrays(box.id, input.boxId, "Box id doesn't match input")
-          .demandSuccess(costTry, s"Invalid transaction $this")
-          .demand(isCostValid, s"Input script verification failed for input #$idx ($box) of tx $this: $costTry")
-          .map(_ + scriptCost)
-      }
+      // Check that there are no overflow in input and output values
```
oskin1:
```suggestion
      // Check that there is no overflow in input and output values
```

PR #668, opened 2019-03-22
Benchs to InfluxDB

andyceo:
- Commit interim work


catena2w:
Looks like won't implement at least right now

PR #679, opened 2019-04-01
Update after sigma refactoring

aslesarenko:


src/main/scala/org/ergoplatform/api/ApiCodecs.scala
```diff
@@ -151,12 +145,16 @@ trait ApiCodecs {
     )
   }
 
+//  implicit val byteCollEncoder: Encoder[Coll[Byte]] = _.toArray.asJson
```
catena2w: please don't leave commented code

src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
```diff
@@ -57,7 +58,7 @@ trait UtxoStateReader extends ErgoStateReader with TransactionValidation[ErgoTra
   protected[state] def extractEmissionBox(fb: ErgoFullBlock): Option[ErgoBox] = emissionBoxIdOpt match {
     case Some(id) =>
       fb.blockTransactions.txs.view.reverse.find(_.inputs.exists(t => java.util.Arrays.equals(t.boxId, id))) match {
-        case Some(tx) if tx.outputs.head.proposition == constants.settings.chainSettings.monetary.emissionBoxProposition =>
+        case Some(tx) if tx.outputs.head.proposition == constants.settings.chainSettings.monetary.emissionBoxProposition.proposition.asBoolValue =>
```
catena2w: Isn't it better to compare ergoTree here instead of propositions?

src/main/scala/org/ergoplatform/api/ApiCodecs.scala
greenhat: Fixed.

src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
greenhat: Done in https://github.com/ergoplatform/ergo/pull/679/commits/4811c43221c03ebc2b0e6be9d7cae539dfbca635 . Although two ergo trees having the same proposition but different serialization options (like constant segregation) would not be equal. Overall I think it's worth switching to ergo tree comparison.

greenhat:
@catena2w In order to switch to sigma-state from `v2.1` branch https://github.com/ScorexFoundation/sigmastate-interpreter/pull/459 have to be merged first (all green and ready).

PR #681, opened 2019-04-01
Strict type checks in ErgoTreeSerializer

greenhat:
Using sigma from PR https://github.com/ScorexFoundation/sigmastate-interpreter/pull/459

### DIFF
--- a/src/main/scala/org/ergoplatform/api/ApiCodecs.scala
+++ b/src/main/scala/org/ergoplatform/api/ApiCodecs.scala
@@ -96,26 +96,17 @@ trait ApiCodecs {
     decodeErgoTree(_.asInstanceOf[ErgoTree])
   }
 
-  // TODO catena: this should be removed (Value should not be used directly)
-  implicit val valueEncoder: Encoder[Value[SType]] = { value =>
+  implicit val evaluatedValueEncoder: Encoder[EvaluatedValue[SType]] = { value =>
     ValueSerializer.serialize(value).asJson
   }
 
-  implicit val booleanValueEncoder: Encoder[Value[SBoolean.type]] = { value =>
-    valueEncoder(value)
-  }
-
-  implicit val booleanValueDecoder: Decoder[Value[SBoolean.type]] = {
-    valueDecoder(_.asInstanceOf[Value[SBoolean.type]])
-  }
-
   implicit val evaluatedValueDecoder: Decoder[EvaluatedValue[SType]] = {
-    valueDecoder(_.asInstanceOf[EvaluatedValue[SType]])
+    decodeEvaluatedValue(_.asInstanceOf[EvaluatedValue[SType]])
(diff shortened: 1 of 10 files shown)


src/main/scala/org/ergoplatform/api/ApiCodecs.scala
```diff
@@ -98,7 +98,7 @@ trait ApiCodecs {
 
   // TODO catena: this should be removed (Value should not be used directly)
   implicit val valueEncoder: Encoder[Value[SType]] = { value =>
-    ErgoTreeSerializer.DefaultSerializer.serializeWithSegregation(value).asJson
+    ValueSerializer.serialize(value).asJson
```
aslesarenko: @greenhat Isn't it a step back? From explicit ErgoTree serialization back to implicit value serialization.

src/test/scala/org/ergoplatform/mining/ErgoMinerPropSpec.scala
```diff
@@ -18,7 +18,7 @@ class ErgoMinerPropSpec extends ErgoPropertyTest {
   val delta: Int = settings.chainSettings.monetary.minerRewardDelay
 
   private def expectedRewardOutputScriptBytes(pk: ProveDlog): Array[Byte] =
-    ErgoScriptPredef.rewardOutputScript(delta, pk).bytes
+    ErgoScriptPredef.rewardOutputScript(delta, pk).treeWithSegregation.bytes
```
aslesarenko: @greenhat rewardOutputScript should return ErgoTree.

greenhat: Done.

src/main/scala/org/ergoplatform/api/ApiCodecs.scala
greenhat: I don't think so. `valueEncoder` used extensively (additional registers encoder, context extension encoder, in tx's `toString`, etc.). I see no way to use `ErgoTreeSerializer` since it requires SigmaProp values. Previously it did cast any value as SigmaProp. I've tried to remove `valueEncoder`completely and got into the "rabbit hole" quickly.

aslesarenko: 1) additional registers contain only constants
2) context extensions also contain only constants
Usage of Value[SType] here basically means "I don't know which values this can be, so let it be any possible value.".
It is better to have more concrete encoders (actually only Constants is needed) than single universal encoder, that just hide all potential problems.

greenhat: Done in https://github.com/ergoplatform/ergo/pull/681/commits/240cd33d25a23771323d59019f41c3cdd38ac4cf

PR #687, opened 2019-04-02
Run as unprivileged user

sergelogvinov:
Create user "ergo" and use it to run daemon.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,9 +17,13 @@ CMD ["/usr/bin/java", "-jar", "/ergo.jar"]
 
 FROM openjdk:10-jre-slim
 LABEL maintainer="Andrey Andreev <andyceo@yandex.ru> (@andyceo)"
-COPY --from=builder /ergo.jar /ergo.jar
+RUN adduser --disabled-password --home /var/lib/ergo --uid 9052 --gecos "ErgoPlatform" ergo && \
+    install -m 0755 -o ergo -g ergo -d /var/cache/ergo /var/cache/ergo/data
+COPY --from=builder /ergo.jar /var/lib/ergo/ergo.jar
+USER ergo
 EXPOSE 9007 9052
-WORKDIR /root
-VOLUME ["/root/ergo/data"]
-ENTRYPOINT ["/usr/bin/java", "-jar", "/ergo.jar"]
+WORKDIR /var/cache/ergo
+VOLUME ["/var/cache/ergo/data"]
+ENV DATADIR=/var/cache/ergo/data
+ENTRYPOINT ["/usr/bin/java", "-jar", "/var/lib/ergo/ergo.jar"]
 CMD [""]
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1,6 +1,7 @@
 ergo {
   # Directory to keep data
   directory = ${user.dir}"/ergo/data"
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -17,9 +17,13 @@ CMD ["/usr/bin/java", "-jar", "/ergo.jar"]
 
 FROM openjdk:10-jre-slim
 LABEL maintainer="Andrey Andreev <andyceo@yandex.ru> (@andyceo)"
-COPY --from=builder /ergo.jar /ergo.jar
+RUN adduser --disabled-password --home /var/lib/ergo --uid 9052 --gecos "ErgoPlatform" ergo && \
```
andyceo: Home dir /var/lib/ergo not used further. Think better use folderless user, but this behaviour require node config changes

Dockerfile
```diff
@@ -17,9 +17,13 @@ CMD ["/usr/bin/java", "-jar", "/ergo.jar"]
...
+    install -m 0755 -o ergo -g ergo -d /var/cache/ergo /var/cache/ergo/data
+COPY --from=builder /ergo.jar /var/lib/ergo/ergo.jar
+USER ergo
 EXPOSE 9007 9052
```
andyceo:
```suggestion
EXPOSE 9020 9052
```

andyceo: Not related to your PR but better be fixed

Dockerfile
```diff
@@ -17,9 +17,13 @@ CMD ["/usr/bin/java", "-jar", "/ergo.jar"]
...
-WORKDIR /root
-VOLUME ["/root/ergo/data"]
-ENTRYPOINT ["/usr/bin/java", "-jar", "/ergo.jar"]
+WORKDIR /var/cache/ergo
+VOLUME ["/var/cache/ergo/data"]
```
andyceo:
```suggestion
VOLUME ["/var/lib/ergo"]
VOLUME ["/var/cache/ergo/data"]
```

andyceo: May be user folder need to be in volumes too.

src/main/resources/application.conf
```diff
@@ -1,6 +1,7 @@
 ergo {
   # Directory to keep data
   directory = ${user.dir}"/ergo/data"
+  directory = ${?DATADIR}
```
andyceo: Not sure if previous `directory = ${user.dir}"/ergo/data"` overwrited to empty string if DATADIR not set

src/main/resources/application.conf
```diff
@@ -185,6 +186,7 @@ scorex {
     maxInvObjects = 400
     desiredInvObjects = 400
     nodeName = "ergo-testnet-2.0"
+    nodeName = ${?NODENAME}
```
andyceo: Not sure if previous `nodeName = "ergo-testnet-2.0"` overwrited to empty string if NODENAME not set

Dockerfile
andyceo: /home/ergo

src/main/resources/application.conf
```diff
@@ -1,6 +1,7 @@
 ergo {
   # Directory to keep data
   directory = ${user.dir}"/ergo/data"
```
andyceo:
```suggestion
  directory = ${user.dir}"/.ergo"
```

PR #692, opened 2019-04-17
Extract validation rules for further soft-forkability implementation

### LINKED ISSUE #581: Soft-forkability
The goal of this issue is to allow to update protocol via soft forks after mainnet launch.

catena2w:
Relates to #581 
Relies on Scorex PR https://github.com/ScorexFoundation/Scorex/pull/337

# Questions to discuss:
- **hdrTooOld**: we do not apply block sections to headers, older than <config.keepVersions>. This is a consensus rule, that is different for all nodes.
- **hdrFutureTimestamp**: we do not allow to apply blocks far than 20 minutes in future. This error is marked as fatal so the block will never be applied. An attacker may try to split the network using this rule, as far as local timestamps are different
- **hdrVotes**: it is now presented as a single big rule, while actually consists of multiple smaller ones.
- **hdrPoW**: it is now presented as a single big rule, while actually consists of multiple smaller ones.
- **Parameters.maxBlockSize***: is not checked anywhere
- **recoverable errors*: there are some recoverable errors, that are implementation-specific, I did not include them into a yellow paper


src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -67,7 +67,9 @@ trait ErgoState[IState <: MinimalState[ErgoPersistentModifier, IState]]
           .validateNoFailure(txBoxesToSpend, boxesToSpendTry)
           .validateNoFailure(txDataBoxes, dataBoxesTry)
           .payload[Long](r.value)
-          .validate(tx.validateStateful(boxesToSpendTry.get.toIndexedSeq, dataBoxesTry.get.toIndexedSeq, currentStateContext, r.value).result)
+          .validateTry(dataBoxesTry, e => ModifierValidator.fatal("Missed data boxes", e)) { case (_, dataBoxes) =>
+            tx.validateStateful(boxesToSpendTry.get.toIndexedSeq, dataBoxes.toIndexedSeq, currentStateContext, r.value).result
```
oskin1: Plase remove another one `.get` also

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -40,35 +43,42 @@ trait ErgoState[IState <: MinimalState[ErgoPersistentModifier, IState]]
...
-            .sequence
-          val dataBoxesTry: Try[List[ErgoBox]] = tx.dataInputs.toList
-            .map(in => checkBoxExistence(in.boxId))
-            .sequence
-          boxesToSpendTry.flatMap { boxes =>
-            dataBoxesTry.flatMap { dataBoxes =>
-              tx.statefulValidity(boxes.toIndexedSeq, dataBoxes.toIndexedSeq, currentStateContext, accumulatedCost)
-            }
+    implicit val vs: ValidationSettings = ValidationRules.initialSettings
+
```
kushti: The new execTx is not aborting on the first invalid transaction, which could be harmful (open to DoS attacks).

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -221,8 +220,8 @@ class ErgoWalletActor(ergoSettings: ErgoSettings, boxSelector: BoxSelector)
         val lockWithAddress = (addressOpt orElse publicKeys.headOption)
           .getOrElse(throw new Exception("No address available for box locking"))
         val minimalErgoAmount =
-          BoxUtils.minimalErgoAmountSimulated(lockWithAddress.script, Colls.fromItems(assetId -> amount), nonMandatoryRegisters, parameters)
-        new ErgoBoxCandidate(minimalErgoAmount, lockWithAddress.script, height, Colls.fromItems(assetId -> amount), nonMandatoryRegisters)
+          BoxUtils.minimalErgoAmountSimulated(lockWithAddress.script, Seq(assetId -> amount).toColl, nonMandatoryRegisters, parameters)
```
kushti: Why to do the excessive conversion?

src/main/scala/org/ergoplatform/nodeView/state/DigestState.scala
```diff
@@ -52,11 +53,9 @@ class DigestState protected(override val version: VersionTag,
...
       .fold[Try[ErgoBox]](Failure(new Exception(s"Box with id ${Algos.encode(id)} not found")))(Success(_))
 
-    execTransactionsTry(transactions, currentStateContext)(checkBoxExistence) match {
-      case Success(executionCost) if executionCost <= currentStateContext.currentParameters.maxBlockCost => Success(())
-      case Success(executionCost) => Failure(new Exception(s"Transaction cost $executionCost exceeds limit"))
-      case failure => failure.map(_ => ())
-    }
+    execTransactions(transactions, currentStateContext)(checkBoxExistence)
+      .toTry
+      .map(tc => ().ensuring(tc <= currentStateContext.currentParameters.maxBlockCost))
```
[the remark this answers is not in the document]
catena2w: We use `ensuring` quite often to double-check ourselves, but I can remove it if you wish

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
catena2w: Looks like a merge bug, fixed

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
catena2w: It does not, the invalid transaction will lead to `r: Invalid`

kushti:
Approved, but please remove .ensuring (or make checks in both the UTXO- and DigestState).

PR #705, opened 2019-04-25
Wallet persistence

### LINKED ISSUE #695: Wallet persistance

oskin1:
Close #695


src/main/resources/application.conf
```diff
@@ -182,6 +182,12 @@ ergo {
     # which are potentially wallet's. But to be sure, script execution is needed, which could be costly to do in a bulk.
     # So we check from a queue only one box per "scanningInterval".
     scanningInterval = 1s
+
+    # Perform blocks scanning when the wallet is locked (may consume additinal disk space)
```
catena2w: how can we perform block scanning when the wallet is locked?

src/main/resources/application.conf
```diff
@@ -182,6 +182,12 @@ ergo {
...
+    postponedScanning = true
+
+    # Save used boxes (may consume additinal disk space) or delete them immediately
+    keepHistory = false
```
catena2w: Don't like the name. Let's rename to something like `keepSpentBoxes`. Also I propose to set it to true by default

src/main/scala/org/ergoplatform/nodeView/wallet/IdUtils.scala
```diff
@@ -0,0 +1,18 @@
+package org.ergoplatform.nodeView.wallet
+
+import org.ergoplatform.settings.Algos
+
+object IdUtils {
```
catena2w: We already have these functions in `Scorex`.  May be the right place for this class is `scorex-utils` package.
Also we use supertagged to avoid type errors, in this case `EncodedBoxId` should not be used as `EncodedTokenId`

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/OffChainRegistry.scala
```diff
@@ -0,0 +1,67 @@
+package org.ergoplatform.nodeView.wallet.persistence
+
+import org.ergoplatform.nodeView.history.ErgoHistory
+import org.ergoplatform.wallet.boxes.TrackedBox
+
+/**
+  * Holds version-agnostic indexes (such as off-chain boxes) in runtime memory.
```
catena2w: I see it also holds `onChainBalances` and `height` that are not version-agnostic (and it is not clear what do you mean by `version`)

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/OffChainRegistry.scala
```diff
@@ -0,0 +1,67 @@
...
+    val balance = offChainBalances.map(_.value).sum + onChainBalances.map(_.value).sum
+    val tokensBalance = (offChainBalances ++ onChainBalances)
+      .flatMap(_.assets)
+      .foldLeft(Map.empty[EncodedTokenId, Long]) { case (acc, (id, amt)) =>
+        acc.updated(id, acc.getOrElse(id, 0L) + amt)
+      }
+    RegistryIndex(height, balance, tokensBalance, Seq.empty)
+  }
+
+  def updated(certainBoxes: Seq[TrackedBox],
```
catena2w: Please add method annotation, including parameters.
The same for `updateOnBlock`, it is not clear what is the difference between them and what is the difference between `certainBoxes` and `allCertainBoxes` in these methods

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/OffChainRegistry.scala
```diff
@@ -0,0 +1,67 @@
+package org.ergoplatform.nodeView.wallet.persistence
+
+import org.ergoplatform.nodeView.history.ErgoHistory
+import org.ergoplatform.wallet.boxes.TrackedBox
+
+/**
+  * Holds version-agnostic indexes (such as off-chain boxes) in runtime memory.
+  */
+final case class OffChainRegistry(height: Int,
```
catena2w: please add parameters annotations.
E.g. it is not clear - if I have an onchain box `A` and put a transaction that spend box `A` and creates a new one box `B` - `onChainBalances` contains box `A` and  `offChainBalances` contains box `B`? Or box `A` is removed from `onChainBalances`?

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/PostponedBlock.scala
```diff
@@ -0,0 +1,57 @@
...
+import org.ergoplatform.nodeView.wallet.IdUtils._
+import org.ergoplatform.settings.Constants
+import scorex.core.serialization.ScorexSerializer
+import scorex.util.{ModifierId, bytesToId, idToBytes}
+import scorex.util.serialization.{Reader, Writer}
+
+/**
+  * Wallet-critic data extracted from full block to be processed later.
+  */
+final case class PostponedBlock(id: ModifierId,
```
catena2w: Add parameters description

src/main/resources/application.conf
oskin1: Once the wallet is unlocked addresses to track are loaded to persistent storage, so in the future the wallet is able to save outputs containing these addresses to try to resolve them when the prover is available.

src/main/resources/application.conf
oskin1: Spent boxes are not used for now, there is also no interface to view them.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/OffChainRegistry.scala
oskin1: The latter.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/OffChainRegistry.scala
oskin1: In this case `onChainBalances` appear to be version-agnostic as it can be updated in result of an off-chain spending.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/PostponedBlock.scala
oskin1: Done

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -93,113 +90,165 @@ class ErgoWalletActor(ergoSettings: ErgoSettings, boxSelector: BoxSelector)
...
+          case ((outAcc, inAcc), tx) =>
+            (outAcc ++ extractOutputs(tx).map(tx.id -> _), inAcc ++ extractInputs(tx).map(tx.id -> _))
+        }
+      if (outputs.nonEmpty || inputs.nonEmpty) {
+        if (proverOpt.isDefined) {
+          processBlock(block.id, block.height, inputs, outputs)
+        } else if (walletSettings.postponedScanning) {
+          // save wallet-critical data from block to process it later.
+          val postponedBlock = PostponedBlock(block.id, block.height, inputs, outputs)
+          storage.putBlock(postponedBlock)
```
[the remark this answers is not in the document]
oskin1: Reading from the mutable source from a separate thread doesn't seem to be a good idea.

PR #719, opened 2019-05-11
Sigma with soft-forkability

aslesarenko:


catena2w:
replaced by https://github.com/ergoplatform/ergo/pull/754

PR #742, opened 2019-05-26
Sigma constants check

gostkin:


src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -107,6 +107,10 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
     validateStateful(boxesToSpend, dataBoxes, stateContext, accumulatedCost).result.toTry
   }
 
+  def getByteLength[T](b: T, serializer: ScorexSerializer[T]): Long = {
+    return serializer.toBytes(b).length
```
oskin1: `return` is redundant syntax

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -120,6 +124,27 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
       .validate(txManyInputs, inputs.size <= Short.MaxValue, s"$id: ${inputs.size}")
       .validate(txManyDataInputs, dataInputs.size <= Short.MaxValue, s"$id: ${dataInputs.size}")
       .validate(txManyOutputs, outputCandidates.size <= Short.MaxValue, s"$id: ${outputCandidates.size}")
+      .validateSeq(outputCandidates)((validationState, out) => validationState
+        .validate(txMaxBoxSize, getByteLength[ErgoBox](out.toBox(id, 0), ErgoBoxSerializer) <= ErgoConstants.MaxBoxSize.get)
+        .validate(txAdditionalRegistersCount, out.additionalRegisters.size <= ErgoBox.nonMandatoryRegistersCount)
+        .validateSeq(out.additionalRegisters)((validationState, reg) => validationState
```
oskin1: Please break the line after arrow ascription (`=>`)

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -120,6 +124,27 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
+          .validate(txRegisterData, reg._2 match {
+            case (coll: ConcreteCollection[_]) =>
+              val w = SigmaSerializer.startWriter()
+              ValueSerializer.serialize(coll, w)
+              w.length <= ErgoConstants.MaxByteArrayLength.get
```
oskin1: Please avoid using `.get` on `Option`s and `Try`s. Using it outside `Try` is unsafe.

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -120,6 +124,27 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
+            case (constant: Constant[_]) =>
+              constant.tpe match {
+                case l: SBigInt.type =>
+                  val w = SigmaSerializer.startWriter()
+                  DataSerializer.serialize(constant.value, constant.tpe, w)
+                  w.length <= ErgoConstants.MaxBigIntSizeInBytes.get
```
oskin1: Same as above

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
gostkin: It's not an Option or Try, it is SizeConstant class that always can run get

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
gostkin: It's not an `Option` or `Try`, it is `SizeConstant` class that always can run `.get`

catena2w:
This checks will remain to be done inside sigmastate-interpreter

PR #778, opened 2019-06-08
Different address prefixes for testnet and devnet

oskin1:


src/main/resources/devnet.conf
```diff
@@ -1,8 +1,7 @@
 ergo {
   chain {
-    # Network address prefix, currently reserved values are 0x00 (money chain mainnet) and 0x20 (32 in decimal,
-    # money chain testnet)
-    addressPrefix = 0
+    # Network address prefix, currently reserved values are 0 (money chain mainnet) and 16 (money chain testnet)
+    addressPrefix = 32
```
kushti: why value is 32 there but 16 in comments and other places?

oskin1: It is devnet config, but only mainnet (0) and testnet (16) are mentioned in comments.

PR #813, opened 2019-06-26
Wallet API improvements

oskin1:


src/main/resources/api/openapi.yaml
```diff
@@ -1690,6 +1690,73 @@ paths:
...
+        schema:
+          type: integer
+          format: int32
+          minimum: 0
+          default: 0
+    get:
+      security:
+        - ApiKeyAuth: [api_key]
+      summary: Get a list of all wallet-related boxes
+      operationId: walletUnspentBoxes
```
catena2w: it is not about unspent boxes only

src/main/resources/api/openapi.yaml
```diff
@@ -1690,6 +1690,73 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiError'
 
+  /wallet/transactions:
+    get:
```
catena2w: Let's also add a filter from/until height?

src/main/resources/api/openapi.yaml
```diff
@@ -1690,6 +1690,73 @@ paths:
...
+        - wallet
+      responses:
+        '200':
+          description: A list of all wallet-related transactions
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ErgoTransaction'
```
catena2w: Can we also return inclusion height here?

PR #829, opened 2019-07-02
Only include transactions with low complexity

catena2w:


src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -317,13 +318,14 @@ class ErgoMiner(ergoSettings: ErgoSettings,
     //only transactions valid from against the current utxo state we take from the mem pool
     val emissionTxOpt = ErgoMiner.collectEmission(state, minerPk, ergoSettings.chainSettings.emissionRules).map { tx =>
       implicit val verifier: ErgoInterpreter = ErgoInterpreter(state.stateContext.currentParameters)
-      val cost = state.validateWithCost(tx, Some(upcomingContext)).get
+      val cost = state.validateWithCost(tx, Some(upcomingContext), Int.MaxValue).get
```
aslesarenko: Why not use maxTransactionComplexity?

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -474,7 +477,8 @@ object ErgoMiner extends ScorexLogging {
                     current -> invalidTxs
                   }
               }
-            case _ =>
+            case Failure(e) =>
+              log.debug(s"Do not incude transaction ${tx.id} due to ${e.getMessage}")
```
aslesarenko:
```suggestion
              log.debug(s"Do not include transaction ${tx.id} due to ${e.getMessage}")
```

src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
```diff
@@ -44,7 +44,7 @@ class UtxoStateSpecification extends ErgoPropertyTest with ErgoTransactionGenera
       val newBoxes = IndexedSeq(newFoundersBox, rewardBox)
       val unsignedTx = new UnsignedErgoTransaction(inputs, IndexedSeq(), newBoxes)
       val tx: ErgoTransaction = ErgoTransaction(defaultProver.sign(unsignedTx, IndexedSeq(foundersBox), emptyDataBoxes, us.stateContext).get)
-      us.validateWithCost(tx, None).get should be <= 100000L
+      us.validateWithCost(tx, None, Int.MaxValue).get should be <= 100000L
```
aslesarenko: Let's introduce a constant for cost limit long with Constants.DefaultComplexityLimit.

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
catena2w: We'd like to collect emission anyway, even if its complexity is too high.

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -53,6 +55,75 @@ class ErgoMinerSpec extends FlatSpec with ErgoTestHelpers with ValidBlocksGenera
     empty.copy(nodeSettings = nodeSettings, chainSettings = chainSettings)
   }
 
+  it should "not include too complex transactions" in new TestKit(ActorSystem()) {
```
oskin1: Seems this test produces unstable result

catena2w: Fixed in https://github.com/ergoplatform/ergo/pull/829/commits/e490bebd419e3de0287a4a86d8901b43ebdbdbf0

PR #830, opened 2019-07-02
addressToRaw and rawToAddress API methods

kushti:


kushti:
@oskin1 made roundtrip test for raw/address conversion

PR #834, opened 2019-07-10
Optimised ErgoBox deserialization

aslesarenko:


aslesarenko:
Already merged as part of v3.0.6

PR #851, opened 2019-07-21
DB migration

oskin1:
In this PR:
- Key-value storage and Versioned key-value storage are implemented on top of LevelDB
- Wallet storage updated
- History storage updated


kushti:
@oskin1 please resolve conflicts w. master

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
```diff
@@ -19,76 +19,68 @@ import scala.util.{Failure, Success, Try}
...
-          HistoryModifierSerializer.parseBytesTry(bBytes) match {
-            case Success(pm) =>
-              log.trace(s"Cache miss for existing modifier $id")
-              modifiersCache.put(id, pm)
-              Some(pm)
-            case Failure(_) =>
-              log.warn(s"Failed to parse modifier ${encoder.encode(id)} from db (bytes are: ${Algos.encode(bBytes)})")
-              None
-          }
+  def modifierById(id: ModifierId): Option[ErgoPersistentModifier] =
```
kushti: log.trace(s"Got modifier $id from cache") has been lost during "refactoring"

src/main/scala/org/ergoplatform/db/LDBFactory.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import scala.util.Try
+
+object LDBFactory extends ScorexLogging {
+
+  private val nativeFactory = "org.fusesource.leveldbjni.JniDBFactory"
+  private val javaFactory   = "org.iq80.leveldb.impl.Iq80DBFactory"
+
+  lazy val factory: DBFactory = {
+    val pairs = for {
```
kushti: please use .map / .flatmap / .withFilter instead of a for compehension in the main code

src/main/scala/org/ergoplatform/db/KVStore.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  type K = Array[Byte]
+  type V = Array[Byte]
+
+  protected val db: DB
+
+  def get(key: K): Option[V] =
+    Option(db.get(key))
+
+  def getAll(cond: (K, V) => Boolean): Seq[(K, V)] = {
```
kushti: why not to convert iterator into seq in getAll()? The more flexible option is to return Iterator and then  it is up to client whether to read it fully or partially.

src/main/scala/org/ergoplatform/db/VersionedLDBKVStore.scala
```diff
@@ -0,0 +1,136 @@
...
+  def update(toInsert: Seq[(K, V)], toRemove: Seq[K])(version: VersionId): Unit = {
+    require(version.length == Constants.HashLength, "Illegal version id size")
+    val ro = new ReadOptions()
+    ro.snapshot(db.getSnapshot)
+
+    require(Option(db.get(version, ro)).isEmpty, "Version id is already used")
+
+    val insertedKeys = mutable.ArrayBuffer.empty[K]
+    val altered = mutable.ArrayBuffer.empty[(K, V)]
+    for ((k, _) <- toInsert) Option(db.get(k, ro))
```
kushti: please avoid for compehensions

src/main/scala/org/ergoplatform/db/KVStore.scala
oskin1: Iterators are not used in the project, so all we got in case of using iterators is convertion overhead.

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
oskin1: It was not lost but removed as modifiers cache is already tested so there is no need in such logging.

PR #852, opened 2019-07-22
Sigma simplification and optimisation

aslesarenko:


src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
```diff
@@ -63,23 +64,26 @@ class ErgoStateContext(val lastHeaders: Seq[Header],
...
 
-  def sigmaPreHeader: special.sigma.PreHeader = PreHeader.toSigma(lastHeaders.head)
+  override def sigmaPreHeader: special.sigma.PreHeader =
+    PreHeader.toSigma(lastHeaders.headOption.getOrElse(PreHeader.fake))
 
-  def sigmaLastHeaders: Coll[special.sigma.Header] = new CollOverArray(lastHeaders.tail.map(h => Header.toSigma(h)).toArray)
-
-  // todo remove from ErgoLikeContext and from ErgoStateContext
-  def lastBlockMinerPk: Array[Byte] = sigmaPreHeader.minerPk.getEncoded.toArray
+  override def sigmaLastHeaders: Coll[special.sigma.Header] = new CollOverArray(lastHeaders.drop(1).map(h => Header.toSigma(h)).toArray)
```
aslesarenko: why `tail` is replaced with `drop(1)`?

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
```diff
@@ -254,12 +258,11 @@ object ErgoStateContext {
...
 
   /**
-    * Initialize empty state context with fake PreHeader
+    * Initialize empty state context
     */
   def empty(genesisStateDigest: ADDigest, settings: ErgoSettings): ErgoStateContext = {
-    new ErgoStateContext(Seq.empty, None, genesisStateDigest, LaunchParameters, ErgoValidationSettings.initial, VotingData.empty)(settings.chainSettings.voting)
-      .upcoming(org.ergoplatform.mining.group.generator, 0L, settings.chainSettings.initialNBits, Array.fill(3)(0.toByte),
-        ErgoValidationSettingsUpdate.empty, 0.toByte)
+    new ErgoStateContext(Seq.empty, None, genesisStateDigest, LaunchParameters, ErgoValidationSettings.initial,
```
aslesarenko: Why is this, some comment for this method is required.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -597,7 +597,7 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
     if (secrets.size == 1) {
       Success(DerivationPath(List(0, 1), publicBranch = false))
     } else {
-      nextPath(List.empty, secrets.map(_.path.decodedPath.tail))
+      nextPath(List.empty, secrets.map(_.path.decodedPath.tail.toList))
```
aslesarenko: Why this is changed?

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
greenhat: To handle empty `lastHeaders` (without throwing).

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
greenhat: @aslesarenko DerivationPath.decodedPath in the wallet was changed from List to Seq. Actually, I've updated this code in https://github.com/ergoplatform/ergo/pull/852/commits/d9b99f333ae6980244010cb842b7581adb0fe44c
with a better fit for this change.

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
aslesarenko: This can be described as an underlying reason for the new code.
"ErgoStateContext.empty to return ErgoStateContext rather than UpcomingStateContext since it later ends up in ErgoStateContextSerializer.serialize and parse does not support it thus PreHeader was lost" can be put in the code, not in PR comments. Then you don't need to comment at all. 
Actually adding comment in the code is better, than just responding in PR comments. 
The efforts are not that larger.

greenhat: @aslesarenko 
It’s not about the efforts. My main beef with this kind of comment is that Cons outweighs the Pros.

Pros:
- you see explanation why the code was changed right beside the code change(in a PR);

Cons:
- the comment only makes sense in the context of the change (PR);
- the comment will probably confuse a future developer(without the context of the change) ; 
- the comment will inevitably be outdated since no one will update it when the code(far from this comment) is changed; 

Thus, for a one-time convenience, we'll pay too much.
I believe that the best place for a comment like that is a commit message. The second best place is PR comment.

PR #858, opened 2019-07-31
ErgoLikeContext with json codecs and removed currentHeight and minerPk

greenhat:
Sigma PR https://github.com/ScorexFoundation/sigmastate-interpreter/pull/575
Wallet PR https://github.com/ergoplatform/ergo-wallet/pull/37

Todo: 
- [x] switch to sigma's json codecs;

**Overview of the changes.**

`ErgoStateContext`:
- removed `lastBlockMinerPk` (use `PreHeader.minerPk`);
- made `sigmaPreHeader` to return `PreHeader.fake` if `lastHeaders` are empty;
- made `currentHeight` to just return `sigmaPreHeader.height` (without the fallback on empty `lastHeaders`);
- made `ErgoStateContext.empty` to return `ErgoStateContext` rather than `UpcomingStateContext` since it later ends up in `ErgoStateContextSerializer.serialize` and `parse` does not support it thus PreHeader was lost;
- added `assert` in `ErgoStateContextSerializer.serialize` to fail if `UpcomingStateContext` gets into serializer; 

Json codecs:
- added `ErgoBox.transactionId` and `ErgoBox.index` to Json repr (not required in tx `outputs` decoding);
- moved most of the Json codecs to sigma into `JsonCodecs` preserving Json repr of the existing types;

Various `Algos` methods moved to sigma `ErgoCodecs`.


src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
```diff
@@ -286,6 +284,12 @@ object ErgoStateContext {
 case class ErgoStateContextSerializer(votingSettings: VotingSettings) extends ScorexSerializer[ErgoStateContext] {
 
   override def serialize(obj: ErgoStateContext, w: Writer): Unit = {
+    /* NOHF PROOF:
+    Changed: added assert to not let `UpcomingStateContext` get serialized.
+    Motivation: only `ErgoStateContext` is supported in `parse`.
+    Safety: used only in `ErgoMiner.createCandidate` and does get serialized.
```
aslesarenko: not clear what *used only in ...*?

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
```diff
@@ -23,7 +24,7 @@ object AugWalletTransaction {
       "address" -> e.fromProposition(box.ergoTree).toOption.map(_.toString).asJson,
       "assets" -> box.additionalTokens.toArray.toSeq.asJson,
       "creationHeight" -> box.creationHeight.asJson,
-      "additionalRegisters" -> registersEncoder(box.additionalRegisters)
+      "additionalRegisters" -> box.additionalRegisters.asInstanceOf[Map[NonMandatoryRegisterId, EvaluatedValue[SType]]].asJson
```
aslesarenko: Is this necessary change? Looks like the previous version was better :)

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
```diff
@@ -33,7 +34,7 @@ object AugWalletTransaction {
       "id" -> obj.wtx.tx.id.asJson,
       "inputs" -> obj.wtx.tx.inputs.asJson,
       "dataInputs" -> obj.wtx.tx.dataInputs.asJson,
-      "outputs" -> obj.wtx.tx.outputs.asJson,
+      "outputs" -> obj.wtx.tx.outputs.toSeq.asJson(Encoder.encodeSeq(enc)),
```
aslesarenko: same here, why this is necessary?

src/main/scala/org/ergoplatform/settings/Algos.scala
```diff
@@ -1,20 +1,16 @@
...
 
-import scala.util.Try
+object Algos extends ErgoAlgos with scorex.core.utils.ScorexEncoding {
 
-object Algos extends ScorexEncoding {
-
-  type HF = Blake2b256.type
-
-  val hash: HF = Blake2b256
+  // due to ErgoAlgos from sigma extends scorex.util.ScorexEncoding where encoder is BytesEncoder
```
aslesarenko: please describe in more detail, why this override is here?

src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
```diff
@@ -223,15 +223,15 @@ class UtxoStateSpecification extends ErgoPropertyTest with ErgoTransactionGenera
       // trying to apply transactions with missing data inputs
       val missedId: BoxId = ADKey @@ scorex.util.Random.randomBytes()
       us.boxById(missedId) shouldBe None
-      val missingDataInputs = (missedId +: existingBoxes).map(DataInput).toIndexedSeq
+      val missingDataInputs = (missedId +: existingBoxes).map(DataInput.apply).toIndexedSeq
```
aslesarenko: Is this necessary change?

src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
```diff
@@ -223,15 +223,15 @@ class UtxoStateSpecification extends ErgoPropertyTest with ErgoTransactionGenera
...
       val txWithMissedDataInputs = ErgoTransaction(headTx.inputs, missingDataInputs, headTx.outputCandidates)
       val incorrectTransactions = IndexedSeq(txWithMissedDataInputs)
       // proof for transaction works correctly, providing proof-of-non-existence for missed input
       val digest2 = us.proofsForTransactions(incorrectTransactions).get._2
       us.applyTransactions(incorrectTransactions, digest2, emptyStateContext) shouldBe 'failure
 
       // trying to apply transactions with correct data inputs
-      val existingDataInputs = existingBoxes.map(DataInput).toIndexedSeq
+      val existingDataInputs = existingBoxes.map(DataInput.apply).toIndexedSeq
```
aslesarenko: ?

src/main/scala/org/ergoplatform/modifiers/history/Header.scala
```diff
@@ -107,37 +108,24 @@ object PreGenesisHeader extends Header(
...
-    override def powOnetimePk: GroupElement = CGroupElement(header.powSolution.w)
-
-    override def powNonce: Coll[Byte] = new CollOverArray(header.powSolution.n)
-
-    override def powDistance: sigma.BigInt = CBigInt(header.powSolution.d.bigInteger)
-
-    override def votes: Coll[Byte] = new CollOverArray(header.votes)
-  }
+  def toSigma(header: Header): special.sigma.Header =
+    CHeader(
```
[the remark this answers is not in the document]
greenhat: @aslesarenko for this we need to move ergo's `Header` (https://github.com/ergoplatform/ergo/blob/6e16b9bac06c218c189ff1fd2b5b548f7a516ec7/src/main/scala/org/ergoplatform/modifiers/history/Header.scala#L31) to sigma first. Look like an overkill. I can switch back to anonymous types.

src/main/scala/org/ergoplatform/modifiers/history/PreHeader.scala
```diff
@@ -25,23 +25,26 @@ trait PreHeader {
...
-
-    override def height: Int = preHeader.height
-
-    override def minerPk: GroupElement = CGroupElement(preHeader.minerPk)
+object PreHeader {
 
-    override def votes: Coll[Byte] = new CollOverArray(preHeader.votes)
-  }
+  def toSigma(preHeader: PreHeader): special.sigma.PreHeader =
+    sigmastate.eval.CPreHeader(
```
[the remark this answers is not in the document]
greenhat: See my comment for `Header`. Same here.

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
```diff
@@ -63,23 +64,21 @@ class ErgoStateContext(val lastHeaders: Seq[Header],
 
   override type M = ErgoStateContext
 
-  def sigmaPreHeader: special.sigma.PreHeader = PreHeader.toSigma(lastHeaders.head)
+  override def sigmaPreHeader: special.sigma.PreHeader =
+    PreHeader.toSigma(lastHeaders.headOption.getOrElse(PreHeader.fake))
```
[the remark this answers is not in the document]
greenhat: @aslesarenko I agree that explicit NPE is better. But if we make `ErgoLikeContext.preHeader` to be `Option[PreHeader]` but left it non-optional in script `CONTEXT.preHeader` we would crash in runtime, right? Why is it better than do a hard unwrap and crash here?

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
```diff
@@ -63,23 +64,21 @@ class ErgoStateContext(val lastHeaders: Seq[Header],
...
-  def previousStateDigest: ADDigest = if (sigmaLastHeaders.toArray.nonEmpty) {
+  override def previousStateDigest: ADDigest = if (sigmaLastHeaders.toArray.nonEmpty) {
     ADDigest @@ sigmaLastHeaders.toArray.head.stateRoot.digest.toArray
   } else {
     genesisStateDigest
   }
 
   // todo remove from ErgoLikeContext and from ErgoStateContext
-  def currentHeight: Int = Try(sigmaPreHeader.height).getOrElse(ErgoHistory.EmptyHistoryHeight)
+  def currentHeight: Int = sigmaPreHeader.height
```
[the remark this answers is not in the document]
greenhat: @aslesarenko No. Well, in a way, yes. `getOrElse` is in `sigmaPreheader` with `ErgoHistory.EmptyHistoryHeight` in `PreHeader.fake`.

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
```diff
@@ -254,12 +253,11 @@ object ErgoStateContext {
...
 
   /**
-    * Initialize empty state context with fake PreHeader
+    * Initialize empty state context
     */
   def empty(genesisStateDigest: ADDigest, settings: ErgoSettings): ErgoStateContext = {
-    new ErgoStateContext(Seq.empty, None, genesisStateDigest, LaunchParameters, ErgoValidationSettings.initial, VotingData.empty)(settings.chainSettings.voting)
-      .upcoming(org.ergoplatform.mining.group.generator, 0L, settings.chainSettings.initialNBits, Array.fill(3)(0.toByte),
-        ErgoValidationSettingsUpdate.empty, 0.toByte)
+    new ErgoStateContext(Seq.empty, None, genesisStateDigest, LaunchParameters, ErgoValidationSettings.initial,
```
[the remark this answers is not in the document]
greenhat: @aslesarenko It's not. I described why I did this in PR's description:
>made ErgoStateContext.empty to return ErgoStateContext rather than UpcomingStateContext since it later ends up in ErgoStateContextSerializer.serialize and parse does not support it thus PreHeader was lost;

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
greenhat: Yes, somehow I missed the main part. :) Fixed to "Safety: `UpcomingStateContext` is used only in `ErgoMiner.createCandidate` and does not get serialized."

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
greenhat: @aslesarenko I removed cast. It uses implicitly `JsonCodecs.registersEncoder`. Do you want to call it implicitly everywhere?

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
greenhat: @aslesarenko `enc` is brought in scope above as custom `ErgoBox` encoder. I wanted to emphisize that we're using it and not the one from `JsonCodecs`.

src/main/scala/org/ergoplatform/settings/Algos.scala
greenhat: Changed to:
```
  // ErgoAlgos in sigmastate extends scorex.util.ScorexEncoding where encoder is BytesEncoder
  // but here we use scorex.core.utils.ScorexEncoding where encoder is ScorexEncoder
```
After `ScorexEncoder` is moved (there is even a TODO for that) from `scorex.core` to `scorex.util` we can fix this ugliness.

src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
greenhat: Not any more. Used to have `object DataInput`. Thanks for spotting this. Fixed.

src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
greenhat: Fixed.

src/main/scala/org/ergoplatform/modifiers/history/Header.scala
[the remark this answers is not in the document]
aslesarenko: No, I mean just decouple ergo code from directly referencing constructors. You can create factory methods on sigma side, which just incapsulate construction.

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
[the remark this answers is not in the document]
aslesarenko: It is not a problem if the script crashes at runtime, this means `false`.
And I don't see how the code "do a hard unwrap and crash here".

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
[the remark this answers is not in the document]
aslesarenko: Please add NO-HF PROOF comment

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
aslesarenko: I mean the previous code is shorter and more clear. Why you changed it?

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
aslesarenko: Let's leave this as a comment in the code

src/main/scala/org/ergoplatform/settings/Algos.scala
aslesarenko: please add `After ScorexEncoder is moved (there is even a TODO for that) from scorex.core to scorex.util we can fix this ugliness.` in the code as well

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
```diff
@@ -63,23 +64,21 @@ class ErgoStateContext(val lastHeaders: Seq[Header],
 
   override type M = ErgoStateContext
 
-  def sigmaPreHeader: special.sigma.PreHeader = PreHeader.toSigma(lastHeaders.head)
+  override def sigmaPreHeader: special.sigma.PreHeader =
+    PreHeader.toSigma(lastHeaders.headOption.getOrElse(PreHeader.fake))
 
-  def sigmaLastHeaders: Coll[special.sigma.Header] = new CollOverArray(lastHeaders.tail.map(h => Header.toSigma(h)).toArray)
```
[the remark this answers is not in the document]
catena2w: Yes, in ErgoStateContext first header is used as a preHeader, while the rest one are used as sigmaLastHeaders.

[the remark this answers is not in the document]
aslesarenko: Great, thanks

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
greenhat: This diff is outdated. Now it's `"additionalRegisters" -> box.additionalRegisters.asJson`.

src/main/scala/org/ergoplatform/nodeView/wallet/AugWalletTransaction.scala
greenhat: Added comment.

src/main/scala/org/ergoplatform/settings/Algos.scala
greenhat: Added.

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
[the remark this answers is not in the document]
greenhat: Done.

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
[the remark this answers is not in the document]
greenhat: You're right, we don't have to crash, we return `false`. But making `ErgoLikeContext.preHeader` optional would have bigger implications since we take `HEIGHT` and `minerPk` from it. And since it cannot be null by design I believe we should express this in types. Otherwise, it should be optional in ergo as well.

src/main/scala/org/ergoplatform/modifiers/history/Header.scala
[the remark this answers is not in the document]
greenhat: @aslesarenko I can only implement this factory method on sigma side passing `Header` fields as a method parameters. Which looks kinda ugly. Do you see any other way?

greenhat:
All changes are already in develop branch.

PR #862, opened 2019-08-13
TxSender

catena2w:
Actor that distributes all the coins in the node wallet to other addresses right after they become spendable.


src/main/scala/org/ergoplatform/local/TxSender.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+    * Will generate and broadcast transaction, that sends all coins in a wallet to addresses from withdrawDistribution.
+    *
+    * @return `Success` if transaction was sent,
+    *         `Failure` otherwise
+    */
+  private def withdraw(hr: ErgoHistoryReader, wr: ErgoWalletReader): Try[Unit] = Try {
+    val historyHeight = hr.headersHeight
+    lazy val registry = Await.result(wr.confirmedBalances, timeout.duration)
```
oskin1: Usage of `Async.await` is usually a bad practice. Why not making withdrawal in non-blocking fashion? -- wallet actor would anyway process requests synchronously.

PR #864, opened 2019-08-13
/wallet/p2s_address and /wallet/p2sh_address moved into /script/p2sAddress and /script/p2shAddress

kushti:


src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -0,0 +1,91 @@
...
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
+import akka.pattern.ask
+import org.ergoplatform.nodeView.wallet.requests.PaymentRequestDecoder
+import scorex.core.settings.RESTApiSettings
+
+case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
+                    (implicit val context: ActorRefFactory) extends ErgoBaseApiRoute with ApiCodecs {
```
oskin1: Please align 2nd parameter list with the first one.

PR #880, opened 2019-08-19
Fix for #876: onChainFilter passed to a box selector filters out user-provided boxes

### LINKED ISSUE #876: onChainFilter filters out user-provided boxes
In *generateTransactionWithOutputs* new onChainFilter introduced in #859 filters out boxes from inputsRaw.

 Possible solution : if user provides inputs (inputsRaw is not empty), pass *_=>true* as *filterFn* argument to boxSelector.select

kushti:
Close #876 

Strangely, there's the test "Generate transaction with user-defined input" which passed before the fix and passes now, while manual testing shows that before the fix /wallet/transaction/generate failed with "No enough boxes to assemble a transaction".


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -396,17 +405,19 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
                   AssetUtils.mergeAssets(boxTokens, acc)
                 }
 
-              val inputBoxes = if (inputs.nonEmpty) {
-                inputs
+              val (inputBoxes, filter) = if (inputs.nonEmpty) {
+                //inputs are provided externally, no need for filtering
+                (inputs
```
oskin1: Long methods chaining inside tuple constructor doesn't looks like a readable code -- try move `inputs.map { .. }.toIterator` to a separate variable.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -396,17 +405,19 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
...
                   .map { box => // declare fake inclusion height in order to confirm the box is onchain
                     TrackedBox(box.transactionId, box.index, Some(1), None, None, box, BoxCertainty.Certain)
                   }
-                  .toIterator
+                  .toIterator, noFilter: TrackedBox => Boolean)
```
oskin1: `noFilter: TrackedBox => Boolean` explicit type declaration looks redundant here.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -396,17 +405,19 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
...
               } else {
-                registry.readCertainUnspentBoxes.toIterator
+                //inputs are to be selected by the wallet
+                (registry.readCertainUnspentBoxes.toIterator, onChainFilter: TrackedBox => Boolean)
```
oskin1: Same for explicit type declaration here.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: done

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: Swapped type declarations with function literals

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: Swapped type declarations with function literals

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -358,6 +368,17 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
         }
     }
 
+  /**
+    * A helper method which makes TrackedBox sequence out of boxes provided
+    */
+  private def boxesToFakeTracked(inputs: Seq[ErgoBox]): Iterator[TrackedBox] = {
```
oskin1: Why making separate method which is used only once and locally inside another function?

kushti: To make generateTransactionWithOutputs shorter and simpler

oskin1: @kushti, I think it wouldn't be far less simpler if this piece of code was assigned to `val` inside `if` statement where it is used.

kushti: @oskin1 as alternative I propose to move generating functions to a dedicated package (object)

kushti: @oskin1 then helper functions even used once look pretty natural

PR #909, opened 2019-09-17
New sections #Open project in IDE; #FAQ

aslesarenko:


README.md
```diff
@@ -83,6 +84,22 @@ Ergo node works normally behind NAT, so you can keep closed your `9030` port, he
 
 It is also a good practice to keep closed REST API port `9053`, and connect to your node from inside another container in the same Docker network (this case not covered by this short quick start manual).
 
+## Open project in IDE
+
+Your can use IntelliJ IDEA (Community or Ultimate edition).
```
oskin1: Maybe [Metals ](https://scalameta.org/metals/) is also woth mentioning as an alternative to Idea?

aslesarenko:
Yes, please add the reference, but also mention that we don't use it and never tried it ourselves.

oskin1:
> Yes, please add the reference, but also mention that we don't use it and never tried it ourselves.

I've tried. Not as handy for big projects as Idea but more lightweight.

PR #910, opened 2019-09-18
Transactions rebroadcasting

### LINKED ISSUE #896: Rebroadcast mempool transactions
Currently mempool transaction being broadcasted only once, so if there are any problems with connectivity around that moment transactions remains in the local pool forever. 

As a solution, re-broadcasting for random transactions in the pool could be done by MempoolAuditor / CleanupWorker after cleaning up the pool.

oskin1:
Close #896


src/main/scala/org/ergoplatform/local/MempoolAuditor.scala
```diff
@@ -78,6 +85,18 @@ class MempoolAuditor(nodeViewHolderRef: ActorRef,
     context become working // ignore other triggers until work is done
   }
 
+  private def broadcastTransactions(): Unit = {
+    poolReaderOpt.foreach { pr =>
+      val txs = pr.digest(settings.nodeSettings.reBroadcastVolume)
```
aslesarenko: Let's give `digest` more descriptive name, such that makes the logic of this line readable.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -51,6 +51,16 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
     new ErgoMemPool(pool.filter(condition))
   }
 
+  override def digest(size: Int): Seq[ErgoTransaction] = {
+    val txs = pool.orderedTransactions.values.toArray
+    val (digest, _, _) = (0 to size).foldLeft(Seq.empty[ErgoTransaction], txs, txs.length) {
```
aslesarenko: I suggest to implement helper function which generated random subset of indexes in an 0 to size range.
Maybe such function already exists
Then you can use simple array indexing to extract txs.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolReader.scala
```diff
@@ -14,11 +14,16 @@ trait ErgoMemPoolReader extends MempoolReader[ErgoTransaction] {
...
+    */
   def getAllPrioritized: Seq[ErgoTransaction]
 
   def take(limit: Int): Iterable[ErgoTransaction]
 
   def modifierById(modifierId: ModifierId): Option[ErgoTransaction]
 
+  /** Returns a list of randomly selected transactions
+    */
+  def digest(size: Int): Seq[ErgoTransaction]
```
aslesarenko: @oskin1 write @param description and you will see that `size` is not natural name.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -88,13 +83,14 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
...
-    val propositionBytes = settings.chainSettings.monetary.feePropositionBytes
+  private def extractFee(tx: ErgoTransaction): Long =
     ErgoState.boxChanges(Seq(tx))._2
       .filter(_.ergoTree == settings.chainSettings.monetary.feeProposition)
       .map(_.value)
       .sum
-  }
+
+  private def randIndexes(qty: Int, max: Int): Seq[Int] =
+    (0 to qty).map(_ => Random.nextInt(max))
```
aslesarenko: The result may have duplicates. 
You probably may need to keep track of previously chosen indexes.
Or maybe google an algorithm for this task, it may have nice and efficient solution.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -78,12 +83,23 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
...
       .filter(_.ergoTree == settings.chainSettings.monetary.feeProposition)
       .map(_.value)
       .sum
+
+  private def randIndexes(qty: Int, max: Int): Seq[Int] = {
+    require(qty <= max)
+    @scala.annotation.tailrec
+    def loop(acc: Seq[Int], qtyLeft: Int): Seq[Int] =
+      if (qtyLeft > 0) {
+        val idx = Random.nextInt(max)
```
aslesarenko: This is problematic in case qty is close to max. You will have undeterministic number of cycles until you find all indexes.
Your original also was better actually.
The best I can think of is to:
1) put `0 .. max` range into debox.Set.
2) generate random index i in `0 .. set.size`
3) remove using `val wasRemoved = set.remove(i)`
4) check if wasRemoved otherwise try i+1 
This algorithm in O(qty)

aslesarenko: @oskin1 What is the expected size of max? For quantity it is 200, right?

oskin1: @aslesarenko, there is a requirement require(qty <= max), so loop always terminates. In case qty == max all the indexes in range 0 .. qty are picked.

oskin1: @aslesarenko, upper bound is defined by `ergo.node.mempoolCapacity` and the default is `1000`.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -78,12 +84,23 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
...
+    require(qty <= max)
+    val idx = Random.nextInt(max)
+    @scala.annotation.tailrec
+    def loop(acc: Seq[Int], idx: Int, leftQty: Int): Seq[Int] =
+      if (leftQty > 0) {
+        if (idx <= max) loop(acc :+ idx, idx + 1, leftQty - 1) else loop(acc, 0, leftQty)
+      } else {
+        acc
+      }
+    loop(Array.empty[Int], idx, qty)
```
aslesarenko: @oskin1 `val res = idx to (idx + qty)`?

oskin1: @aslesarenko, And what if we want to select `10` transactions having `100` transactions in the mempool and `Random.nextInt(100) = 95`? In your case the result would be `95, 96, ... 105`, so we get `5` transactions as a result instead of `10`

src/main/resources/reference.conf
```diff
@@ -56,6 +56,9 @@ ergo {
     # Time window within which a node performs mempool cleanup in between blocks application
     mempoolCleanupDuration = 20s
 
+    # Number of transactions from mempool to be re-broadcasted at each epoch
+    reBroadcastVolume = 200
```
catena2w: Isn't it too much for the default value? E.g. a node has 100 connections, it will receive 20k messages every 2 minutes which may lead to problems.

oskin1:
```suggestion
    reBroadcastVolume = 20
```

oskin1: This setting is now decreased down to 20 transactions.

kushti: finally, to 3

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: @oskin1 (idx to (idx + qty)).map( _ % max) ?

PR #917, opened 2019-09-25
Execute with context

### LINKED ISSUE #902: API: /script/execute or /script/executeWithContext
From morphic on Slack:

> Not exactly, the script has to be executed in the target context and the resulting SigmaProp (which is tree data structure in general) should be investigated. Then existing keys should be matched with those in the tree. If there are AND, OR or THRESHOLD connectives in the tree, corresponding sigma protocol operations have to be executed to generate signatures. (proofs in Sigma terminology)

I think along with #901, we also need the ability to execute a script in a given target context so that we can generate proofs for the resulting SigmaProp.

aslesarenko:
Resolves #902 

- [x] implement AnyValueEncoder/Decoder in sigma (create PR based on v3.0.4 branch)
- [x] use AnyValueEncoder
- [x] add unit tests
- [ ] add integration tests


src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -91,6 +114,74 @@ case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
...
+                            ctx: ErgoLikeContext)
+    extends TransactionRequest
+
+
+  implicit final val decodeMap: Decoder[Any] = new Decoder[Any] {
+    final def apply(c: HCursor): Result[Any] = {
+      val x = DataJsonEncoder.decode(c.value)
+      Right(x)
+    }
+  }
```
aslesarenko: @greenhat please help to replace this decoder with AnyValue decoder

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -91,6 +114,74 @@ case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
...
+
+  class ExecuteRequestDecoder(settings: ErgoSettings) extends Decoder[ExecuteRequest] with JsonCodecs {
+    def apply(cursor: HCursor): Decoder.Result[ExecuteRequest] = {
+      for {
+        script <- cursor.downField("script").as[String]
+        env <- cursor.downField("env").as[Map[String,Any]]
```
aslesarenko: @greenhat I think here should be Map[String, AnyValue], so AnyValueDecoder should work

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -91,6 +114,74 @@ case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
...
+
+  implicit val executeRequestDecoder: ExecuteRequestDecoder = new ExecuteRequestDecoder(ergoSettings)
+
+  implicit def sigmaBooleanEncoder: Encoder[SigmaBoolean] = {
+    sigma =>
+      val op = sigma.opCode.toByte.asJson
+      sigma match {
+        case dlog: ProveDlog   => Map("op" -> op, "h"->dlog.h.asJson).asJson
+        case dht: ProveDHTuple => Map("op" -> op, "g"->dht.g.asJson, "h"->dht.h.asJson, "u"->dht.u.asJson, "v"->dht.v.asJson).asJson
+        case _: TrivialProp => Map("op" -> op).asJson // besides opCode no additional bytes
```
aslesarenko: @knizhnik it should also save `condition` (see `abstract class TrivialProp(val condition: Boolean)`)

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -37,8 +59,9 @@ case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
     toStrictEntity(10.seconds) {
       corsHandler {
         p2shAddressR ~
-          p2sAddressR ~
-          addressToTree
+        p2sAddressR ~
+        addressToTree ~
+         executeWithContext
```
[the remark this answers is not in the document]
knizhnik: What about addressToTree?

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
greenhat: @aslesarenko `DataJsonEncoder` is in `master` and not in `core-opt`. I suggest to make `v3.0.4` out of master and rebase `core-opt` on top of it.

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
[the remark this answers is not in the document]
aslesarenko: It can also be renamed

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -91,7 +97,67 @@ case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
     }
   }
 
-  def addressToTree: Route = (get & path("addressToTree" / Segment)) { addressStr =>
+
+  case class ExecuteRequest(script: String,
```
aslesarenko:
```suggestion
  /**
   * Represent a request for execution of a script in a given context.
   * @param script  ErgoScript source code of the contract to execute
   * @param env      environment map of named constants used to compile the script
   * @param cox      script execution context 
   */
  case class ExecuteRequest(script: String,
```

src/test/scala/org/ergoplatform/api/routes/ScriptApiRouteSpec.scala
```diff
@@ -42,6 +42,22 @@ class ScriptApiRouteSpec  extends FlatSpec
       |}
       |""".stripMargin
 
+  it should "execute script with context" in {
+    val suffix = "/executeWithContext"
+    val stream = ClassLoader.getSystemClassLoader.getResourceAsStream("execute-script.json")
```
aslesarenko: There is one problem with this test. It is hard to judge what script is actually executed, and whether the checks actually verify anything.
I suggest to: 
1) compose ErgoLikeContext manually in the test
2) serialize it into Json (relying on ergoLikeContextEncoder from sigma)
3) pass the json to API

src/test/scala/org/ergoplatform/api/routes/ScriptApiRouteSpec.scala
```diff
@@ -42,6 +42,22 @@ class ScriptApiRouteSpec  extends FlatSpec
...
+      val value = json.hcursor.downField("value").downField("op").as[Int].right.get
+      val condition = json.hcursor.downField("value").downField("condition").as[Boolean].right.get
+      val cost = json.hcursor.downField("cost").as[Int].right.get
+      value shouldEqual -45
+      condition shouldEqual true
+      cost shouldEqual 40
+    }
+    val json = io.circe.parser.parse(req)
+    Post(prefix + suffix, json) ~> route ~> check(assertion(responseAs[Json]))
+  }
```
aslesarenko: In addition to the checks above, it make sense to call reduceToCrypto directly (i.e. without API) and compare the results.

src/test/scala/org/ergoplatform/api/routes/ScriptApiRouteSpec.scala
knizhnik: The sample data for ErgoLikeContext you have sent to me was too large to be hardcoded in the sources. Certainly it is possible to try to reduce it, skipping some non-required fields and reduced sample values. But it is not trivial (I tried to cut tail of large hex strings "0202020202...", but then this request is rejected). And also the less data is present in sample request, the less code coverage we have. This is why I put sample data in resource file. I do not think that it significantly complicates understanding which request is executed (we are not writing all code in the same file, aren't we?)

src/test/scala/org/ergoplatform/api/routes/ScriptApiRouteSpec.scala
knizhnik: What kind of functionality we will test in this case? Correct transfer of JSON response though HTTP protocol? Why don't we do in other similar requests (like p2sAddress)?

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
knizhnik: Comment added

src/test/scala/org/ergoplatform/api/routes/ScriptApiRouteSpec.scala
aslesarenko: If this file can be parsed in Json means you can do it manually and parse all components one by one.
Then you can create ErgoLikeContext from these components.
Maybe you don't need it here, but you may need in for the equivalence tests.

src/test/scala/org/ergoplatform/api/routes/ScriptApiRouteSpec.scala
aslesarenko: Yes, it is what API for. 
It opens access to internal components, and we need a collection of scripts which checks that
executing via API is equivalent to executing directly for every script in the collection.
Here we assume that executing directly is ground truth, since the Interpreter is tested in sigma.
To do this, you actually need to be able to create different contexts.

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
aslesarenko: I expected you just commit suggestion.

src/test/scala/org/ergoplatform/api/routes/ScriptApiRouteSpec.scala
knizhnik: Sorry, I do not understand. There are 3 components in executeWithContext request: script, environment and context. First two are small. Context is large. I can hardcode  script and environment in code and load context from resource file. But frankly speaking I do not understand why it is better. 

It may be useful to construct context from subcomponents  it there are many repeated subcomponents. But in is not the case of ErgoLikeContext, isn't it?

aslesarenko: This `json` is randomly generated context. So let's leave it like it is.

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
knizhnik: Sorry, I missed this option.

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -91,7 +97,73 @@ case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
     }
   }
 
-  def addressToTree: Route = (get & path("addressToTree" / Segment)) { addressStr =>
+
+  /**
+   * Represent a request for execution of a script in a given context.
```
oskin1: Please use Scala comment style instead of Java one:
```
/**
  * Comment
  */
```

src/main/scala/org/ergoplatform/api/ScriptApiRoute.scala
```diff
@@ -91,7 +97,73 @@ case class ScriptApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSettings)
...
+    }
+  }
+
+  implicit val executeRequestDecoder: ExecuteRequestDecoder = new ExecuteRequestDecoder(ergoSettings)
+
+  implicit def sigmaBooleanEncoder: Encoder[SigmaBoolean] = {
+    sigma =>
+      val op = sigma.opCode.toByte.asJson
+      sigma match {
+        case dlog: ProveDlog   => Map("op" -> op, "h"->dlog.h.asJson).asJson
```
oskin1: Please use spaces among tuple operator `->` operands.

src/main/resources/api/openapi.yaml
```diff
@@ -303,6 +307,282 @@ components:
...
+        dataInputs:
+          type: array
+          items:
+            $ref: '#/components/schemas/ErgoTransactionDataInput'
+        outputs:
+          type: array
+          items:
+            $ref: '#/components/schemas/ErgoTransactionOutput'
+
+    Header:
```
oskin1: There is `BlockHeader` which is already defined. If structure of the object described above is different please make It clear (encode it in the naming somehow).

knizhnik: "Header" and "BlockHeader" definitions  looks very similar.
ErgoLikeContext refers Header class and it is decoded in the following way in JsoncCodes:

`implicit val headerDecoder: Decoder[Header] = { cursor =>
    for {
      id <- cursor.downField("id").as[Coll[Byte]]
      version <- cursor.downField("version").as[Byte]
      parentId <- cursor.downField("parentId").as[Coll[Byte]]
      adProofsRoot <- cursor.downField("adProofsRoot").as[Coll[Byte]]
      stateRoot <- cursor.downField("stateRoot").as[AvlTreeData]
      transactionsRoot <- cursor.downField("transactionsRoot").as[Coll[Byte]]
      timestamp <- cursor.downField("timestamp").as[Long]
      nBits <- cursor.downField("nBits").as[Long]
      height <- cursor.downField("height").as[Int]
      extensionRoot <- cursor.downField("extensionRoot").as[Coll[Byte]]
      minerPk <- cursor.downField("minerPk").as[Coll[Byte]]
      powOnetimePk <- cursor.downField("powOnetimePk").as[Coll[Byte]]
      powNonce <- cursor.downField("powNonce").as[Coll[Byte]]
      powDistance <- cursor.downField("powDistance").as[special.sigma.BigInt]
      votes <- cursor.downField("votes").as[Coll[Byte]]
    } yield new CHeader(id, version, parentId, adProofsRoot, stateRoot, transactionsRoot, timestamp, nBits,
      height, extensionRoot, SigmaDsl.decodePoint(minerPk), SigmaDsl.decodePoint(powOnetimePk), powNonce, powDistance, votes)`

Some fields (powDistance, powNonce, powOnetimePk, minerPk, extensionRoot) are missed in "BlockHeader" definition. This is why I have to add "Header" section to openapi.yaml.
I have no idea how to "encode it in the naming somehow" if class Header really describes block header. I can change existed definition of BlockHeader, but I afraid that it will break some other code. For example it contains "extensionHash" and "extensionId" fields, while Header class contains "extensionRoot".

I still wonder of openapi.yaml is written manually or it is generated automatically from Scala object definitions.

src/main/resources/api/openapi.yaml
```diff
@@ -303,6 +315,285 @@ components:
...
+          format: int64
+          description: initial value of execution cost already accumulated before Interpreter.verify is called
+
+
+    ExecuteScript:
+      type: object
+      required:
+      - script
+      - env
+      - ctx
```
aslesarenko: this should be renamed `ctx`  -> `context`, because it is public API

aslesarenko: for environment let's keep it short

src/test/resources/execute-script.json
```diff
@@ -0,0 +1,1184 @@
...
...202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020
```
aslesarenko: this section is critical for practical use cases, so it should be covered by tests.
It is better to put some values here.

src/main/resources/api/openapi.yaml
knizhnik: Renamed "ctx" to "context",  but frankly speaking I do not understand logic of such renaming: either we should avoid shortcuts in all cases, either allow to use them. IMHO "ctx" is not more confusing then "env".

src/test/resources/execute-script.json
knizhnik: My understanding of Ergo data model is not enough to provide some alternative values here:(

src/main/resources/api/openapi.yaml
aslesarenko: Ok, let's rename `env` -> `namedConstants`, `ctx` -> `context`.
Environment is misleading, it is almost always confused with script parameters, which is not the case here.

src/test/resources/execute-script.json
aslesarenko: This corresponds to ContextExtension of ErgoLikeContext.
You can check usages of ContextExtension in existing tests to find typical usages.
The key is Byte integer and value is any value supported by DataJsonEncoder.

src/main/resources/api/openapi.yaml
knizhnik: Renamed

src/test/resources/execute-script.json
knizhnik: Sorry, where I can find this tests? I searched for ContextExtension in the ergo project but didn't find any references to it in tests. Also I looked at tests in ergo/src/test/scala/org/ergoplatform/http/routes but didn't find something relevant.

knizhnik: I investigated it a little bit more. Why do you say that this fragment corresponds to ContextExtension? It is part of ErgoTransactionOutput which is decoded by JsonCodes as ErgoBox. But I still failed to find any test for it.

PR #918, opened 2019-09-27
ErgoTree to address API method impl

### LINKED ISSUE #913: API method to get address from ErgoTree
Do an API method to get address from ErgoTree (given in Base16 form) if it is possible to extract the address.

oskin1:
Close #913

### DIFF
--- a/src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
+++ b/src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
@@ -10,20 +10,29 @@ import scorex.core.api.http.ApiError.BadRequest
 import scorex.core.api.http.{ApiResponse, UtilsApiRoute}
 import scorex.util.encode.Base16
 import sigmastate.basics.DLogProtocol.ProveDlog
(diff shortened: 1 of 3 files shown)


src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -33,17 +42,29 @@ class ErgoUtilsApiRoute(ergoSettings: ErgoSettings)(implicit context: ActorRefFa
...
+        e => BadRequest(e.getMessage),
+        raw => ApiResponse(Map("raw" -> raw).asJson)
+      )
+  }
+
+  def ergoTreeToAddressR: Route = (get & path("ergoTreeToAddress" / Segment)) { ergoTreeHex =>
+    Base16.decode(ergoTreeHex)
+      .flatMap { etBytes =>
+        ergoAddressEncoder.fromProposition(
+          treeSerializer.deserializeErgoTree(etBytes).toProposition(replaceConstants = false))
```
kushti: .toProposition(replaceConstants = false) is not needed, and I'm not sure why you are doing this step.

Issue #955, opened 2019-10-28
Example values in /utils/{hash/blake2b, seed, seed/{length}} are without required quotes


kushti:
Example values in Swagger for inputs and outputs of /utils/{hash/blake2b, seed, seed/{length}} are without required quotes.

PR #956, opened 2019-10-28
Fix for #955: example values in /utils/{hash/blake2b, seed, seed/{length}} are without required quotes

kushti:
This PR fixes improper example values in /utils/{hash/blake2b, seed, seed/{length}} (quotes are missed)

### DIFF
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -1493,7 +1493,7 @@ paths:
             application/json:
               schema:
                 type: string
-                example: '7e1e79dd4936bdc7d09f4ba9212849136b589fba4bcf4263a0961a95b65d08cb16'
+                example: '"7e1e79dd4936bdc7d09f4ba9212849136b589fba4bcf4263a0961a95b65d08cb16"'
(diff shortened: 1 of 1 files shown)


src/main/resources/api/openapi.yaml
```diff
@@ -1652,15 +1652,15 @@ paths:
...
+              example: '"7yaASMijGEGTbttYHg1MrXnWB8EbzjJnFLSWvmNoHrXV"'
       responses:
         '200':
           description: Base16-encoded 32 byte hash
           content:
             application/json:
               schema:
                 type: string
-                example: '7e1e79dd4936bdc7d09f4ba9212849136b589fba4bcf4263a0961a95b65d08cb16'
+                example: '"6ed54addddaf10fe8fcda330bd443a57914fbce38a9fa27248b07e361cc76a41"'
```
aslesarenko: @kushti Why the value changed while in other cases it is not?

kushti: @aslesarenko if the string is inside an object, Swagger generates quotes automatically

Issue #872, opened 2019-08-16
Inefficient difficulty check


kushti:
Currently, *requiredDifficultyAfter* in *HeadersProcessor* doing complete re-calculation for each block based on previous epochs. Though all the headers are probably cached to this moment, so no excessive reads from database, still, this is very inefficient way. Difficulty recalculation should use cached epoch difficulties.

Help wanted!

Bonus: 200 Ergs

knizhnik:
I do not completely understand how to simulate workload on which HeadersProcessor becomes a bottleneck. I tried to start new node according at my desktop. Synchronization of node takes substantial amount of time (several minutes), bust most of the time CPU usage is zero. Looks like it is waiting for receiving some data from network.

Only once I was able to catch the moment when application is actually doing something:
`nonEmpty:114, TraversableOnce (scala.collection)
nonEmpty$:114, TraversableOnce (scala.collection)
nonEmpty:108, AbstractTraversable (scala.collection)
loop$1:55, Growable (scala.collection.generic)
$plus$plus$eq:61, Growable (scala.collection.generic)
$plus$plus$eq$:53, Growable (scala.collection.generic)
$plus$plus$eq:184, ListBuffer (scala.collection.mutable)
$plus$plus$eq:47, ListBuffer (scala.collection.mutable)
$colon$plus:568, SeqLike (scala.collection)
$colon$plus$:566, SeqLike (scala.collection)
$colon$plus:45, AbstractSeq (scala.collection)
loop$1:198, HeadersProcessor (org.ergoplatform.nodeView.history.storage.modifierprocessors)
headerChainBack:210, HeadersProcessor (org.ergoplatform.nodeView.history.storage.modifierprocessors)
headerChainBack$:190, HeadersProcessor (org.ergoplatform.nodeView.history.storage.modifierprocessors)
headerChainBack:242, ErgoHistory$$anon$3 (org.ergoplatform.nodeView.history)
requiredDifficultyAfter:252, HeadersProcessor (org.ergoplatform.nodeView.history.storage.modifierprocessors)
requiredDifficultyAfter$:241, HeadersProcessor (org.ergoplatform.nodeView.history.storage.modifierprocessors)
requiredDifficultyAfter:242, ErgoHistory$$anon$3 (org.ergoplatform.nodeView.history)
$anonfun$validateChildBlockHeader$5:293, HeadersProcessor$HeaderValidator (org.ergoplatform.nodeView.history.storage.modifierprocessors)
[...]

PR #974, opened 2019-11-21
Replace headerChainBack with headerIdsAtHeight

knizhnik:
Issue: #872
In most cases requiredDifficultyAfter calculates difficulty of a block based on its parent. But on epoch boundary (each 1024 blocks) it also inspects blocks from useLastEpochs = 8 epochs.
Currently it is done using headesChainBack method which has to traverse backward 8*1024 blocks. But only 8 of them are actually needed. By using headerIdsAtHeight we can reduce number of accessed blocks 1024 times.


src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -249,8 +249,7 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
     if (heights.lengthCompare(1) == 0) {
       difficultyCalculator.calculate(Seq(parent))
     } else {
-      val chain = headerChainBack(heights.max - heights.min + 1, parent, _ => false)
-      val headers = chain.headers.filter(p => heights.contains(p.height))
+      val headers = heights.map(height => headerIdsAtHeight(height).lastOption).flatten.map(id=>typedModifierById[Header](id)).flatten
```
oskin1:
```suggestion
      val headers = heights.flatMap(height => headerIdsAtHeight(height).lastOption).flatMap(typedModifierById[Header])
```

kushti:
The issue we need to check is whether optimized difficulty recalculation may fail when we're validating a header from not a best chain

knizhnik:
> The issue we need to check is whether optimized difficulty recalculation may fail when we're validating a header from not a best chain

I have modified requiredDifficultyAfter so that it uses headerIdsAtHeight only in branch-less case, when there is exactly one header with specified height.

Issue #975, opened 2019-11-22
It is possible to run more than one ergo node instance


andyceo:
## Actual behaviour

It is possible to run more than one ergo node instance on the same jar file and data dir and ports.

## Steps to reproduce

1. Open first terminal and run `java -jar ergo.jar --mainnet` Node starts ok and start syncing, writing to it's data dir and open ports

2. Open second terminal and run `java -jar ergo.jar --mainnet` again Node starts, but that looks strange, because at least it can not occupy the same ports and must fail to run. But second node instance is running, also produce heavy CPU load 

## Expected behaviour

Trying to execute second instance of node on same ports, data dir must fail.

knizhnik:
is my understanding correct that we need to handle java.net.BindException as fatal error and shutdown application in this case?

I am not familiar with akko, but after some googling I find out that akka actors can be configured like this (https://doc.akka.io/docs/akka/current/fault-tolerance.html):

```
override val supervisorStrategy =
    OneForOneStrategy(maxNrOfRetries = 1) {
      case _: java.net.BindException => Stop
      case _: Exception              => Escalate
    }
```

I am not sure if it is enough to properly shutdown ergo node.
But looks like this code should be added to scorex.core.network.NetworkController
which is not part of ergo project. Should I change this library (if so, how to get it?) or may be I should look for some other way of configuring error handling policies for Akka?

andyceo:
For the note: it is easy to stop second node instance if you add empty `.lock` file (and/or `ergo.pid` containing first instance pid) to data directory. This how Bitcoin handle this (node is running):

```
# cd .bitcoin
# ls -lah
total 71M
drwxr-xr-x 5 root root 4.0K Nov 25 19:36 .
drwx------ 1 root root 4.0K Nov 16 15:17 ..
-rw------- 1 root root   75 Nov 16 15:17 .cookie
-rw------- 1 root root    0 Mar 18  2019 .lock
-rw------- 1 root root    0 Mar 18  2019 .walletlock
-rw------- 1 root root  739 Nov 25 18:21 banlist.dat
-rw------- 1 root root    2 Nov 16 15:17 bitcoind.pid
drwx------ 3 root root 120K Nov 24 19:39 blocks
drwx------ 2 root root  92K Nov 25 19:26 chainstate
drwx------ 2 root root 4.0K Nov 16 15:17 database
-rw------- 1 root root    0 Mar 18  2019 db.log
[...]

PR #977, opened 2019-11-26
Prevent running multiple node instances in the same working directory or bounded to the same port

knizhnik:
Issue: https://github.com/ergoplatform/ergo/issues/975

Terminate process when attempt to bind socket to the port failed or storage can not be opened (due to level-db lock, preventing concurrent access)


kushti:
@knizhnik please make clear title and description of what this PR is actually doing

PR #990, opened 2019-12-12
NiPoPoW basic algos and structure 

kushti:
Code here is imported from the "i211_proving" branch. It contains only PoPoW-related functions and tests, no refactoring or (sometimes breaking) changes in networking which present in i211_proving.

@oskin1 please do JSON serialization, API method to get nipopow proof for a given header (available if the node is in certain regime only), more tests (e.g.  chainOfLevel && isBetterThan functions are not covered by tests) and more documentation (for every method and PoPoWParams class)

PR #1009, opened 2019-12-19
API methods to get popow-header (header + interlinks vector)

kushti:
This PR is introducing methods to get header+interlink by giver header id or height. Examples: 

* http://127.0.0.1:9053/blocks/popowHeaderByHeight/100000
* http://127.0.0.1:9053/blocks/popowHeaderById/bfbba05dcf8c38a63e5a6ccc0783f23919d6aa9a987c39f9c8be9f5bb6eeb86b

Interlinks are stored in reverse order: first element is always genesis header, then level of lowest target met etc

openapi.yaml description is not done yet, thus the PR is in WIP status


src/main/scala/org/ergoplatform/modifiers/history/popow/PoPowAlgos.scala
```diff
@@ -0,0 +1,270 @@
...
+    * @param level      - Level of super-chain (µ)
+    */
+  def superChainQuality(chain: Seq[Header], superChain: Seq[Header], level: Int)
+                       (params: PoPowParams): Boolean = {
+    val downChain = chain
+      .filter(h => h.height >= superChain.head.height && h.height <= superChain.last.height) // C[C↑µ[0]:C↑µ[−1]] or C'↓
+    @scala.annotation.tailrec
+    def checkLocalGoodnessAt(mValue: Int): Boolean = {
+      val superChainSuffixSize = superChain.takeRight(mValue).size
+      val downChainSuffixSize = downChain.takeRight(mValue).size
```
ivanovs-4: Superchain quality states that for all m' ≥ m, it holds that 
`local-good δ (C↑ μ [−m' :], C↑ μ [−m' :]↓ , μ)`
As far as i can tell this operator means
`(C↑)↓ μ [−m' :]`
and this is different from the required
`C↑ μ [−m' :]↓`
Is it really correct?

oskin1: Right, seems we need to take M elements from super-chain first and then take the size of its projection on the down-chain

kushti: @oskin1 please make an issue and a fix based on #990

kushti: @ivanovs-4 @oskin1 please review my fix

ivanovs-4: It looks like that now it conforms to the specification.

PR #1010, opened 2019-12-23
Fixes for native-image generation

aslesarenko:


ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/EncryptedSecret.scala
```diff
@@ -24,4 +28,40 @@ object EncryptedSecret {
       Base16.encode(authTag), cipherParams)
   }
 
+  implicit object EncryptedSecretEncoder extends Encoder[EncryptedSecret] {
```
oskin1: Why adding this boilerplate -- encoder/decoder for this class is easily derived by circe?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/EncryptedSecret.scala
```diff
@@ -24,4 +28,40 @@ object EncryptedSecret {
...
+
+  }
+
+//  def main(args: Array[String]): Unit = {
+//    val res = decode[EncryptedSecret](
+//      """{ "cipherText": "", "salt": "", "iv": "", "authTag": "", "cipherParams": { "prf":"abc", "c": 10, "dkLen": 100 }}""")
+//    val jsonRaw = res.right.get.asJson.noSpaces
+//    println(res.toString)
+//    println(jsonRaw)
+//  }
```
oskin1: Do you want to leave it in the code?

kushti: commented out code should be avoided I guess

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/EncryptedSecret.scala
aslesarenko: Unfortunately, that macros generates code not supported by native-image of GraalVM. 
This is the point of this PR, to make the code native-image compatible.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/EncryptedSecret.scala
aslesarenko: removed

PR #1012, opened 2019-12-25
Take in account outputs of transactions from mempool

knizhnik:
This pull request is introducing a possibility to include a into a memory pool a transaction which is spending outputs of  unconfirmed transactions (i.e. other transactions from memory pool). This PR is also considering weight of transaction to be calculated for a family of dependent transactions, not a single transaction.


src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -32,48 +36,100 @@ final case class OrderedTxPool(orderedTransactions: TreeMap[WeightedTxId, ErgoTr
...
   }
 
   def invalidate(tx: ErgoTransaction): OrderedTxPool = {
     val inv = if (invalidated.size >= blacklistCapacity) invalidated - invalidated.firstKey else invalidated
     val ts = System.currentTimeMillis()
-    OrderedTxPool(orderedTransactions - weighted(tx),
-      transactionsRegistry - tx.id, inv.updated(tx.id, ts))
+    transactionsRegistry.get(tx.id).fold(OrderedTxPool(orderedTransactions, transactionsRegistry, inv.updated(tx.id, ts), outputs))(wtx =>
+      OrderedTxPool(orderedTransactions - wtx, transactionsRegistry - tx.id, inv.updated(tx.id, ts),
+        outputs -- tx.outputs.map(_.id)).updateFamily(tx, -wtx.weight))
```
oskin1: Lets avoid complex expressions inside function calls, it makes code less readable

oskin1: Or at least use named arguments

knizhnik: Sorry, I do not have much experience in Scala programming, so I don't know how to satisfy your request. Certainly I can rewrite this code in C-like way:

```
val entry = transactionsRegistry.get(tx.id)
if (entry.isEmpty) {
   OrderedTxPool(orderedTransactions, transactionsRegistry, inv.updated(tx.id, ts), outputs)
} else {
   val wtx =  entry.get
   val newPool = OrderedTxPool(orderedTransactions - wtx, 
       transactionsRegistry - tx.id, 
       inv.updated(tx.id, ts),
       outputs -- tx.outputs.map(_.id))
   newPool.updateFamily(tx, -wtx.weight)
}
```

I have used fold() thinking that it is more functional and "Scala" style.

Your suggestion to use named arguments seems to have no sense here, because names of parameters are the same as names of arguments. So writing `OrderedTxPool(orderedTransactions=orderedTransactions,...)`
just more complicates the expression and even making it more obscure.

PR #1019, opened 2020-01-06
Configurable parameter for headers-chain syncing status

kushti:
A node is doing headers-chain sync first. It is considered that the headers-chain is synced if currently observed tip is no older than some some parameter (previously hardcoded, 100 blocks ~= 200 minutes on average). After syncing headers-chain the node is deciding which full blocks to download. E.g. in the pruned regime, it downloads only certain suffix of the blockchain (determined by blocksToKeep parameter).

In this PR, this difference to decide syncing status (for the headers-chain) is made configurable.


src/main/resources/reference.conf
```diff
@@ -61,6 +61,11 @@ ergo {
 
     # Minimal fee amount of transactions mempool accepts
     minimalFeeAmount = 1000000
+
+    # A node is considering that the chain is synced if sees a block header with timestamp no more
+    # than headerChainTimeDiff blocks on average from future
```
aslesarenko: Please reformulate is simplier way, maybe with example.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -54,10 +56,13 @@ trait ToDownloadProcessor extends BasicReaders with ScorexLogging {
     }
 
     bestFullBlockOpt match {
+        //do not download full blocks if no headers-chain synced yet or SPV mode
```
aslesarenko: let's move comments inside the body of the corresponding branch.

src/test/scala/org/ergoplatform/settings/ErgoSettingsSpecification.scala
```diff
@@ -17,21 +17,21 @@ class ErgoSettingsSpecification extends ErgoPropertyTest {
     settings.nodeSettings shouldBe NodeConfigurationSettings(StateType.Utxo, verifyTransactions = true, 1000,
       poPoWBootstrap = false, 10, mining = false, Constants.DefaultComplexityLimit, 1.second, useExternalMiner = false, miningPubKeyHex = None,
       offlineGeneration = false, keepVersions = 200, mempoolCapacity = 100000, blacklistCapacity = 100000,
-      mempoolCleanupDuration = 10.seconds, rebroadcastCount = 3, minimalFeeAmount = 0)
+      mempoolCleanupDuration = 10.seconds, rebroadcastCount = 3, minimalFeeAmount = 0, headerChainTimeDiff = 100)
```
aslesarenko: we are not testing that parameter can be changed. Maybe clone the test cases and use some other value?

kushti: In these tests we're reading settings from a file and check that expected values being read from it. Values can not be changed. We can check another branch in a way like: 

(settings.nodeSettings == NodeConfigurationSettings(StateType.Utxo, verifyTransactions = true, 1000,
      poPoWBootstrap = false, 10, mining = false, Constants.DefaultComplexityLimit, 1.second, useExternalMiner = false, miningPubKeyHex = None,
      offlineGeneration = false, keepVersions = 200, mempoolCapacity = 100000, blacklistCapacity = 100000,
      mempoolCleanupDuration = 10.seconds, rebroadcastCount = 3, minimalFeeAmount = 0, headerChainTimeDiff = 10)) shouldBe false

but I'm not sure we're checking something new here

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
kushti: done

src/main/resources/reference.conf
kushti: Improved description.

kushti:
@knizhnik @aslesarenko thanks for the reviews!

renaming done: "headerChainTimeDiff" -> "headerChainDiff"

comments improved.

Please make another round!

PR #1020, opened 2020-01-08
IODB & old versioned leveldb based storage removed

kushti:
This PR is removing IODB dependency and also old LevelDB-based solution for versioned storage. Both wasn't used in 3.2.0 (outside of tests & benchmarks), and this PR removes the entities completely.


avldb/src/main/java/scorex/ByteUtils.java
```diff
@@ -0,0 +1,55 @@
...
+            int b2 = o2[i] & 0xFF;
+            if (b1 != b2)
+                return b1 - b2;
+        }
+        return o1.length - o2.length;
+    }
+
+
+    public static int byteArrayHashCode(byte[] data) {
+        //do not use Arrays.hashCode, it generates too many collisions (31 is too low)
```
aslesarenko: @greenhat @kushti @knizhnik We should remember to follow this observation (here is the [issue](https://github.com/ergoplatform/ergo/issues/1037))

avldb/src/main/java/scorex/ByteUtils.java
```diff
@@ -0,0 +1,55 @@
...
+        return h;
+    }
+
+    /**  returns byte array, which is greater than any other array of given size */
+    public static byte[] greatest(int size){
+        byte[] ret = new byte[size];
+        for(int i=0;i<size;i++){
+            ret[i] = (byte) 0xFF;
+        }
+        return ret;
```
aslesarenko: The loop can be replaced with java.util.Arrays.fill(ret, (byte) 0xFF);

benchmarks/src/test/scala/org/ergoplatform/db/LDBVersionedStoreRollbackBench.scala
```diff
@@ -30,16 +23,16 @@ object VLDBStoreRollbackBench
...
-    store.update(toInsert, toRemove)(version)
+    store.update(version, toRemove, toInsert)
   }
 
-  private val version = store.versions.last
+  private val version = store.rollbackVersions().head
 
   val et = Utils.time {
-    val result = store.rollbackTo(version)
+    val result = store.rollback(version)
```
aslesarenko: the previous name was better, because default meaning of `store.rollback(version)` is like removing the given version from db.

avldb/benchmarks/src/main/scala/scorex/crypto/authds/benchmarks/Helper.scala
```diff
@@ -64,7 +65,7 @@ object Helper {
     }
   }
 
-  def getProver(baseOperationsCount: Int = 0): BatchAVLProver[Digest32, HF] = {
+  def prover(baseOperationsCount: Int = 0): BatchAVLProver[Digest32, HF] = {
```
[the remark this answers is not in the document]
kushti: fixed

avldb/src/main/java/scorex/ByteUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class ByteUtils {
+    /**
+     * Compares primitive Byte Arrays.
+     * It uses unsigned binary comparation; so the byte with negative value is always higher than byte with non-negative value.
+     */
+    public static final Comparator<byte[]> BYTE_ARRAY_COMPARATOR = (o1, o2) -> compare(o1, o2);
+
+    public static int compare(byte[] o1, byte[] o2) {
+//            if (o1 == o2) return 0;
```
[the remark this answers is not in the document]
kushti: removed

avldb/src/main/java/scorex/ByteUtils.java
kushti: actually the method was not used, removed it

benchmarks/src/test/scala/org/ergoplatform/db/LDBVersionedStoreRollbackBench.scala
kushti: reverted

PR #1027, opened 2020-01-14
Candidate for 3.2.1

kushti:
Candidate for version 3.2.1.

It includes following PRs:

Ergo: 
#1012 and #1045 : mempool and miner made friendly to transactions spending outputs of offchain transactions
#1044 : pure-Java LevelDB dependency got back (to make possible to run the node on 32-bit platforms such as Raspberry Pi, see #1038 )
#1042: seed node list updates for both mainnet and testnet
#1020: IODB dependency finally removed 
#1019: configurable parameter for syncing detection during headers-chain download

Scorex: 

    ScorexFoundation/Scorex#361 - Circe updated to 0.9.0
    ScorexFoundation/Scorex#359 - keep-alive flag removed from socket options
    ScorexFoundation/Scorex#358 - improved banning of peers for sending inappropriate messages
    ScorexFoundation/Scorex#355 - terminate application if port to bind is in use


src/main/resources/reference.conf
```diff
@@ -61,6 +61,17 @@ ergo {
 
     # Minimal fee amount of transactions mempool accepts
     minimalFeeAmount = 1000000
+
+    # The node is downloading headers first and only then full blocks. Depending on settings, the node downloads whether
```
aslesarenko:
```suggestion
    # The node is downloading headers first and only then full blocks. Depending on settings, the node downloads either
```

aslesarenko: @kushti also consider switching to sigma 3.2.0

Issue #1033, opened 2020-01-17
Interpreter context is not reentrant


knizhnik:
sigmastate.eval.IRContext is not reentrant but it is used in ErgoTransaction.validateStateful which can be called concurrently from multiple threads.

PR #1034, opened 2020-01-17
[refer #1033] Fix race condition in access IRContext in ErgoTransaction.validateStateful

knizhnik:


src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -121,102 +121,103 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
-                  validationState.validate(txAssetsPreservation,
-                    inAmount >= outAmount || (outAssetId == newAssetId && outAmount > 0),
-                    s"$id: Amount in = $inAmount, out = $outAmount. Allowed new asset = $newAssetId, out = $outAssetId")
-              }
-              .payload(newCost)
-          case Failure(e) =>
-            // should never be here as far as we've already checked this when we've created the box
-            ModifierValidator.fatal(e.getMessage)
+    verifier.synchronized {
+      verifier.IR.resetContext() // ensure there is no garbage in the IRContext
```
aslesarenko: I gave it a second thought and this line kind of invalidate this solution. 
It is impossible to share the same context between two thread, not only becase of race conditions, but
also because this `resetContext` may destroy other thread's intermediate work.
So the solution is to have separate verifiers for each thread. 
It will be faster to have two separate threads working on independent contexts. (no sync, no false sharing, separate hash tables, etc.)

knizhnik: Я ещё раз поизучал код и если я гтчегло не упкстил, то едиственным источником гонок является emptyVerifier, который используется в getTxCost в ValidBlocksGenerator.
Таком образом - это всё таки проблема только теста.
И проще всего её исправить, сделав emptyVerifier функцией вместо lazy val.

aslesarenko: As explained by @catena2w, the script evaluation can be executed from two parallel threads. 
So the problem here is not related to tests. 
The problem is with current production code, which may be not safe.
So, it is necessary to guarantee:
1) that this specific verifier is not used in other methods (i.e. used ONLY in this method)
2) The usage in this method is serialized, i.e. if this method is called in parallel, then invocations are strictly serialized.

knizhnik: Script may be evaluated by different threads, but there will be race condition only if them are sharing the same verifier instance. Verifier is passed in validateStateful function as implicit parameter. I have investigated all passes through which this function can be called:
In the following places  new instance of verifier is always created:

* ErgoState.execTransactions
* ErgoMiner.collectTxs
* UtxoStateReader.validateWithCost

The only exception is ValidBlockGenerators.getTxCost which uses ErgoTestConstants.emptyVerifier.

This is why I have made the conclusion that synchronization problems exists only in UtxoTestSpecification parallel test. And it can be fixed by generating new instance of verifier at each access of emptyVerifier.

If you do not agree, please point me at the path where the same verifier instance can be shared between two threads.

aslesarenko: @knizhnik Well, in this case I don't have reasons to disagree. This was the task, either to fix problem or prove that there is nothing to fix. And your arguments above lead me to the conclusion that there is nothing to fix.
So I'm in favor of accepting this PR.

PR #1036, opened 2020-01-19
Transaction cost calculation fixed.

### LINKED ISSUE #1035: [BUG] Incorrect calculation of cost in ergo-wallet
### Problem
Ergo-wallet is computing transaction cost incorrectly. 
1) This https://github.com/ergoplatform/ergo/blob/8524b0ac85b27d480768f6e81e6102cf9d7dc096/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoProvingInterpreter.scala#L55
should be moved outside the loop
2) current accumulated cost should be passed as initCost argument of Context

oskin1:
Close #1035


kushti:
Fix copied into #1118 , closing this.

PR #1045, opened 2020-02-04
Making memory pool and miner friendly to transactions spending off-chain boxes

kushti:
This PR is a preparation for deployment of #1032 (which allows to create transactions spending outputs of offchain transactions) and contains certain modification to the memory pool and miner logic to support storing transactions spending outputs of mempool transactions and packing them into the blocks.

### DIFF
--- a/src/main/scala/org/ergoplatform/local/CleanupWorker.scala
+++ b/src/main/scala/org/ergoplatform/local/CleanupWorker.scala
@@ -58,10 +58,12 @@ class CleanupWorker(nodeViewHolderRef: ActorRef,
   private def validatePool(validator: TransactionValidation[ErgoTransaction],
                            mempool: ErgoMemPoolReader): Seq[ModifierId] = {
 
+    //internal loop function validating transactions, returns validated and invalidated transaction ids
     @tailrec
     def validationLoop(txs: Seq[ErgoTransaction],
+                       validated: Seq[ModifierId],
                        invalidated: Seq[ModifierId],
-                       etAcc: Long): Seq[ModifierId] = {
+                       etAcc: Long): (Seq[ModifierId], Seq[ModifierId]) = {
       txs match {
         case head :: tail if etAcc < nodeSettings.mempoolCleanupDuration.toNanos && !validatedIndex.contains(head.id) =>
 
@@ -76,25 +78,27 @@ class CleanupWorker(nodeViewHolderRef: ActorRef,
           val validationResult = state.validate(head)
           val t1 = System.nanoTime()
           val accumulatedTime = etAcc + (t1 - t0)
+
+          val txId = head.id
           validationResult match {
-            case Success(_) => validationLoop(tail, invalidated, accumulatedTime)
+            case Success(_) =>
+              validationLoop(tail, validated :+ txId, invalidated, accumulatedTime)
             case Failure(e) =>
-              log.debug(s"Transaction ${head.id} invalidated: ${e.getMessage}")
-              validationLoop(tail, invalidated :+ head.id, accumulatedTime)
+              log.debug(s"Transaction $txId invalidated: ${e.getMessage}")
+              validationLoop(tail, validated, invalidated :+ txId, accumulatedTime)
           }
         case _ :: tail if etAcc < nodeSettings.mempoolCleanupDuration.toNanos =>
           // this transaction was validated earlier, skip it
-          validationLoop(tail, invalidated, etAcc)
+          validationLoop(tail, validated, invalidated, etAcc)
         case _ =>
-          invalidated
+          validated -> invalidated
       }
     }
 
     // Check transactions sorted by priority. Parent transaction comes before its children.
     val txsToValidate = mempool.getAllPrioritized.toList
 
-    val invalidatedIds = validationLoop(txsToValidate, Seq.empty, 0L)
-    val validatedIds = txsToValidate.map(_.id).filterNot(invalidatedIds.contains)
+    val (validatedIds, invalidatedIds) = validationLoop(txsToValidate, Seq.empty, Seq.empty, 0L)
 
     epochNr += 1
     if (epochNr % CleanupWorker.IndexRevisionInterval == 0) {
--- a/src/main/scala/org/ergoplatform/local/CleanupWorker.scala
+++ b/src/main/scala/org/ergoplatform/local/CleanupWorker.scala
@@ -28,10 +28,16 @@ class CleanupWorker(nodeViewHolderRef: ActorRef,
   // count validation sessions in order to perform index cleanup.
   private var epochNr: Int = 0
 
+  override def preStart(): Unit = {
+    log.info("Cleanup worker started")
+  }
+
   override def receive: Receive = {
     case RunCleanup(validator, mempool) =>
       runCleanup(validator, mempool)
       sender() ! CleanupDone
+
+    case a: Any => log.warn(s"Strange input: $a")
   }
 
   private def runCleanup(validator: TransactionValidation[ErgoTransaction],
@@ -54,35 +60,36 @@ class CleanupWorker(nodeViewHolderRef: ActorRef,
     @tailrec
     def validationLoop(txs: Seq[ErgoTransaction],
                        invalidated: Seq[ModifierId],
-                       etAcc: Long): Seq[ModifierId] = txs match {
-      case head :: tail if etAcc < nodeSettings.mempoolCleanupDuration.toNanos
-        && !validatedIndex.contains(head.id) =>
+                       etAcc: Long): Seq[ModifierId] = {
+      txs match {
+        case head :: tail if etAcc < nodeSettings.mempoolCleanupDuration.toNanos && !validatedIndex.contains(head.id) =>
 
-        // Take into account previously validated transactions from the pool.
-        // This provides possibility to validate transactions which are spending off-chain outputs.
-        val state = validator match {
-          case u: UtxoStateReader => u.withTransactions(txs)
-          case _ => validator
-        }
+          // Take into account previously validated transactions from the pool.
+          // This provides possibility to validate transactions which are spending off-chain outputs.
+          val state = validator match {
+            case u: UtxoStateReader => u.withTransactions(txs)
+            case _ => validator
+          }
 
-        val t0 = System.nanoTime()
-        val validationResult = state.validate(head)
-        val t1 = System.nanoTime()
-        val accumulatedTime = etAcc + (t1 - t0)
-        validationResult match {
-          case Success(_) => validationLoop(tail, invalidated, accumulatedTime)
-          case Failure(e) =>
-            log.debug(s"Transaction ${head.id} invalidated: ${e.getMessage}")
-            validationLoop(tail, invalidated :+ head.id, accumulatedTime)
-        }
-      case _ :: tail if etAcc < nodeSettings.mempoolCleanupDuration.toNanos =>
-        // this transaction was validated earlier, skip it
-        validationLoop(tail, invalidated, etAcc)
-      case _ =>
-        invalidated
+          val t0 = System.nanoTime()
+          val validationResult = state.validate(head)
+          val t1 = System.nanoTime()
+          val accumulatedTime = etAcc + (t1 - t0)
+          validationResult match {
+            case Success(_) => validationLoop(tail, invalidated, accumulatedTime)
+            case Failure(e) =>
+              log.debug(s"Transaction ${head.id} invalidated: ${e.getMessage}")
+              validationLoop(tail, invalidated :+ head.id, accumulatedTime)
+          }
+        case _ :: tail if etAcc < nodeSettings.mempoolCleanupDuration.toNanos =>
+          // this transaction was validated earlier, skip it
+          validationLoop(tail, invalidated, etAcc)
+        case _ =>
+          invalidated
+      }
     }
 
-    val txsToValidate = mempool.getAllPrioritized
+    val txsToValidate = mempool.getAllPrioritized.toList
 
     val invalidatedIds = validationLoop(txsToValidate, Seq.empty, 0L)
     val validatedIds = txsToValidate.map(_.id).filterNot(invalidatedIds.contains)
--- a/src/main/scala/org/ergoplatform/local/CleanupWorker.scala
+++ b/src/main/scala/org/ergoplatform/local/CleanupWorker.scala
@@ -58,6 +58,8 @@ class CleanupWorker(nodeViewHolderRef: ActorRef,
       case head :: tail if etAcc < nodeSettings.mempoolCleanupDuration.toNanos
         && !validatedIndex.contains(head.id) =>
 
+        // Take into account previously validated transactions from the pool.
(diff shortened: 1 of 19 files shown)


src/main/scala/org/ergoplatform/local/ErgoMiner.scala
```diff
@@ -322,20 +321,35 @@ class ErgoMiner(ergoSettings: ErgoSettings,
...
-    state.proofsForTransactions(txs).map { case (adProof, adDigest) =>
-      CandidateBlock(bestHeaderOpt, version, nBits, adDigest, adProof, txs, timestamp, extensionCandidate, votes)
+    state.proofsForTransactions(txs) match {
+      case Success((adProof, adDigest)) =>
+        Success(CandidateBlock(bestHeaderOpt, version, nBits, adDigest, adProof, txs, timestamp, extensionCandidate, votes))
+      case Failure(t: Throwable) =>
+        emissionTxOpt match {
+          case Some(emissionTx) =>
+            log.error("Failed to produce proofs for transactions, but emission box is found: ", t)
+            val fallbackTxs = Seq(emissionTx._1)
```
aslesarenko: Please add description, why emissionTx is used as a fallback.

src/main/scala/org/ergoplatform/local/MempoolAuditor.scala
```diff
@@ -71,19 +72,21 @@ class MempoolAuditor(nodeViewHolderRef: ActorRef,
       nodeViewHolderRef ! GetNodeViewChanges(history = false, state = true, mempool = true, vault = false)
```
aslesarenko: Seems like the message `log.warn(s"Cleanup worker failed during initialization with: $e")`  is not correct

src/main/scala/org/ergoplatform/local/MempoolAuditor.scala
```diff
@@ -71,19 +72,21 @@ class MempoolAuditor(nodeViewHolderRef: ActorRef,
...
 
     case ChangedMempool(mp: ErgoMemPoolReader) =>
-      stateReaderOpt.fold[Any](poolReaderOpt = Some(mp))(initiateCleanup(_, mp))
+      poolReaderOpt = Some(mp)
```
aslesarenko: This was a bug?

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -30,7 +31,7 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
 
   override def getAll(ids: Seq[ModifierId]): Seq[ErgoTransaction] = ids.flatMap(pool.get)
 
-  override def getAllPrioritized: Seq[ErgoTransaction] = pool.orderedTransactions.values.toSeq.reverse
+  override def getAllPrioritized: Seq[ErgoTransaction] = pool.orderedTransactions.values.toSeq
```
aslesarenko: So it was a bug? Please add ScalaDocs to avoid confusion in future.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -72,6 +67,8 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
             new ErgoMemPool(pool.invalidate(tx)) -> ProcessingOutcome.Invalidated(_),
             _ => new ErgoMemPool(pool.put(tx)) -> ProcessingOutcome.Accepted
           )
+        // transaction validation currently works only for UtxoState, so this branch currently
+        // will not be triggered probably
```
aslesarenko: This comment belongs to which branch?

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolReader.scala
```diff
@@ -14,17 +15,25 @@ trait ErgoMemPoolReader extends MempoolReader[ErgoTransaction] {
 
   def getAll: Seq[ErgoTransaction]
 
-  /** Returns all transactions resided in pool sorted by priority
+  /**
+    * Returns all transactions resided in pool sorted by weight
```
aslesarenko: Ascending of descending?

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolReader.scala
```diff
@@ -14,17 +15,25 @@ trait ErgoMemPoolReader extends MempoolReader[ErgoTransaction] {
...
     */
   def getAllPrioritized: Seq[ErgoTransaction]
 
+
+  /**
+    * Returns given number of transactions resided in pool sorted by weight
```
aslesarenko: Sort direction?

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -118,9 +116,9 @@ final case class OrderedTxPool(orderedTransactions: TreeMap[WeightedTxId, ErgoTr
     * @param weight
```
aslesarenko: Why not  create a graph of transactions where edge X -> Y iff Y spends some output from X. 
And then update the graph as new transactions are added.
Then the algorithm will be fully deterministic and more flexible for future improvements and extensions.

src/main/scala/org/ergoplatform/local/CleanupWorker.scala
```diff
@@ -27,10 +28,16 @@ class CleanupWorker(nodeViewHolderRef: ActorRef,
...
+  override def preStart(): Unit = {
+    log.info("Cleanup worker started")
+  }
+
   override def receive: Receive = {
     case RunCleanup(validator, mempool) =>
       runCleanup(validator, mempool)
       sender() ! CleanupDone
+
+    case a: Any => log.warn(s"Strange input: $a")
```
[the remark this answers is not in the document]
kushti: The actor is not sending signals to self, thus a signal coming is not dependent on validatedIndex and epochNr

src/main/scala/org/ergoplatform/local/CleanupWorker.scala
```diff
@@ -49,38 +56,48 @@ class CleanupWorker(nodeViewHolderRef: ActorRef,
...
+          invalidated
+      }
     }
 
-    val mempoolTxs = mempool.getAll.toList
-    val txsToValidate = Random.shuffle(mempoolTxs)
+    val txsToValidate = mempool.getAllPrioritized.toList
+
     val invalidatedIds = validationLoop(txsToValidate, Seq.empty, 0L)
     val validatedIds = txsToValidate.map(_.id).filterNot(invalidatedIds.contains)
```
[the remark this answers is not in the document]
kushti: Yes, but then during a round transactions which are spending outputs from UTXO set and payinf most will be checked. And such transactions will be included into a next block and cleared out of the mempool after its processing.

With random shuffle the problem is that we can first check child transaction and throw it away. By how prioritization works, a parent transaction in the pool comes before its children.

src/main/scala/org/ergoplatform/local/ErgoMiner.scala
kushti: Done

src/main/scala/org/ergoplatform/local/MempoolAuditor.scala
kushti: You comment has no connection with the line of the code.

    case e: ActorInitializationException =>
      log.warn(s"Cleanup worker failed during initialization with: $e")
      Stop

is correct

src/main/scala/org/ergoplatform/local/MempoolAuditor.scala
kushti: I guess so

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: Not bug, but sorting now done in descending order, so need for reverse

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: The branch below , added line break.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolReader.scala
kushti: Desc improved

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolReader.scala
kushti: Desc improved

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
kushti: Makes sense, but @knizhnik did not take this approach. Let's make low-priority issue on that?

kushti:
@aslesarenko I've  addressed you comments, please make another pass !

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
knizhnik: We actually have such graph now! The only question is how edges of this graph are represented. Right now there are no direct links between transaction. Instead of it we have list of input/output boxes IDs in transaction and map of box-id to transaction which produced it. 

Certainly storing direct links can eliminate extra map lookups and speed-up algorithm. The question is how critical it is...   When I setup new node and load about 100k transactions, only <10 of them are off-chain. 

So may be there some other use cases where there are much more off-chain transactions. In any case before optimizing this code, we should first prove that it is a bottleneck.

Also just having materialized graph doesn't help itself to order transactions by priority, because we should take in account not only weight of transaction itself, but also weight of all dependent transactions.

src/main/scala/org/ergoplatform/local/CleanupWorker.scala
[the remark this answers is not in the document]
aslesarenko: Ok, just a suggestion of better debugging.

src/main/scala/org/ergoplatform/local/CleanupWorker.scala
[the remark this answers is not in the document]
aslesarenko: I feel like the pool of transactions should really be a graph of transactions.

src/main/scala/org/ergoplatform/local/MempoolAuditor.scala
aslesarenko: Then maybe add tests to check the internal state is changed consistently.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
aslesarenko: Well, why not make it clear with ScalaDoc?

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
aslesarenko: I think it is better to put comment IN the scope they belong to.

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
aslesarenko: I suggested to have graph explicitly not for optimization but for clarity. The dependencies are strict constraints which you cannot violate, the priorities are some additional weighting above the graph. When you prioritize and maybe shuffle around the pool you should strictly respect the graph. Once we have dependencies in the pool (which is a set of items) which we should respect, all algorithms over pool should respect and can utilize the graph. So why not have it explicitly. Again, not for the sake of performance, but for clarity. The performance will be a byproduct of this.

src/main/scala/org/ergoplatform/local/CleanupWorker.scala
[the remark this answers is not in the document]
aslesarenko: I mean why don't put in the message as much info as have at the point of failure?
The comment line `Should not be here, if non-expected signal comes, check logic` can be put in the message and you will have self-describing code.

src/main/scala/org/ergoplatform/local/CleanupWorker.scala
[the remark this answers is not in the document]
aslesarenko: Then `validatedIds` is missleading, since not all ids in the list were validated.

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
kushti: I've made an issue about that https://github.com/ergoplatform/ergo/issues/1051

src/main/scala/org/ergoplatform/local/CleanupWorker.scala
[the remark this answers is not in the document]
kushti: Good point, changed validationLoop() signature so now it returns "validatedIds". Please check the changes.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: Comment was done for a method definition in the base class, copied scaladoc to the concrete definition also

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: ok, done

src/main/scala/org/ergoplatform/local/MempoolAuditor.scala
kushti: Made new "rebroadcast transactions correctly" test in MempoolAuditorSpec, where a MempoolAuditor instance is receiving ChangedState and ChangedMempool signals, and then we're checking that it is doing cleanup and rebroadcasting transactions indeed.

src/test/scala/org/ergoplatform/local/MempoolAuditorSpec.scala
```diff
@@ -63,4 +71,55 @@ class MempoolAuditorSpec extends FlatSpec with NodeViewTestOps with ErgoTestHelp
...
+
+      override def take(limit: Int): Iterable[ErgoTransaction] = txs.take(limit)
+    }
+
+    implicit val system = ActorSystem()
+    val probe = TestProbe()
+
+    val auditor: ActorRef = TestActorRef(new MempoolAuditor(probe.ref, probe.ref, settingsToTest))
+
+    val coin = Random.nextBoolean()
```
aslesarenko: Maybe change the test into a property with coin obtained from a generator?

PR #1064, opened 2020-03-05
NIPoPoW Verification

gtklocker:


src/main/scala/org/ergoplatform/local/PoPoWVerifier.scala
```diff
@@ -0,0 +1,24 @@
...
+
+    def bestChain(): Seq[Header] = {
+        bestProof.map(_.headersChain).getOrElse(Seq())
+    }
+
+    def process(newProof: PoPowProof) {
+        if (newProof.headersChain.head.id == genesisId) {
+            // TODO: order matters here and it shouldn't
+            // only isBetterThan's argument is checked for validity
+            if (!bestProof.map(_.isBetterThan(newProof)).getOrElse(false)) {
```
kushti: So if no bestProof is set, non-valid one could be accepted?

src/main/scala/org/ergoplatform/modifiers/history/popow/PoPowProof.scala
```diff
@@ -44,12 +44,34 @@ case class PoPowProof(m: Int,
...
-    val (thisDivergingChain, thatDivergingChain) = lowestCommonAncestor(headersChain, that.headersChain)
-      .map(h => headersChain.filter(_.height > h.height) -> that.headersChain.filter(_.height > h.height))
-      .getOrElse(headersChain -> that.headersChain)
-    bestArg(thisDivergingChain)(m) > bestArg(thatDivergingChain)(m)
+    if (that.isValid()) {
+      lowestCommonAncestor(headersChain, that.headersChain)
+        .map(h => headersChain.filter(_.height > h.height) -> that.headersChain.filter(_.height > h.height))
+        .map({ case (thisDivergingChain, thatDivergingChain) =>
+          bestArg(thisDivergingChain)(m) > bestArg(thatDivergingChain)(m) })
+        .getOrElse(false)
```
kushti: IDEA is proposing to replace map + getOrElse with exists

kushti:
@gtklocker looks good, but please address the comments. Also please comment the code, e.g. ScalaDoc for methods, descriptions of non-trivial steps.

src/main/scala/org/ergoplatform/local/PoPoWVerifier.scala
gtklocker: If no bestProof is set, any proof that has a correct genesis is accepted is what this code is saying.

PR #1072, opened 2020-03-24
Candidate for 3.2.2

kushti:
Candidate for version 3.2.2

It includes following PRs:

* #1068 : Fix bug at restore wallet form at node interface
* #1060 : 
Fix for "calculated and received parameters differ in parameter" in suffix bootstrapping
* #1055 : Fix for wrong status after unlock attempt with wrong password 
* #1048 : Logback-test.xml removed from main application
* #1049  : Fix for improper block extension in ExpirationSpecification tests  
* #1066 : Add -release to avldb build, streamline travis setup
* #1069 : Sources for the papers "On Contractual Money" & "Bypassing NO PoW Schemes..."


greenhat:
> I wonder why in build.sbt for alvdb target we always add -release 8 commands line option and for avldb_benchmark do it only if scalaBinaryVersion.value != "2.11"
@knizhnik Not for `avldb_benchmark`, but for `ergoWallet` because it's the only project build for 2.11. Diff cut-off is a bit misleading.

Issue #940, opened 2019-10-10
[API] Specify secrets while creating transaction


scalahub:
In ErgoMix, for example, we require that certain public keys are used only once. In particular, security is lost if Bob uses the same public key in any other transaction. Thus, Bob's public key in the mix-step can be used only once. (the leak does not require spending from the reused key. Just using it to receive funds reveals the secret bit b).

Current API does not allow us to use arbitrary secrets to sign transactions. Rather it uses existing private keys as secrets. If we let Bob use these secrets in mixing, then the wallet *may* reuse the address in a different transaction. 

Hence, ideally, we should not be required to use secrets generated via the wallet. Rather we should be allowed to use arbitrary secrets generated outside the wallet.

One way to do this would be to extend the /wallet/transaction/generate by adding yet another array:

	{
	  "requests": [...],
	  "fee": 1000000,
	  "inputsRaw": [
	    "1a2b", "2b3c"
	  ],
          ...
	  "secrets": ["12314", "4567"]
	}

Perhaps secrets are better encoded in hex? I presume they are always BigInts.

kushti:
Secrets are big ints (mod q) indeed, how signer should use them though?

scalahub:
I think it should be done in "use and forget" mode.

jasondavies:
I think Bob has to use their secret twice: once to create the full mix box, and then again to spend their output?

jasondavies:
I've implemented a much more general transaction signing API that takes a full unsigned transaction and attempts to sign it.  The resulting signed transaction can then be POSTed to /transactions to broadcast it.  The reason is that I wanted to select the boxes to spend myself.

This API doesn't even really need a wallet (though ideally it should still use wallet secrets if available).  I'm now working on encoding secrets to be sent with the transaction for signing (currently just hard coded).

Of course, it's not really ideal to send secrets via the API due to the possibility of eavesdropping, but in ErgoMix, Bob needs access to their secret to calculate g^xy.

Issue #901, opened 2019-09-11
API: /wallet/sign (or /wallet/prove?)


jasondavies:
Motivation: this would allow external applications to build an unsigned transaction, serialise to a byte array and then send it to the node for signing. The resulting proof can then be used in `spendingProof.proofBytes` for a particular box (the API would need to be called multiple times for multiple keys).

PR #1085, opened 2020-04-09
Allow signing arbitrary transactions with arbitrary private inputs (DLog and DHT)

jasondavies:
I'm not really that familiar with Scala, so I would appreciate some help parsing the JSON:

```json
{
  "tx": ErgoTransaction,
  "secrets": {
    "dlog": ["base16" (32 bytes BigInt), ...], // DLogProverInput(SigmaDsl.BigInt(new BigInteger(1, Base16.decode("base16").get)))
    "dht": [["base16" (32 bytes BigInt), ["base16" (33 bytes compressed GroupElement), "same", "same", "same"]], ...] // DiffieHellmanTupleProverInput(..., ProveDHTuple(...))
  }
}
```

I've tested this with hard-coded private inputs and it works.

I also imagine there may be a better way to handoff secrets; I had to create my own `SecretsProvingInterpreter` class because I didn't want to mess around with the stuff in dependencies (sigmastate-interpreter).

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.2.8
+sbt.version=1.3.2
--- a/src/main/scala/org/ergoplatform/SecretsProvingInterpreter.scala
+++ b/src/main/scala/org/ergoplatform/SecretsProvingInterpreter.scala
@@ -1,0 +1,71 @@
+package org.ergoplatform
+
+import java.util
+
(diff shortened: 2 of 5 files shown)


jasondavies:
Related #901 and #940.

kushti:
Reworked in #1089 , this PR not needed anymore

PR #1086, opened 2020-04-09
Data inputs added to API methods /wallet/transaction/{generate, sent}

kushti:
This PR adds "dataInputsRaw" field to /wallet/transaction/{generate, sent} API methods inputs and corresponding processing for provided data inputs.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -419,18 +419,18 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
     */
   private def generateTransactionWithOutputs(requests: Seq[TransactionRequest],
                                              inputsRaw: Seq[String],
-                                             dataInputsRaw: Seq[String]): Try[ErgoTransaction] =
+                                             dataInputsRaw: Seq[String]): Try[ErgoTransaction] = Try {
     proverOpt match {
       case Some(prover) =>
-         val inputs = inputsRaw
-                .toList
-                .map(in => Base16.decode(in).flatMap(ErgoBoxSerializer.parseBytesTry))
-                .map(_.get) //todo .get
+        val inputs = inputsRaw
+          .toList
+          .map(in => Base16.decode(in).flatMap(ErgoBoxSerializer.parseBytesTry))
+          .map(_.get)
 
-         val dataInputs = dataInputsRaw
-                .toIndexedSeq
-                .map(in => Base16.decode(in).flatMap(ErgoBoxSerializer.parseBytesTry))
-                .map(_.get) //todo .get
+        val dataInputs = dataInputsRaw
+          .toIndexedSeq
+          .map(in => Base16.decode(in).flatMap(ErgoBoxSerializer.parseBytesTry))
+          .map(_.get)
 
         requestsToBoxCandidates(requests).flatMap { payTo =>
           require(prover.pubKeys.nonEmpty, "No public keys in the prover to extract change address from")
@@ -466,7 +466,7 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
 
           val selectionOpt = boxSelector.select(inputBoxes, filter, targetBalance, targetAssets)
 
-          selectionOpt.map {selectionResult =>
+          selectionOpt.map { selectionResult =>
             prepareTransaction(prover, payTo, selectionResult, dataInputs)
           } match {
             case Right(txTry) => txTry.map(ErgoTransaction.apply)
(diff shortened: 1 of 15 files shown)


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -418,68 +418,76 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
...
+        val dataInputs = dataInputsRaw
+          .toIndexedSeq
+          .map(in => Base16.decode(in).flatMap(ErgoBoxSerializer.parseBytesTry))
+          .map(_.get)
+
+        requestsToBoxCandidates(requests).flatMap { payTo =>
+          require(prover.pubKeys.nonEmpty, "No public keys in the prover to extract change address from")
+          require(requests.count(_.isInstanceOf[AssetIssueRequest]) <= 1, "Too many asset issue requests")
+          require(payTo.forall(c => c.value >= BoxUtils.minimalErgoAmountSimulated(c, parameters)), "Minimal ERG value not met")
+          require(payTo.forall(_.additionalTokens.forall(_._2 >= 0)), "Negative asset value")
```
greenhat, line 439: Is token amount == 0 allowed?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -418,68 +418,76 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
...
+
+          val targetBalance = payTo
+            .map(_.value)
+            .sum
+
+          val targetAssets = payTo
+            .filterNot(bx => assetIssueBox.contains(bx))
+            .foldLeft(Map.empty[ModifierId, Long]) { case (acc, bx) =>
+              // TODO optimize: avoid toArray and use mapFirst
+              val boxTokens = bx.additionalTokens.toArray.map(t => bytesToId(t._1) -> t._2).toMap
```
greenhat, line 455: There is the same code in wallet: 
https://github.com/ergoplatform/ergo/blob/f471e448b131d791cc4061f3168e97eb8a1517f0/ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala#L30-L31

Maybe it's time to extract and re-use it. Also `AssetUtils` can be moved to wallet and used in place of 
https://github.com/ergoplatform/ergo/blob/f471e448b131d791cc4061f3168e97eb8a1517f0/ergo-wallet/src/main/scala/org/ergoplatform/wallet/boxes/BoxSelector.scala#L55-L78

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -418,68 +418,76 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
...
-                //inputs are provided externally, no need for filtering
-                (boxesToFakeTracked(inputs), noFilter)
-              } else {
-                //inputs are to be selected by the wallet
-                (registry.readCertainUnspentBoxes.toIterator, onChainFilter)
-              }
+        val inputs = inputsRaw
+          .toList
+          .map(in => Base16.decode(in).flatMap(ErgoBoxSerializer.parseBytesTry))
+          .map(_.get)
```
aslesarenko, line 428: Instead of `get` I suggest to use an extension method [`getOrThrow`](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/829f5da7698c1b732b20db26bf5ca6f79d343c4f/sigmastate/src/main/scala/sigmastate/utils/Helpers.scala#L107) defined in sigma.
Search for `TryOps` implicit class.

It will propagate exception up the Try hierarchy

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -418,68 +418,76 @@ class ErgoWalletActor(settings: ErgoSettings, boxSelector: BoxSelector)
...
+
+        val dataInputs = dataInputsRaw
+          .toIndexedSeq
+          .map(in => Base16.decode(in).flatMap(ErgoBoxSerializer.parseBytesTry))
+          .map(_.get)
```
aslesarenko, line 433: use getOrThrow

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: @aslesarenko please provide clear motivation to switch. Also please note that the same approach is used throughout the class and the whole Ergo codebase, so make an issue if you're willing to change approach (it should be done not in the single place obviously)

aslesarenko: @kushti  Ok, seems like `get` will also throw the exception from `Failure`

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: Fixed (amount==0 not allowed by the protocol actually).

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: @greenhat done!

PR #1089, opened 2020-04-21
Transaction signing with secrets provided (/wallet/transaction/sign)

kushti:
This PR is introducing:
 * A hierarchy for secret keys, starting from basic SecretKey trait. The hierarchy combines primitive (DLOG/DHT) and HD(BIP32) keys  
 * ErgoProvingInterpreter modified to be friendly to the hierarchy above
 * New Wallet API method /wallet/transaction/signWithSecretsGiven to sign an arbitrary transaction with secrets provided (both DLOG and DHT)

Based on #1085 , related #901 and #940.

PR #1093, opened 2020-04-24
Candidate for version 3.2.3

kushti:
This candidate PR for version 3.2.3 is including following features and fixes:

* Transaction signing with secrets provided (/wallet/transaction/sign) (PR #1089 )
* Allow unconfirmed boxes to be spent via /transactions API method call (PR #1092 ) 
* /transactions/check API method to check a transaction without propagating it  (PR #1091 )
* Data inputs added to API methods /wallet/transaction/{generate, sent} ( PR #1086 )
* Setting up native LevelDB buffer (for better performance) ( PR #1073 )
* Fixing rounding issues in /panel (PR #1087)
* Implementation of transaction membership proof and extension element proof (PR #1076 #1081 )
* Extract abstract box selection and add transaction building (PR #1071 )
* Fixing some typos in openapi.yaml (PR #1088 )
* build.sbt: Upgrade an http resolver to https (PR #1098 ) 

Big thanks to all the contributors: @deadit , @greenhat , @gtklocker , @jasondavies , @knizhnik , @kushti


jasondavies:
The new `/transactions/check` API seems to actually send the transaction rather than just validate it.  The implementation looks like it's the same as the path for sending (checkTransactionR is the same as sendTransactionR).

src/main/scala/org/ergoplatform/http/api/WalletApiRoute.scala
```diff
@@ -53,6 +55,8 @@ case class WalletApiRoute(readersHolder: ActorRef, nodeViewActorRef: ActorRef, e
         deriveNextKeyR ~
         updateChangeAddressR
     }
+  } ~ pathPrefix("wallet") {
+    signTransactionR
```
jasondavies, line 59: Shouldn't this be under `toStrictEntity`?  I ran into a strange error, "Substream Source cannot be materialized more than once", but moving `signTransactionR` to be inside `toStrictEntity` seemed to fix the issue.

jasondavies: I'm not sure what the error means or why it was happening (maybe JSON data over a particular size?)

jasondavies: Also, just noticed that this should be protected by `withAuth` too.

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -48,32 +47,25 @@ case class TransactionsApiRoute(readersHolder: ActorRef, nodeViewActorRef: Actor
...
+          processFn(tx)
           ApiResponse(tx.id)
         }
       )
     }
   }
 
-  def getUnconfirmedTransactionsR: Route = (path("unconfirmed") & get & paging) { (offset, limit) =>
-    ApiResponse(getUnconfirmedTransactions(offset, limit))
+  def sendTransactionR: Route = (post & entity(as[ErgoTransaction])) { tx =>
```
jasondavies, line 57: I would suggest adding `pathEnd` to avoid possible confusion.

kushti: Added!

src/main/scala/org/ergoplatform/http/api/WalletApiRoute.scala
kushti: now /sign is protected, what and moved under toStrictEntity

kushti:
@jasondavies /check actually called send , some issue with routing, now fixed. All other issues fixed, thanks for reviewing!

src/main/scala/org/ergoplatform/http/api/WalletApiRoute.scala
```diff
@@ -122,32 +125,85 @@ case class WalletApiRoute(readersHolder: ActorRef, nodeViewActorRef: ActorRef, e
...
-  def generateTransactionR: Route = (path("transaction" / "generate") & post
-    & entity(as[RequestsHolder])) (holder => generateTransaction(holder.withFee, holder.inputsRaw))
+        if (boxesToSpend.size == tx.inputs.size && dataBoxes.size == tx.dataInputs.size) {
+          r.w.signTransaction(secrets, tx, boxesToSpend, dataBoxes)
+        } else {
+          Future(Failure(new Exception("Can't parse input boxes provided")))
+        }
+      } else {
+        r.s match {
+          case utxoSet: UtxoStateReader =>
```
jasondavies, line 183: Would be nice to fetch boxes from the mempool here too, so that chained transactions can be easily signed.

aslesarenko:
@kushti Since this is RC, full node resync test is required.

src/main/scala/org/ergoplatform/http/api/WalletApiRoute.scala
jasondavies: Looks like this is done now. Thanks!

kushti: Yes, done!

PR #1098, opened 2020-05-01
build.sbt: Upgrade http resolver to https

gtklocker:
Fixes error with sbt 1.3.x:

```
[warn] insecure HTTP request is deprecated 'http://repo.typesafe.com/typesafe/maven-releases/'; switch to HTTPS or opt-in as ("Typesafe maven releases" at "http://repo.typesafe.com/typesafe/maven-releases/").withAllowInsecureProtocol(true)
```


gtklocker:
Can we maybe prune it completely? Do we need all those resolvers?

PR #1104, opened 2020-05-14
Complexity limit raised to 250K

kushti:
This PR is raising default complexity limit per transaction to 250K.

### DIFF
--- a/src/main/scala/org/ergoplatform/settings/Constants.scala
+++ b/src/main/scala/org/ergoplatform/settings/Constants.scala
@@ -58,10 +58,13 @@ object Constants {
 
   // Maximum extension size
   val MaxExtensionSize: Int = 32 * 1024
+
   // Maximum extension size during bytes parsing. Allows to move MaxExtensionSize to Parameters in future
   val MaxExtensionSizeMax: Int = 1024 * 1024
+
   // Default limit for transaction complexity to be included into block.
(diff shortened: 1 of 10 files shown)


src/main/scala/org/ergoplatform/settings/Constants.scala
```diff
@@ -58,10 +58,13 @@ object Constants {
...
   // Maximum extension size
   val MaxExtensionSize: Int = 32 * 1024
+
   // Maximum extension size during bytes parsing. Allows to move MaxExtensionSize to Parameters in future
   val MaxExtensionSizeMax: Int = 1024 * 1024
+
   // Default limit for transaction complexity to be included into block.
-  val DefaultComplexityLimit: Int = 100000
+  // Used in tests only.
+  val DefaultComplexityLimit: Int = 250000
```
greenhat, line 67: Can we move it to `test` scope? Maybe to `ErgoTestConstants`.

aslesarenko: @kushti shouldn't test use the value form reference.conf file?

kushti: @aslesarenko good catch! removed DefaultComplexityLimit totally, now reading from "initSettings" in tests

Issue #747, opened 2019-05-28
Wallet consistency check on startup


oskin1:

kushti:
done in 3.3.0

PR #1149, opened 2020-07-29
Check Wallet scan consistency and fix if the wallet state is inconsistent

### LINKED ISSUE #1063: Wallet: "Blocks were skipped during wallet scanning"
Currently if a new block is coming, the wallet could skip validating postponed blocks. In 3.2.1 this error is caught with log output like: 

14:53:38.694 INFO  [ctor.default-dispatcher-6] o.e.n.w.ErgoWalletActor - Processing block at height 156836 with 0 wallet transactions
14:53:40.666 ERROR [ctor.default-dispatcher-6] o.e.n.w.p.WalletRegistry - Blocks were skipped during wallet scanning, from 3974 until 156836

kushti:
Close #1063  and #747 : 

this PR is introducing consistency checks for blocks application in the wallet. If a block comes out-of-order, the wallet now ignores it then asking for proper blocks if they're available and applies them.

In addition, this PR also allows for easy upgrade from 3.2.x to 3.3.x by just removing wallet/registry folder in the data folder. Then on the first full block arrived the full wallet rescan will be done.

### DIFF
--- a/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
+++ b/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
@@ -81,7 +81,7 @@ class DeepRollBackSpec extends FreeSpec with IntegrationSuite {
       minerBBestBlock shouldEqual minerABestBlock
     }
 
-    Await.result(result, 15.minutes)
+    Await.result(result, 25.minutes)
 
   }
 
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -121,9 +121,9 @@ class ErgoWalletActor(settings: ErgoSettings,
           case Some(block) =>
             scanBlock(block)
           case None =>
-            // We may do not have a block if, for example, the blockchain is pruned. This is okay, just skip it.
+          // We may do not have a block if, for example, the blockchain is pruned. This is okay, just skip it.
         }
-        if(blockHeight < fullHeight) {
+        if (blockHeight < fullHeight) {
           self ! ScanInThePast(blockHeight + 1)
         }
       }
@@ -355,7 +355,7 @@ class ErgoWalletActor(settings: ErgoSettings,
       sender() ! RemoveScanResponse(res)
 
     case AddScan(appRequest) =>
-      val res: Try[Scan] = storage.addScan(appRequest, fullHeight + 1)
+      val res: Try[Scan] = storage.addScan(appRequest)
       res.foreach(app => walletVars = walletVars.addScan(app))
       sender() ! AddScanResponse(res)
 
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -68,7 +68,7 @@ class ErgoWalletActor(settings: ErgoSettings,
   private def stateContext: ErgoStateContext = storage.readStateContext
 
   /**
-    * Height of full-blocks chain as reported by the state
+    * Height of the chain as reported by the state
     * (i.e. height of a last block applied to the state, not the wallet)
     * Wallet's height may be behind it.
     */
@@ -150,7 +150,7 @@ class ErgoWalletActor(settings: ErgoSettings,
         log.warn(s"Wallet: skipped blocks found starting from $expectedHeight, going back to scan them")
         self ! ScanInThePast(expectedHeight)
       } else {
-        log.warn(s"Wallet: $expectedHeight > block.height for a block at ${block.height}, blockId: ${block.id}")
+        log.warn(s"Wallet: block in the past reported at ${block.height}, blockId: ${block.id}")
       }
 
     case Rollback(version: VersionTag) =>
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -300,7 +300,8 @@ class ErgoWalletActor(settings: ErgoSettings,
       secretStorageOpt.foreach(_.lock())
 
     case GetWalletStatus =>
-      val status = WalletStatus(secretIsSet, walletVars.proverOpt.isDefined, changeAddress)
+      val lastScannedHeight = registry.fetchDigest().height
+      val status = WalletStatus(secretIsSet, walletVars.proverOpt.isDefined, changeAddress, lastScannedHeight)
       sender() ! status
 
     case GenerateTransaction(requests, inputsRaw, dataInputsRaw) =>
@@ -811,8 +812,12 @@ object ErgoWalletActor {
     * @param initialized   - whether wallet is initialized or not
     * @param unlocked      - whether wallet is unlocked or not
     * @param changeAddress - address used for change (optional)
+    * @param height        - last height scanned
     */
-  case class WalletStatus(initialized: Boolean, unlocked: Boolean, changeAddress: Option[P2PKAddress])
+  case class WalletStatus(initialized: Boolean,
+                          unlocked: Boolean,
+                          changeAddress: Option[P2PKAddress],
+                          height: ErgoHistory.Height)
 
   /**
     * Get root secret key (used in miner)
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -122,12 +122,14 @@ class ErgoWalletActor(settings: ErgoSettings,
     //scan block transactions
     case ScanOnChain(block) =>
       val lastHeight = registry.fetchDigest().height
-      if(lastHeight + 1 == block.height) {
+      val expectedHeight = lastHeight + 1
+      if (expectedHeight == block.height) {
         scanBlock(block)
-      } else if (lastHeight + 1 < block.height) {
-        ScanInThePast(lastHeight + 1, wallet.Constants.PaymentsScanId)
+      } else if (expectedHeight < block.height) {
+        log.warn(s"Wallet: skipped blocks found starting from $expectedHeight, going back to scan them")
+        ScanInThePast(expectedHeight, wallet.Constants.PaymentsScanId)
       } else {
-        log.warn(s"lastHeight + 1 > block.height for a block at ${block.height}, blockId: ${block.id}")
+        log.warn(s"Wallet: $expectedHeight > block.height for a block at ${block.height}, blockId: ${block.id}")
       }
 
     case Rollback(version: VersionTag) =>
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -96,6 +96,13 @@ class ErgoWalletActor(settings: ErgoSettings,
     }
   }
 
+  def scanBlock(block: ErgoFullBlock): Unit = {
+    val (reg, offReg) =
+      WalletScanLogic.scanBlockTransactions(registry, offChainRegistry, stateContext, walletVars, block)
+    registry = reg
+    offChainRegistry = offReg
+  }
+
   private def scanLogic: Receive = {
     //scan mempool transaction
     case ScanOffChain(tx) =>
@@ -103,12 +110,25 @@ class ErgoWalletActor(settings: ErgoSettings,
       val inputs = WalletScanLogic.extractInputBoxes(tx)
       offChainRegistry = offChainRegistry.updateOnTransaction(newWalletBoxes, inputs)
 
+    case ScanInThePast(blockHeight, scanId) =>
+      historyReader.bestFullBlockAt(blockHeight) match {
+        case Some(block) =>
+          //scan id ignored for now
+          scanBlock(block)
+          self ! ScanInThePast(blockHeight + 1, scanId)
+        case None =>
+      }
+
     //scan block transactions
     case ScanOnChain(block) =>
-      val (reg, offReg) =
-        WalletScanLogic.scanBlockTransactions(registry, offChainRegistry, stateContext, walletVars, block)
-      registry = reg
-      offChainRegistry = offReg
+      val lastHeight = registry.fetchDigest().height
+      if(lastHeight + 1 == block.height) {
+        scanBlock(block)
+      } else if (lastHeight + 1 < block.height) {
+        ScanInThePast(lastHeight + 1, wallet.Constants.PaymentsScanId)
+      } else {
+        log.warn(s"lastHeight + 1 > block.height for a block at ${block.height}, blockId: ${block.id}")
+      }
 
     case Rollback(version: VersionTag) =>
       registry.rollback(version).fold(
@@ -323,7 +343,7 @@ class ErgoWalletActor(settings: ErgoSettings,
       sender() ! RemoveScanResponse(res)
 
     case AddScan(appRequest) =>
-      val res: Try[Scan] = storage.addScan(appRequest)
+      val res: Try[Scan] = storage.addScan(appRequest, height + 1)
       res.foreach(app => walletVars = walletVars.addScan(app))
       sender() ! AddScanResponse(res)
 
@@ -629,6 +649,8 @@ object ErgoWalletActor {
     * @param block - block to scan
     */
   final case class ScanOnChain(block: ErgoFullBlock)
+
+  final case class ScanInThePast(blockHeight: ErgoHistory.Height, scanId: ScanId)
 
   /**
     * Rollback to previous version of the wallet, by throwing away effects of blocks after the version
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -57,7 +57,6 @@ class ErgoWalletActor(settings: ErgoSettings,
 
   private var walletVars = WalletVars.apply(storage, settings)
 
-
   //todo: temporary 3.2.x collection and readers
   private var stateReaderOpt: Option[ErgoStateReader] = None
   private var mempoolReaderOpt: Option[ErgoMemPoolReader] = None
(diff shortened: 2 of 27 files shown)


src/main/scala/org/ergoplatform/nodeView/wallet/WalletScanLogic.scala
```diff
@@ -65,6 +65,11 @@ object WalletScanLogic extends ScorexLogging {
                             blockId: ModifierId,
                             transactions: Seq[ErgoTransaction]): (WalletRegistry, OffChainRegistry) = {
 
+    //input tx id, input box id, tracked box
+    type InputData = Seq[(ModifierId, EncodedBoxId, TrackedBox)]
+    //outputs, input ids, related transactions
+    type ScanResults = (Seq[TrackedBox], InputData, Seq[WalletTransaction])
```
greenhat, line 71: I suggest to convert tuples into case classes to improve readability.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -68,7 +67,10 @@ class ErgoWalletActor(settings: ErgoSettings,
   // The state context is being updated by listening to state updates.
   private def stateContext: ErgoStateContext = storage.readStateContext
 
-  private var height: Int = ErgoHistory.GenesisHeight
+  /**
+    * Height of full-blocks chain as reported by the state processor
```
aslesarenko, line 71: Please define `full-blocks chain` or give reference to the definition.
Also, what is `state processor`, if it is a class use '[[ClassReference]]` syntax of ScalaDoc. If not, then please define.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -96,19 +98,48 @@ class ErgoWalletActor(settings: ErgoSettings,
     }
   }
 
+  def scanBlock(block: ErgoFullBlock): Unit = {
```
aslesarenko, line 101: Please add ScalaDoc in particular describing the effects of this method on the Wallet state.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -96,19 +98,48 @@ class ErgoWalletActor(settings: ErgoSettings,
...
+      val lastHeight = registry.fetchDigest().height
+      val expectedHeight = lastHeight + 1
+      if (expectedHeight == block.height) {
+        scanBlock(block)
+      } else if (expectedHeight < block.height) {
+        log.warn(s"Wallet: skipped blocks found starting from $expectedHeight, going back to scan them")
+        self ! ScanInThePast(expectedHeight)
+      } else {
+        log.warn(s"Wallet: $expectedHeight > block.height for a block at ${block.height}, blockId: ${block.id}")
+      }
```
aslesarenko, line 142: I suggest to describe all state transitions of the Wallet in details. I cannot get the logic from the code itself. What is `expectedHeight`? Expected by whom?

The wallet is state machine, hence all the states should be described and the logic of all state changes should be also documented.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -96,19 +98,48 @@ class ErgoWalletActor(settings: ErgoSettings,
...
-      offChainRegistry = offReg
+      val lastHeight = registry.fetchDigest().height
+      val expectedHeight = lastHeight + 1
+      if (expectedHeight == block.height) {
+        scanBlock(block)
+      } else if (expectedHeight < block.height) {
+        log.warn(s"Wallet: skipped blocks found starting from $expectedHeight, going back to scan them")
+        self ! ScanInThePast(expectedHeight)
+      } else {
+        log.warn(s"Wallet: $expectedHeight > block.height for a block at ${block.height}, blockId: ${block.id}")
```
aslesarenko, line 141: What this message means?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -780,8 +814,12 @@ object ErgoWalletActor {
     * @param initialized   - whether wallet is initialized or not
     * @param unlocked      - whether wallet is unlocked or not
     * @param changeAddress - address used for change (optional)
+    * @param height        - last height scanned
     */
-  case class WalletStatus(initialized: Boolean, unlocked: Boolean, changeAddress: Option[P2PKAddress])
+  case class WalletStatus(initialized: Boolean,
+                          unlocked: Boolean,
+                          changeAddress: Option[P2PKAddress],
+                          height: ErgoHistory.Height)
```
aslesarenko, line 822: Why not name the field `lastScannedHeight`?

src/main/scala/org/ergoplatform/settings/NodeConfigurationSettings.scala
```diff
@@ -28,10 +28,17 @@ case class NodeConfigurationSettings(stateType: StateType,
                                      mempoolCleanupDuration: FiniteDuration,
                                      rebroadcastCount: Int,
                                      minimalFeeAmount: Long,
-                                     headerChainDiff: Int)
+                                     headerChainDiff: Int) {
+  /**
+    * Whether the node keeping all the full blocks of the blockchain or not.
+    * @return true if the blockchain is pruned, false if not
+    */
+  val fullBlocksPruned: Boolean = blocksToKeep >= 0
```
aslesarenko, line 36: The name sounds like the number of blocks pruned, not as the boolean flag. I suggest `isFullBlocksPruned`

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -68,7 +67,10 @@ class ErgoWalletActor(settings: ErgoSettings,
   // The state context is being updated by listening to state updates.
   private def stateContext: ErgoStateContext = storage.readStateContext
 
-  private var height: Int = ErgoHistory.GenesisHeight
+  /**
+    * Height of full-blocks chain as reported by the state processor
+    */
+  private var fullHeight: Int = ErgoHistory.EmptyHistoryHeight
```
aslesarenko, line 73: The name is not descriptive enough. What the `full` part means?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -271,7 +302,8 @@ class ErgoWalletActor(settings: ErgoSettings,
       secretStorageOpt.foreach(_.lock())
 
     case GetWalletStatus =>
-      val status = WalletStatus(secretIsSet, walletVars.proverOpt.isDefined, changeAddress)
+      val lastScannedHeight = registry.fetchDigest().height
```
aslesarenko, line 305: This seems like a better name instead of `fullHeight`. But maybe I just don't understand the difference, in which case the code need readability improvements.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: @aslesarenko it means height of the full-blocks chain (not the headers-chain), actually, it is already written in the comment above.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: I think this is obvious for an API client that height means current height of the wallet, i.e. height of the last block scanned.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: This message means wallet is in some strange state (got the block in the past to scan), thus the developers need to investigate what can be a reason for that.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: expectedHeight is height of a block expected by the wallet.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: No need for the definition of basic blockchain glossary here I suppose. Headers-chain and full-blocks chain are everywhere around the space, Ergo API and Ergo code. The historically first reference would be the Bitcoin whitepaper (SPV section of it), but I can hardly imagine anyone reading this code before the whitepaper.

kushti: On the state processor, changed the comment.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: done

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: On wallet as the state machine, do you mean the scan functionality?

src/main/scala/org/ergoplatform/settings/NodeConfigurationSettings.scala
kushti: done

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: externalized registry.fetchDigest().height calls into walletHeight(), made a comment

src/main/scala/org/ergoplatform/nodeView/wallet/WalletScanLogic.scala
kushti: reverted the class instead, modifying it is out of scope of the PR.

kushti:
@greenhat @aslesarenko answered or addressed the comments, please make another pass!

src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
```diff
@@ -81,7 +81,7 @@ class DeepRollBackSpec extends FreeSpec with IntegrationSuite {
       minerBBestBlock shouldEqual minerABestBlock
     }
 
-    Await.result(result, 15.minutes)
+    Await.result(result, 25.minutes)
```
aslesarenko, line 84: Looks like something wrong with it-tests, which will make them less robust over time.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -68,9 +67,21 @@ class ErgoWalletActor(settings: ErgoSettings,
   // The state context is being updated by listening to state updates.
   private def stateContext: ErgoStateContext = storage.readStateContext
 
-  private var height: Int = ErgoHistory.GenesisHeight
+  /**
+    * Height of full-blocks chain as reported by the state
```
aslesarenko, line 71:
```suggestion
    * Height of full-blocks chain (not the headers-chain) as reported by the state
```
Well, if you add `(not the headers-chain)` hit to the comments it will do the job.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: @kushti It doesn't make much sense to add clarifications here in conversations without changing the code itself. Any question actually means the code is not clear enough or/and require improvement (i.e. add comments). So when I ask the question about the meaning of the code (which I do from time to time) I usually expect to see the answer in the code.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: I mean state machine is characterized by state and how that state changes when something happens (new message arives etc). Without explicit description it often realy hard to understand even for the author after a couple of years has passed. 
It is better to describe it explicitly.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: Not obvious at all since we have more than one kind of height. Better be explicit and unambiguous especially for API.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: State can't report headers-chain height at all.

src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
kushti: checking that still, to be resolved in 3.3.0 branch maybe

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: Still , what kind of functionality are you talking about ? If scan, then this is quite simple, scan from the past, scan incoming block, or rollback.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: I think the answer is clear enough, the message provides devs heights and block id , along with other lines in the log it can be hopefully useful for investigations.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: In the wallet, there's one notion of height I guess. 
Please also note that in "lastScannedHeight" two first words are useless by the blockchain construction. The wallet's height is indeed last one, and that block is indeed scanned.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: This message in the log is cryptic. That is my point. It is better to give more descriptive message in the log, at least to easily search it in the log file.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: I mean documenting the whole ErgoWalletActor, as it represent the state machine.
I assume it is pretty clear what the documentation of state machine should be: 1) what are the states 2) what are the transitions

I suggest to write it down explicitly while you still clearly remember it so that others can understand it in future.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: Well, I give my feedback about code readability in places it is not clear for me. I think it will also be not clear for others.

Again, you better put your explanation above into the code directly. 
(Don't assume people think exactly the way you think, so better to me more explicit in the code)

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: Three word better describe the meaning than one. Why you don't want to be explicit, you are not sure about real meaning, or think it may change in future?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
aslesarenko: You know about it, I don't. And the code is for me to read it. So I gave you a hint where I cannot read it.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: It is better to avoid subjective views when reviewing, as the goal is to improve the code towards some inter-subjective goals. Then you need to clearly understand which group's inter-subjective views you do represent.
I suppose that the comments in the wallet are made not for a person doing a review by skimming through GitHub changes, rather for a person willing to investigate how the wallet works (maybe, in order to make changes then). Supposedly such a person already knows what basic things about Ergo architecture, node settings in regards with different regimes, maybe Scorex tutorial and Yellow paper draft.

Thus providing comments about sync or basic cryprocurrency jargon are just distracting from the goal. So I still do not see a reason for that.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: Out of scope of the PR then. There's a first issue on writing a test for EWA, so let's make an issue about documentation also. Unlikely the design would be about a single state machine though.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: Well, let me explain how the logs are being used (based on tens of cases for the first years of the mainnet existence). A user is sending logs (to me, as only me is doing support atm unfortunately), then I am looking ERROR messages, then WARN, then I look into the message to extract details (like ids) to find correlated messages.

In regards with the workflow, the message looks good to me. If you can propose improvements, please do. To be honest, I do not understand your intentions and concerns.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: I think my comment above explains my logic pretty well.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
kushti: Please see a comment on inter-subjectiveness above

kushti:
@aslesarenko removed "full-blocks", now height comment is "Height of the chain as reported by the state", as older one was not correct for hybrid regimes. Also made log out log.warn(s"Wallet: block in the past reported at ${block.height}, blockId: ${block.id}"). That's all

Issue #991, opened 2019-12-13
Implement API for fee estimation calculation


azhiganov:
In BTC blockchain there is API to get satoshi per byte value, which represents current fee to be needed to get transaction mined. So to undersand which fee should be set, API to get ERG per byte need to be implemented. It allows to calculate fee by multiplying it on size of transaction and to get it mined.

kushti:
Done in 4.0.3

PR #1175, opened 2020-08-19
[refer #991] Add /transactions/poolhist request to obtain mempool fee…

knizhnik:
… histogram


greenhat (changes requested):
Looking good, but:
1. I did not understand what is a "bean". Maybe you meant "bin"?
2. I did not find tests for added features.

knizhnik:
It is not quite clear how to write tests for the functions which behavior depends on system time. Certainly I can put some transactions in mempool, then take them with same delay, but in any case final result will be non determinstic and it is hard to find some reasonable bounds.

kushti:
> It is not quite clear how to write tests for the functions which behavior depends on system time. Certainly I can put some transactions in mempool, then take them with same delay, but in any case final result will be non determinstic and it is hard to find some reasonable bounds.

How's about to change a bit MemPoolStatistics:

def add(wtx : WeightedTxId) : Unit = add(wtx, System.currentTimeMillis())

private[mempool] add add(wtx : WeightedTxId, time: Long) = {
...
}

Then your test (located in package org.ergoplatform.nodeView.mempool ) can work with the private method and you can have precise control over testing data

knizhnik:
Current system time is used not only in add() method, but also in getExpectedWaitTime.
Actually time is not the main problem - I can add one second sleeps and roun d result to integer number of seconds.
Th main problem is that measurement precision (histogram bin and eexpected wait time) are minute. It seems to be too large for regression test.

kushti:
> Current system time is used not only in add() method, but also in getExpectedWaitTime.
> Actually time is not the main problem - I can add one second sleeps and roun d result to integer number of seconds.
> Th main problem is that measurement precision (histogram bin and eexpected wait time) are minute. It seems to be too large for regression test.


Why not to provide fake clocks to getExpectedWaitTime as well ?

knizhnik:
> > Current system time is used not only in add() method, but also in getExpectedWaitTime.
> > Actually time is not the main problem - I can add one second sleeps and roun d result to integer number of seconds.
> > Th main problem is that measurement precision (histogram bin and eexpected wait time) are minute. It seems to be too large for regression test.
> 
> Why not to provide fake clocks to getExpectedWaitTime as well ?

Because it still doesn't help to solve minute resolution problem.

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -68,4 +73,41 @@ case class TransactionsApiRoute(readersHolder: ActorRef, nodeViewActorRef: Actor
...
+    val now = System.currentTimeMillis()
+    val interval = maxWaitTimeMsec / nBins
+    for (wtx <- wtxs) {
+      val waitTime = now - wtx.created
+      val bin = if (waitTime < maxWaitTimeMsec) (waitTime/interval).asInstanceOf[Int] else nBins
+      histogram(bin).nTxns += 1
+      histogram(bin).totalFee += wtx.feePerKb
+    }
+    histogram
+  }
```
aslesarenko, line 89: This method can be (and IMO should be) moved out of this route class (for example into some builder class). The time provider can then be abstracted and the algorithms is unit tested (including providing test vectors for regression, which is necessary for the public API)

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -7,14 +7,61 @@ import org.ergoplatform.nodeView.state.{ErgoState, UtxoState}
...
 import scorex.core.transaction.MemoryPool
 import scorex.core.transaction.state.TransactionValidation
-import scorex.util.ModifierId
+import scorex.util.{ModifierId, bytesToId}
 
 import scala.util.Try
 
+object MemPoolStatisticsParams {
+  val nHistogramBins = 60  /* one hour */
+  val measurementIntervalMsec = 60 * 1000 /* one hour */
```
aslesarenko, line 16: Please add ScalaDoc for these public fields.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -7,14 +7,61 @@ import org.ergoplatform.nodeView.state.{ErgoState, UtxoState}
...
+}
+
+case class FeeHistogramBin(var nTxns: Int, var totalFee: Long)
```
aslesarenko, line 19: please add ScalaDoc

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -7,14 +7,61 @@ import org.ergoplatform.nodeView.state.{ErgoState, UtxoState}
...
+
+case class MemPoolStatistics() {
```
aslesarenko, line 21: Case classes without args is anti-pattern. Why not use `case object` then?
Also please add ScalaDoc.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -7,14 +7,61 @@ import org.ergoplatform.nodeView.state.{ErgoState, UtxoState}
...
+  var startMeasurement: Long = System.currentTimeMillis() // start of measurement interval
+  var takenTxns : Long = 0              // amount of taken transaction since start of measurement
+  var snapTime: Long = startMeasurement // last snapshot time
+  var snapTakenTxns : Long = 0          // amount of transaction at the moment of last snapshot
```
aslesarenko, line 25: Please put these comments into ScalaDocs.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -7,14 +7,61 @@ import org.ergoplatform.nodeView.state.{ErgoState, UtxoState}
...
+  val histogram = Array.fill(MemPoolStatisticsParams.nHistogramBins)(FeeHistogramBin(0,0))
+
+  //                  <.............takenTxns...................
+  //                                        <....snapTakenTxns..
+  // ------------------------------------------------------------> time
+  //                  ^                     ^                  ^
+  //                  |<measurementInterval>|                  |
+  // startMeasurement +            snapTime +      current time+
```
aslesarenko, line 33: This should be put into ScalaDoc using PRE formatting. See ScalaDoc guide at scala-lang.org

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -7,14 +7,61 @@ import org.ergoplatform.nodeView.state.{ErgoState, UtxoState}
...
+
+  /**
+    * Add new entry to mempool statistics. This method is called when transaction is taken from mempool
+    * and placed in blockchain. It is not called when transaction in thrown away from the pool by replacement policy.
+    * To make statistic better represent most recent system behaviour, we periodically (each measurementIntervalMsec)
+    * prune statistic. To avoid situtation when we do not have statistic at all, we actually keep data up to
+    * 2*measurementIntervalMsec and periodically cut half of range.
+    */
+  def add(wtx : WeightedTxId) : Unit = {
+    val now = System.currentTimeMillis()
```
aslesarenko, line 43: The time provider can be introduced to abstract from time source and make this class unit-testable.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -105,6 +152,50 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
...
+      // locate first non-empty histogram bin preceding or equal to the specified wait time
+      for (i <- expectedWaitTimeMinutes to 0 by -1) {
+        val h = stats.histogram(expectedWaitTimeMinutes)
+        if (h.nTxns != 0) {
+          return h.totalFee / h.nTxns * txSize / 1024
+        }
+      }
+    }
+    settings.nodeSettings.minimalFeeAmount
+  }
```
aslesarenko, line 172: Can be moved out to a histogram builder and then unit-tested independently of this mempool.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -105,6 +152,50 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
...
+    val posInPool = pool.orderedTransactions.keySet.until(wtx).size
+
+    // Time since statistics measurement interval (needed to calculate average tx rate)
+    val elapsed = System.currentTimeMillis() - stats.startMeasurement
+    if (stats.takenTxns != 0) {
+      elapsed * posInPool / stats.takenTxns
+    } else {
+      0
+    }
+  }
```
aslesarenko, line 198: Can be moved out to a histogram builder and then unit-tested independently of this mempool, in particular using mock time provider.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -7,14 +7,61 @@ import org.ergoplatform.nodeView.state.{ErgoState, UtxoState}
...
+      histogram(durationMinutes).nTxns += 1
+      histogram(durationMinutes).totalFee += wtx.feePerKb
+    }
+  }
+}
 /**
   * Immutable memory pool implementation.
   */
-class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoSettings)
+class ErgoMemPool private[mempool](pool: OrderedTxPool, stats : MemPoolStatistics)(implicit settings: ErgoSettings)
```
aslesarenko, line 64: Please augment ScalaDoc with missing description of constructor parameters.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -45,19 +92,19 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
...
-    new ErgoMemPool(pool.filter(condition))
+    new ErgoMemPool(pool.filter(condition),stats)
   }
 
   def invalidate(tx: ErgoTransaction): ErgoMemPool = {
-    new ErgoMemPool(pool.invalidate(tx))
+    new ErgoMemPool(pool.invalidate(tx),stats)
   }
 
   def process(tx: ErgoTransaction, state: ErgoState[_]): (ErgoMemPool, ProcessingOutcome) = {
```
aslesarenko, line 110: Please add ScalaDoc for this public method.

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -62,9 +62,11 @@ case class OrderedTxPool(orderedTransactions: TreeMap[WeightedTxId, ErgoTransact
...
   }
 
-  def remove(tx: ErgoTransaction): OrderedTxPool = {
-    transactionsRegistry.get(tx.id).fold(this)(wtx =>
-      OrderedTxPool(orderedTransactions - wtx, transactionsRegistry - tx.id, invalidated, outputs -- tx.outputs.map(_.id)).updateFamily(tx, -wtx.weight))
+  def remove(tx: ErgoTransaction, callback: WeightedTxId => Unit = (wtx:WeightedTxId) => ()): OrderedTxPool = {
+    transactionsRegistry.get(tx.id).fold(this)(wtx => {
+      callback(wtx)
+      OrderedTxPool(orderedTransactions - wtx, transactionsRegistry - tx.id, invalidated, outputs -- tx.outputs.map(_.id)).updateFamily(tx, -wtx.weight)
+    })
```
aslesarenko, line 69: Public method, please add ScalaDoc.

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -138,7 +140,7 @@ case class OrderedTxPool(orderedTransactions: TreeMap[WeightedTxId, ErgoTransact
 
 object OrderedTxPool {
 
-  case class WeightedTxId(id: ModifierId, weight: Long) {
+  case class WeightedTxId(id: ModifierId, weight: Long, feePerKb: Long, created: Long) {
```
aslesarenko, line 143: Please add ScalaDoc

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -162,7 +164,8 @@ object OrderedTxPool {
       .map(_.value)
       .sum
     // We multiply by 1024 for better precision
-    WeightedTxId(tx.id, fee * 1024 / tx.size)
+    val feePerKb = fee * 1024 / tx.size
+    WeightedTxId(tx.id, feePerKb, feePerKb, System.currentTimeMillis())
```
aslesarenko, line 168: The time source can be abstacted using time provider, then the `OrderedTxPool.put` method can be unit-tested (where this method is used).

aslesarenko (changes requested):

PR #1193, opened 2020-09-17
Close #1189: memory leak in ErgoProvingInterpreter

### LINKED ISSUE #1189: Memory leak in ErgoWalletActor
Originally posted here https://github.com/ScorexFoundation/sigmastate-interpreter/issues/687

The root cause is that the same ErgoProvingInterpreter stored in WalletVars and reused during the whole lifecycle of the actor. It should be recycled, from time to time.

Alternatively, the IRContext stored in the prover (and which accumulates all compiled expressions) should be recycled, for example a fresh IRContext can be used for each new block.

IRContext should never be part of the global state, it should instead always have temporary lifetime (like a local variable)

kushti:
In this PR, IR context in the prover being reset on each signing operation


src/main/scala/org/ergoplatform/nodeView/wallet/WalletScanLogic.scala
```diff
@@ -115,10 +115,16 @@ object WalletScanLogic extends ScorexLogging {
...
 
     // Resolve boxes related to mining income
-    val resolvedBoxes = registry.unspentBoxes(MiningScanId).flatMap { tb =>
-      val spendable = resolve(tb.box, walletVars.proverOpt, stateContext, height)
-      if (spendable) Some(tb.copy(scans = Set(PaymentsScanId))) else None
-    }
+    // We take only 5 oldest boxes to avoid long scans in the testnet where one wallet
+    // (corresponding to a node generating almost all the blocks) can possess
+    // hundreds of unspent mining reward boxes.
+    val miningBoxes = registry.unspentBoxes(MiningScanId).sortBy(_.inclusionHeightOpt.getOrElse(0)).take(5)
```
aslesarenko, line 121: Please clarify in the code what it means to take `5 oldest` and why it is safe (make sence). 
Note, before the `take` method comes into play, all set of boxes is returned (as a Seq with underlying ListBuffer probably) and then is it sorted.

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/nodeView/wallet/WalletScanLogic.scala
kushti: As node settings available here, I made a better approach, choosing boxes which are mature enough specifically.

kushti:
@aslesarenko comments addressed, please make another pass!

aslesarenko (approved):

Issue #959, opened 2019-11-01
Wallet: generate keys according to EIP3


catena2w:
[EIP3](https://github.com/ergoplatform/eips/blob/master/eip-0003.md) is accepted, now it's time to use it inside a build-in node wallet.
To keep backward compatibility, I propose to add a new flag to config `wallet.legacyMode=false`. It is false by default and generate private keys according to EIP3. If it is set to true - it generates private keys according to old style.

aslesarenko:
@catena2w I suggest to provide migration from old wallets to the new ones if possible. Thus simplify the reference implementation code.

catena2w:
We should keep in mind, that there are external applications that already relies on this code:
- GPU miner generate keys according to the old scheme.
- Exchanges generate batches of addresses for their users. I don't think that it is possible to requeist from all the exchanges to change deposit addresses for all their users.

kushti:
@catena2w it is unlikely that users will change default values in the settings. 

More, there's the need to support primitive secrets (just dlogs or dlog-equality secret keys), see e.g. https://github.com/ergoplatform/ergo/issues/940

Thus I propose to think how can different schemes co-exist.

catena2w:
Well, that was my idea - new users won't change the default values, and thus will create wallets according to EIP3. 
Existing users will migrate from the current version using a guide and thus will change their config. If their addresses will change after updating to some release, it should not too difficult yo determine how they should change the config to keep their old wallet working. So with time old-style wallets will be replaced with the new one, and all users will have EIP3 keys

kushti:
The problem is that users do not read any guides when updating, also not updating from one version to next one, so there will be need to put this guide in bold for many releases ahead. 

Just personal observations from supporting tens of users during last months.

kushti:
Done  in #1199, @aslesarenko @catena2w please check how

Issue #1082, opened 2020-04-05
Make test for nextPath() in DerivationPath


kushti:
nextPath() method in DerivationPath finds next available path index in BiP-32 implementation. Unfortunately, it is not covered by unit tests. Make nextPath() logic description as well.

PR #1199, opened 2020-09-22
EIP-3 (BIP-44) implementation 

kushti:
This PR is about EIP-3 implementation ( see issue #959 ). For the sake of compatibility with the wallets made with previous version sand miner, following rules apply, based on an observation that users import old wallets via copying files (/wallet folder in case of 3.3.x, /wallet/keystore and /wallet/storage in case of 3.2.x) or /wallet/restore. Please also note that older versions expose public key corresponding to the master key, while new version hides it (while maintaining the master key and using it in the scans) when it is possible.

Thus derivation works as follows:
* if no keys derived yet, derive it against m/44'/429'/0'/0/0 by default, or against m/1 if ergo.wallet.preEip3Derivation = true (false by default)
* if there're keys in the wallet, derive by using old scheme (m/1 -> m/2, m/44'/429'/0'/0/0 -> m/44'/429'/0'/1) if last key is not in the EIP-3 space (not starting with m/44'/429'/0'), otherwise, derive via new procedure (m/44'/429'/0'/0/0 -> m/44'/429'/0'/0/1, note that the result if different from the old scheme).

Thus in case of import from files everything should be ok, in case of restoring a wallet from mnemonic, funds will be found automatically if there were no derived keys there, otherwise, preEip3Derivation flag is needed to be set. 

Also:  
* hide a public key corresponding to the master key if derivation path of the first derived key starts with m/44'/429'/0' in /wallet/addresses output , the key is participating in scans still
* change address by default is set to m/44'/429'/0'/0/0 
* send public key corresponding to master key to the miners , so miningScript (/mining/rewardsAddress) is the same as before.

Close: #959 , #1082


ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/DerivationPath.scala
```diff
@@ -41,6 +50,13 @@ final case class DerivationPath(decodedPath: Seq[Int], publicBranch: Boolean) {
     */
   def toPrivateBranch: DerivationPath = this.copy(publicBranch = false)
 
+  /**
+    * @return whether derivation path corresponds to EIP-3
+    */
+  def isEip3: Boolean = {
```
greenhat, line 56: Should we consider to encode pre-EIP3 vs. EIP3 paths on a type level (ADT, etc.)?

greenhat (changes requested):
Looks good, please, see the comments.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/Constants.scala
```diff
@@ -25,4 +26,28 @@ object Constants {
...
+    * Following this idea we should use next scheme
+    *
+    * m / 44' / 429' / account' / change / address_index
+    */
+  val CoinType = 429
+
+  /**
+    * Pre - EIP3 derivation path
+    */
+  val oldDerivation: DerivationPath = DerivationPath.fromEncoded("m/1").get
```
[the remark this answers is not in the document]
kushti: done

src/main/scala/org/ergoplatform/settings/WalletSettings.scala
```diff
@@ -6,6 +6,7 @@ import org.ergoplatform.wallet.settings.SecretStorageSettings
 case class WalletSettings(secretStorage: SecretStorageSettings,
                           seedStrengthBits: Int,
                           mnemonicPhraseLanguage: String,
+                          oldDerivation: Boolean = false,
```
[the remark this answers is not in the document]
kushti: done (but preEip3Derivation , not *Path)

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/DerivationPath.scala
```diff
@@ -71,8 +87,11 @@ object DerivationPath {
 
   /**
     * Finds next available path index for a new key.
+    * @param secrets - secrets previously generated
+    * @param oldDerivation - whether to use pre-EIP3 derivation or not
     */
-  def nextPath(secrets: IndexedSeq[ExtendedSecretKey]): Try[DerivationPath] = {
+  def nextPath(secrets: IndexedSeq[ExtendedSecretKey],
+               oldDerivation: Boolean = false): Try[DerivationPath] = {
```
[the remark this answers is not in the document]
kushti: done (but preEip3Derivation , not *Path)

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/DerivationPath.scala
kushti: I dunno, maybe not so critical difference to disinguish on the type level?

kushti:
@greenhat addressed comments, please make another pass!

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/DerivationPath.scala
greenhat: Yeah, I’m not completely sure here. Seems useful to further distinguish code paths that handle old and new derivation. Do they intended to co-exist forever or is it more like a “migration” kind of scenario?

greenhat (approved):

ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/DerivationPathSpec.scala
```diff
@@ -19,4 +22,105 @@ class DerivationPathSpec
...
+    implicit val enc = new ErgoAddressEncoder(ErgoAddressEncoder.MainnetNetworkPrefix)
+
+    // This testing pair is checked against CoinBarn and Yoroi
+    val mnemonic = "liar exercise solve delay betray sheriff method empower disease river recall vacuum"
+    val address = "9hAymcGaRfTX7bMADNdfWfk7CKzi2ZpvRBCmtEf6d92n8E26Ax7"
+
+    val seed = Mnemonic.toSeed(mnemonic, None)
+
+    val masterKey = ExtendedSecretKey.deriveMasterKey(seed)
+    val dp = DerivationPath.nextPath(IndexedSeq(masterKey)).get
```
aslesarenko, line 35:
```suggestion
    val dp = DerivationPath.nextPath(IndexedSeq(masterKey), usePreEip3Derivation = false).get
```

aslesarenko: Suggest to pass this explicitly for readability

ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/DerivationPathSpec.scala
```diff
@@ -19,4 +22,105 @@ class DerivationPathSpec
...
+
+    val seed = Mnemonic.toSeed(mnemonic, None)
+
+    val masterKey = ExtendedSecretKey.deriveMasterKey(seed)
+    val dp = DerivationPath.nextPath(IndexedSeq(masterKey), usePreEip3Derivation = true).get
+    val sk = masterKey.derive(dp).asInstanceOf[ExtendedSecretKey]
+    val pk = sk.publicKey.key
+    P2PKAddress(pk).toString() shouldBe address
+
+    val dp2 = DerivationPath.nextPath(IndexedSeq(masterKey, sk)).get
```
aslesarenko, line 76: So, the usePreEip3Derivation = false is ignored in this case?

aslesarenko: Please describe the behavior for this test case in ScalaDoc of the method.

src/main/scala/org/ergoplatform/nodeView/wallet/WalletCache.scala
```diff
@@ -67,20 +67,29 @@ object WalletCache {
...
 
-  def trackedAddresses(trackedPubKeys: Seq[ExtendedPublicKey],
-                       addressEncoder: ErgoAddressEncoder): Seq[P2PKAddress] =
-    trackedPubKeys.map(pk => P2PKAddress(pk.key)(addressEncoder))
+  private def publicKeyAddresses(trackedPubKeys: Seq[ExtendedPublicKey],
+                                 addressEncoder: ErgoAddressEncoder): Seq[P2PKAddress] = {
+    // remove master key from visible addresses if following-up keys are from EIP-3 range
+    val pks = if (trackedPubKeys.size > 1
+      && trackedPubKeys.head.path.isMaster
+      && trackedPubKeys(1).path.isEip3) {
```
aslesarenko, line 78: Suggest to extract this condition into a named predicate.

src/main/scala/org/ergoplatform/nodeView/wallet/WalletCache.scala
```diff
@@ -67,20 +67,29 @@ object WalletCache {
...
+      trackedPubKeys.tail
```
aslesarenko, line 79: Please add a clarifying comment, why `tail` is used here.

aslesarenko (changes requested):

ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/DerivationPathSpec.scala
kushti: done

ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/DerivationPathSpec.scala
kushti: @aslesarenko are we going to introduce ScalaDoc requirements for testing code ?

src/main/scala/org/ergoplatform/nodeView/wallet/WalletCache.scala
kushti: But why? As it is used once in a small method (these lines take like a half of the method), then readability will not be improved I guess.

src/main/scala/org/ergoplatform/nodeView/wallet/WalletCache.scala
kushti: Was done on line 75, add another comment though

kushti:
@aslesarenko comments addressed, please make another pass

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/DerivationPath.scala
kushti: @greenhat As there's (and likely) will be a way to derive from a custom path, the schemes will coexist. 

We do not follow BIP-44 fully, namely, we do not separate between internal and external addresses, maybe some support will come in future but hard to tell atm

ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/DerivationPathSpec.scala
aslesarenko: @kushti no, I just don't understand the code. 
As always, please don't spend time writing he response here, improve the code instead, if you will.

src/main/scala/org/ergoplatform/nodeView/wallet/WalletCache.scala
aslesarenko: What this predicate means? Looks like some intermediate state of the WalletCache. If yes, why don't give it a name?

kushti: @aslesarenko there's the comment (line 75) explaining what is checked in the code

ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/DerivationPathSpec.scala
kushti: @aslesarenko ScalaDoc is not needed then it seems, added comments.

kushti:
@aslesarenko comments has been addressed, please make another pass!

aslesarenko (approved):

Issue #1190, opened 2020-09-17
Optimize HeadersProcessor.headerChainBack method


aslesarenko:
Here is a list of hotspot methods of Ergo node syncing (around 150000 block, Headers syncronization stage)
<img width="451" alt="image" src="https://user-images.githubusercontent.com/654077/93467341-4f0b3800-f8f6-11ea-858c-feb85230229b.png">

Most of the time is spent in bouncycastle. We cannot optimize those methods themselves, we can only try to optimize the number of their invocations, if it is possible at all.

However, there is one low hanging fruit (highlighted above). The method is called here
```scala
def headerChainBack(limit: Int, startHeader: Header, until: Header => Boolean): HeaderChain = {
  @tailrec
  def loop(header: Header, acc: Seq[Header]): Seq[Header] = {
    if (acc.lengthCompare(limit) == 0 || until(header)) {
      acc
    } else {
      typedModifierById[Header](header.parentId) match {
        case Some(parent: Header) =>
          loop(parent, acc :+ parent)  // <--- this $colon$plus 
        case None if acc.contains(header) =>
          acc
        case _ =>
          acc :+ header
      }
    }
  }
```

The problem with this method is that it uses immutable Seq, which leads to reallocation and copying of acc content on each step (so this is O(n^2) number of operations with headers)
<img width="88" alt="image" src="https://user-images.githubusercontent.com/654077/93467624-97c2f100-f8f6-11ea-9fac-87cdfb5b270b.png">

As can be seen, it has an internal loop which consumes all its time.
 
The headerChainBack  method can be rewritten using ArrayBuffer or any other dynamically growable arrays (ArrayList from Java also ok).

After this optimization is done, then `acc.contains(header)` may show up as the hotspot in the profile. So it make sense to optimize both operations.

### Memory allocations
Here is the hotspot methods by object counts
<img width="1111" alt="image" src="https://user-images.githubusercontent.com/654077/93468313-71ea1c00-f8f7-11ea-959d-bbec5fdc38be.png">

The top 2 positions are attributed back to the same method
<img width="1297" alt="image" src="https://user-images.githubusercontent.com/654077/93468166-3f402380-f8f7-11ea-8496-d54535cb35f2.png">

kushti:
Duplication of https://github.com/ergoplatform/ergo/pull/974

aslesarenko:
Let's fix the problem first and then close the issue.

aslesarenko:
fixed in #1198

PR #1202, opened 2020-09-25
Replace Seq with ArrayBuffer in headerChainBack

knizhnik:


kushti (approved):

kushti:
@aslesarenko is this PR fully resolving #1190 ?

Issue #1067, opened 2020-03-11
Crash on OpenJDK11 and Oracle JDK 8 on macOS


greenhat:
Branch: develop

Steps to reproduce: run `sbt test`

Log for OpenJDK11:
```
[info] - applyModifier() - invalid block
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0xa) at pc=0x00000001116703f0, pid=57166, tid=38147
#
# JRE version: OpenJDK Runtime Environment (11.0.6+10) (build 11.0.6+10)
# Java VM: OpenJDK 64-Bit Server VM (11.0.6+10, mixed mode, tiered, compressed oops, g1 gc, bsd-amd64)
# Problematic frame:
# C  [libleveldbjni-64-1-5503110146537748033.8+0x373f0]  typeinfo for leveldb::DB+0x0
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/green/code/WORK/ergo/hs_err_pid57166.log
#
# If you would like to submit a bug report, please visit:
#   https://github.com/AdoptOpenJDK/openjdk-support/issues
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
Exception in thread "Thread-11" java.io.EOFException
	at java.base/java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:3117)
	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1597)
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:464)
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:422)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1$React.react(Framework.scala:818)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1.run(Framework.scala:807)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

Log for OracleJDK8: 
```
[info] - applyModifier() - invalid block
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0xa) at pc=0x00000001303c33f0, pid=65014, tid=0x0000000000005903
#
[...]

PR #1207, opened 2020-10-02
Fix for #1067 : crash on OpenJDK11 and Oracle JDK 8 on macOS

kushti:
The node is crashing on MacOS due to problems LevelDB-JNI.

This PR is about to fix it by using pure-Java LevelDB on MacOS.

### DIFF
--- a/avldb/src/main/scala/scorex/db/LDBFactory.scala
+++ b/avldb/src/main/scala/scorex/db/LDBFactory.scala
@@ -2,8 +2,7 @@ package scorex.db
 
 import java.io.File
 import java.util.concurrent.locks.ReentrantReadWriteLock
-
-import org.iq80.leveldb.{DB, Range, DBFactory, DBIterator, Options, ReadOptions, Snapshot, WriteBatch, WriteOptions}
+import org.iq80.leveldb.{DB, DBFactory, DBIterator, Options, Range, ReadOptions, Snapshot, WriteBatch, WriteOptions}
 import scorex.util.ScorexLogging
 
 import scala.collection.mutable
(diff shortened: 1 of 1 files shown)


knizhnik (approved):

greenhat:
I've been running the node built from this PR (syncing) for 3+ hours without crashes on macOS Catalina. Seems like it fixes the issue.

avldb/src/main/scala/scorex/db/LDBFactory.scala
```diff
@@ -138,15 +137,25 @@ object LDBFactory extends ScorexLogging {
 
   lazy val factory: DBFactory = {
     val loaders = List(ClassLoader.getSystemClassLoader, this.getClass.getClassLoader)
-    val factories = List(nativeFactory, javaFactory)
+
+    // As LevelDB-JNI has problems on Mac (see https://github.com/ergoplatform/ergo/issues/1067),
+    // we are using only pure-Java LevelDB on Mac
+    val isMac = System.getProperty("os.name").toLowerCase().indexOf("mac") >= 0
```
aslesarenko, line 143: I suggest to google for a more principled way of detecting macOS and leave the reference in the code.

aslesarenko (approved):
LGTM, but please see my comment also.

avldb/src/main/scala/scorex/db/LDBFactory.scala
kushti: @aslesarenko there's ready check in Apache common-utils, but this library is not available in the avldb project, and no need to get a dependency for 1 line I guess. I checked an implementation there also, it works like the line of code of mine basically.

Issue #1165, opened 2020-08-11
Do not allow for multiple transactions in the mempool spending the same box


kushti:

jasondavies:
Why shouldn't this be allowed?

kushti:
@jasondavies I think it is about easier DDoS possibility in the first place? For example, if I have 10 outputs, I can create and send to the network 10!=3.6M spending transactions by reordering the inputs, with double spending prevention just 10. 

Other arguments would be more nuanced likely )

aslesarenko:
@kushti I think it would be useful to extend the behavior of transaction send method to somehow communicate to the application, that sending was rejected due to double spending.
Such kind of feedback will be useful for multistage contracts with alternative spending paths.
If I send a transaction using PathA, I may receive rejection from the pool that there is already a transaction spending the same box by following PathB. 

Simple RejectionReason.DoubleSpending will be enough for the application.

kushti:
> @kushti I think it would be useful to extend the behavior of transaction send method to somehow communicate to the application, that sending was rejected due to double spending.
> Such kind of feedback will be useful for multistage contracts with alternative spending paths.
> If I send a transaction using PathA, I may receive rejection from the pool that there is already a transaction spending the same box by following PathB.
> 
> Simple RejectionReason.DoubleSpending will be enough for the application.

You mean HTTP API? Transaction is being sent to the mempool asynchronously there, after validation, so mempool response is not known even know,

aslesarenko:
Yeah, I mean HTTP API. Ok, if it not easy to do now, maybe just keep in mind for the future.

scalahub:
Would this allow us to double spend with higher fee (RBF)? This is needed as a tx is often stuck due to low fee

kushti:
> Would this allow us to double spend with higher fee (RBF)? This is needed as a tx is often stuck due to low fee

Yes, RBF is done

PR #1213, opened 2020-10-10
Fix for #1165: double-spending transactions not allowed in mempool (with RBF replacement strategy)

kushti:
In this PR, a double-spending spending transaction is being rejected from the mempool if it is not paying more than all the conflicting transactions (along with dependent transactions) already sitting in the pool altogether.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
@@ -14,7 +14,7 @@ import scala.util.Try
 /**
   * Immutable memory pool implementation.
   */
-class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoSettings)
+class ErgoMemPool private[mempool](val pool: OrderedTxPool)(implicit settings: ErgoSettings)
   extends MemoryPool[ErgoTransaction, ErgoMemPool] with ErgoMemPoolReader {
 
   import ErgoMemPool._
--- a/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
@@ -68,9 +68,11 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
     }
     doubleSpendingInputOpt match {
       case Some(doubleSpendingInput) =>
+        val ownWtx = weighted(tx)
         val doubleWtx = pool.inputs.get(doubleSpendingInput.boxId).get //.get
-        if (weighted(tx).weight > doubleWtx.weight) {
-          new ErgoMemPool(pool.put(tx)) -> ProcessingOutcome.Accepted
+        if (ownWtx.weight > doubleWtx.weight) {
+          val doubleTx = pool.orderedTransactions.get(doubleWtx).get //.get
+          new ErgoMemPool(pool.put(tx).remove(doubleTx)) -> ProcessingOutcome.Accepted
         } else {
           this -> ProcessingOutcome.DoubleSpendingLoser(doubleWtx.id)
         }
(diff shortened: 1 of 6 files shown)


src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -14,12 +14,14 @@ import scala.util.Try
 /**
   * Immutable memory pool implementation.
   */
-class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoSettings)
+class ErgoMemPool private[mempool](val pool: OrderedTxPool)(implicit settings: ErgoSettings)
```
oskin1, line 17: `pool: OrderedTxPool` should not be exposed in ErgoMemPool API. Why have you made it public?

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -60,6 +62,28 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool)(implicit settings: ErgoS
     new ErgoMemPool(pool.invalidate(tx))
   }
 
+  // Check if transaction is double-spending inputs spent in the mempool.
+  // If so, the new transacting is replacing older ones if it is paying more than all of them.
+  // Otherwise, the new transaction being rejected.
+  private def acceptIfNoDoubleSpend(tx: ErgoTransaction): (ErgoMemPool, ProcessingOutcome) = pool.synchronized {
```
aslesarenko, line 68: This `synchronized` is suspicious. Why it is necessary? Considering pool is immutable, and this method creates a new instance of it, it looks like the syncronization should lock on other object. Probably on a higher level of ErgoNodeViewHolder.txModify method.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -114,6 +134,8 @@ object ErgoMemPool {
 
     case object Accepted extends ProcessingOutcome
 
+    case class DoubleSpendingLoser(winnerTxIds: Set[ModifierId]) extends ProcessingOutcome
```
aslesarenko, line 137: Please add ScalaDoc for all public classes.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -127,4 +149,17 @@ object ErgoMemPool {
   def empty(settings: ErgoSettings): ErgoMemPool =
     new ErgoMemPool(OrderedTxPool.empty(settings))(settings)
```
aslesarenko, line 150: ScalaDoc is missing

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -152,17 +182,12 @@ object OrderedTxPool {
...
-      .map(_.value)
-      .sum
-    // We multiply by 1024 for better precision
-    WeightedTxId(tx.id, fee * 1024 / tx.size)
+    OrderedTxPool(
+      TreeMap.empty[WeightedTxId, ErgoTransaction],
+      TreeMap.empty[ModifierId, WeightedTxId],
+      TreeMap.empty[ModifierId, Long],
+      TreeMap.empty[BoxId, WeightedTxId],
+      TreeMap.empty[BoxId, WeightedTxId])(settings)
```
aslesarenko, line 190: Since these TreeMaps is immutable, to avoid garbage, empty instances can be allocated once and saved as `val`, then shared by all new OrderedTxPool created here.

aslesarenko (changes requested):

greenhat (commented):
LGTM, besides existing comments.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: removed

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: done

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
kushti: Empty pool is created only once, so not an issue likely.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: done

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: Good point, exposed it to made checks in tests, but now just exposing needed data (spentInputs, will be used later in the wallet likely). Reverted

aslesarenko (approved):

PR #1224, opened 2020-10-22
Dropping inactive connections, last seen message time added to /info 

kushti:
This PR is introducing following changes:

* Scorex version with inactive connections dropping being used
* Timestamp of a last seen message (from any peer) added to /info
* 118.189.191.191:9030 removed from seeds list


src/main/resources/mainnet.conf
```diff
@@ -28,15 +28,14 @@ scorex {
...
-      "118.189.191.191:9030",
       "51.75.52.63:9030",
       "159.65.11.55:9030",
       "165.227.26.175:9030",
       "78.46.93.239:9030",
       "159.89.116.15:9030",
       "83.212.114.255:9030"
     ]
-    maxDeliveryChecks = 40
+    maxDeliveryChecks = 4
```
aslesarenko, line 38: I'm a little worry about reducing network parameters. Do we have a risk of loosing connectivity because of this? 
Ideally simulation is required to find out the minimal working values after which we start to observe problems, and then based on those values we can setup safe parameter values.
So if this change is necessary, I suggest to make small step, reduce to 20 and see what happens.

src/main/scala/org/ergoplatform/local/ErgoStatsCollector.scala
```diff
@@ -51,12 +56,20 @@ class ErgoStatsCollector(readersHolder: ActorRef,
     None,
     None,
     None,
-    timeProvider.time(),
+    networkTime(),
+    networkTime(),
```
aslesarenko, line 60: looks suspicious, let's pass arguments by name.

aslesarenko (changes requested):

src/main/resources/mainnet.conf
kushti: Good point, set to 20. On simulations, would be nice to have them, but also very tricky.

src/main/scala/org/ergoplatform/local/ErgoStatsCollector.scala
kushti: done

kushti: @aslesarenko comments addressed, please make another pass!

aslesarenko (approved):

greenhat (approved):

Issue #1226, opened 2020-10-22
Sync protocol 2.0 


kushti:
Currently syncing algorithm is not efficient at all (taken from PoS protocols we developed before): a peer is sending last 1,000 block ids to a random peer, and the neighbour is answering with an suffix of a block it has (latest block usually, but not necessarily). 

However, with PoW we can do better, significantly reducing size of sync message, caching reported heights of neighbours etc. Then parallel headers and blocks downloading ( #1209 ) can be done.

PR #1231, opened 2020-10-30
Peers discovery fix

kushti:
There is a bug introduced in 3.3.4 (caused by this refactoring in Scorex: https://github.com/ScorexFoundation/Scorex/pull/364 ), resulting in peer discovery being broken in 3.3.4-5. This PR is introducing a fix for that.


src/main/scala/org/ergoplatform/ErgoApp.scala
```diff
@@ -107,6 +107,9 @@ class ErgoApp(args: Args) extends ScorexLogging {
     ErgoNodeViewSynchronizer(networkControllerRef, nodeViewHolderRef, ErgoSyncInfoMessageSpec,
       ergoSettings, timeProvider)
 
+  val peerSynchronizer: ActorRef = PeerSynchronizerRef("PeerSynchronizer",
```
greenhat, line 110: I'm not sure I understand what's going on here. `peerSynchronizer` is not used anywhere. Is it just "started" here? Why is it not private?

src/main/resources/mainnet.conf
```diff
@@ -33,9 +33,15 @@ scorex {
...
+      "83.212.114.255:9030",
+      "136.244.110.145:9030",
+      "91.199.118.161:9030",
+      "104.186.109.55:9030",
+      "209.217.206.254:9030",
+      "145.239.4.27:9030",
+      "94.130.108.35:9030"
     ]
-    maxDeliveryChecks = 20
+    maxDeliveryChecks = 18
```
aslesarenko, line 44: Please add comment in the code on what is the rationale behind this value.

aslesarenko (changes requested):
@kushti I looked at the Scorex PR, and the changes here and don't see a relation. 
Please describe, what was the problem, why it appears after the Scorex PR, and what is the solution.

Also, the original Scorex PR seems to be a little problematic. It introduces quite substantial (to my eyes) changes, yet lacking new tests cases (only existing were refactored).

src/main/scala/org/ergoplatform/ErgoApp.scala
kushti: Made it private, also added the comment:
"Launching PeerSynchronizer actor which is then registering itself at network controller"

kushti:
> @kushti I looked at the Scorex PR, and the changes here and don't see a relation.
> Please describe, what was the problem, why it appears after the Scorex PR, and what is the solution.
> 
> Also, the original Scorex PR seems to be a little problematic. It introduces quite substantial (to my eyes) changes, yet lacking new tests cases (only existing were refactored).



> @kushti I looked at the Scorex PR, and the changes here and don't see a relation.
> Please describe, what was the problem, why it appears after the Scorex PR, and what is the solution.
> 
> Also, the original Scorex PR seems to be a little problematic. It introduces quite substantial (to my eyes) changes, yet lacking new tests cases (only existing were refactored).

My bad, the problematic PR moving peer synchronizer creation into an app is this one: https://github.com/ScorexFoundation/Scorex/pull/364 .

src/main/resources/mainnet.conf
kushti: Done in basic file (reference.conf). Copied to mainnet.conf also

kushti:
@greenhat @aslesarenko comments addressed, also, Scorex dependency has been updated, please make another pass

greenhat (approved):

src/main/resources/mainnet.conf
```diff
@@ -33,9 +33,17 @@ scorex {
...
+      "91.199.118.161:9030",
+      "104.186.109.55:9030",
+      "209.217.206.254:9030",
+      "94.130.108.35:9030"
     ]
-    maxDeliveryChecks = 20
+
+    # Max number of delivery checks. Stop expecting modifier (and penalize peer) if it was not delivered after that
+    # number of delivery attempts
+    maxDeliveryChecks = 4
```
aslesarenko, line 46: What is the reason to reduce this, is it network throughput problem?
During profiling I observed that network is mostly idle even during sync, as well as the node is 50% idle.
So I don't think we should reduce connectivity parameters, before we observe the actual overhead.
Or, if you have some measurments on this, please share.

aslesarenko (approved):

src/main/resources/mainnet.conf
kushti: Syncing performance should be better with maxDeliveryChecks being reduced, as you're waiting less in case of non-delivering peers. 
50% idle time will be fixed with sync protocol 2.0 ( #1226 )

PR #1232, opened 2020-10-30
Close #1229: Wallet Balance Updated on /scan/addBox and /scan/stopTracking

### LINKED ISSUE #1229: /scan/addBox is not updating wallet balance (when scanId = 10)

kushti:
This PR is introducing following changes:

* non-versioned data storage removed, only versioned left
* manually added and removed boxes now appended to the latest version (so may disappear after a rollback(
* wallet balances updated on manually added and removed boxes


greenhat (commented):
Does existing tests cover new/updated code? 
Do we have test coverage tracking(coveralls) setup for this repo?

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
```diff
@@ -18,7 +20,44 @@ import scorex.util.Extensions._
   */
 final case class WalletDigest(height: Int,
                               walletBalance: Long,
-                              walletAssetBalances: mutable.LinkedHashMap[EncodedTokenId, Long])
+                              walletAssetBalances: mutable.LinkedHashMap[EncodedTokenId, Long]) {
```
aslesarenko, line 23: let's make this mutable map private, otherwise it is error prone.
Or even better to replace it with immutable data structure.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
```diff
@@ -18,7 +20,44 @@ import scorex.util.Extensions._
...
+    * @param box - box to include into the digest
+    * @return updated digest
+    */
+  def putBox(box: ErgoBox): WalletDigest = {
+    val mutMap = walletAssetBalances
+
+    box.additionalTokens.toArray.foreach { tokenRec =>
+      val key = EncodedTokenId @@ scorex.util.encode.Base16.encode(tokenRec._1)
+      val value = mutMap.getOrElse(key, 0L) + tokenRec._2
+      mutMap.updated(key, value)
```
aslesarenko, line 37: Here we mutate the walletAssetBalances instance in the original WalletDigest instance.
Either walletAssetBalances should be cloned first or putBox should not create new instance.
I think it is better to make walletAssetBalances immutable in a first place.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
```diff
@@ -18,7 +20,44 @@ import scorex.util.Extensions._
...
+    * @param box - box to exclude from the digest
+    * @return updated digest
+    */
+  def removeBox(box: ErgoBox): WalletDigest = {
+    val mutMap = walletAssetBalances
+
+    box.additionalTokens.toArray.foreach { tokenRec =>
+      val key = EncodedTokenId @@ scorex.util.encode.Base16.encode(tokenRec._1)
+      val value = mutMap.getOrElse(key, 0L) - tokenRec._2
+      mutMap.updated(key, value)
```
aslesarenko, line 54: Same problem as above here.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
```diff
@@ -273,22 +269,45 @@ class WalletRegistry(store: HybridLDBKVStore)(ws: WalletSettings) extends Scorex
   /**
     * Updates scans of a box stored in the wallet database,
     * puts the box into the database if it is not there
+    * removes the box from the database if its there and scanIds are empty
     *
-    * @param scanIds
+    * @param newScans
     * @param box
     * @return
```
aslesarenko, line 276: Please add additional details about params and return or cleanup the declarations.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
```diff
@@ -273,22 +269,45 @@ class WalletRegistry(store: HybridLDBKVStore)(ws: WalletSettings) extends Scorex
...
     */
-  def updateScans(scanIds: Set[ScanId], box: ErgoBox): Try[Unit] = Try {
+  def updateScans(newScans: Set[ScanId], box: ErgoBox): Try[Unit] = Try {
     val bag0 = KeyValuePairsBag(toInsert = Seq.empty, toRemove = Seq.empty)
```
aslesarenko, line 279: Can we call it emptyBag and even move out of the method (if it is immutable) to save allocations and also clarify logic.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
```diff
@@ -273,22 +269,45 @@ class WalletRegistry(store: HybridLDBKVStore)(ws: WalletSettings) extends Scorex
...
+        putBox(bag0, newBox)
+      case (true, true) =>
+        throw new Exception("Can't remove a box which does not exist")
     }
-    val bag2 = putBox(bag1, updTb)
-    store.nonVersionedRemove(bag2.toRemove)
-    store.nonVersionedPut(bag2.toInsert)
+
+    val bag2 = if ((oldScans.contains(Constants.PaymentsScanId) || newScans.contains(Constants.PaymentsScanId)) &&
+      !(oldScans.contains(Constants.PaymentsScanId) && newScans.contains(Constants.PaymentsScanId))) {
```
aslesarenko, line 297: Please leave a clarifying comment on the meaning of the condition.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
```diff
@@ -273,22 +269,45 @@ class WalletRegistry(store: HybridLDBKVStore)(ws: WalletSettings) extends Scorex
...
+      val digest = fetchDigest()
+
+      val updDigest = if (!oldScans.contains(Constants.PaymentsScanId) && newScans.contains(Constants.PaymentsScanId)) {
+        digest.putBox(box)
+      } else if (oldScans.contains(Constants.PaymentsScanId) && !newScans.contains(Constants.PaymentsScanId)) {
+        digest.removeBox(box)
+      } else {
```
aslesarenko, line 304: Please leave a comment for each branch.

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
kushti: Done, swapped with already existing KeyValuePairsBag.empty

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
kushti: Made a comment on an extracted digestChanged value

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
kushti: done, made immutable

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
kushti: Done, with immutable walletAssetBalances

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
kushti: Done, with immutable walletAssetBalances

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
kushti: Done, added descriptions

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
kushti: Done

kushti:
> Does existing tests cover new/updated code?
> Do we have test coverage tracking(coveralls) setup for this repo?

Yes, there are tests in WalletRegistrySpec in regards with updated functionality:

"update scans correctly"
"remove application from a box correctly"
"remove box-scan correspondence and then rollback - one app"
"remove box-scan correspondence and then rollback - multiple apps"

No coverage tool set up yet, would be good to consider soon I guess.

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
```diff
@@ -45,13 +78,13 @@ object WalletDigestSerializer extends ScorexSerializer[WalletDigest] {
 
     val walletAssetBalancesSize = r.getUInt().toIntExact
 
-    val walletAssetBalances = mutable.LinkedHashMap.empty[EncodedTokenId, Long]
+    val walletAssetBalances = scala.collection.mutable.LinkedHashMap.empty[EncodedTokenId, Long]
```
aslesarenko, line 81: Seems like replacing LinkedHashMap with Map is going to be breaking change. 
The roundtrip property will be violated, since walletAssetBalances.toMap is not order preserving.

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletDigest.scala
kushti: Not sure this property is needed, but ok reverted and removed putBox & removeBox

aslesarenko: I didn't mean to revert back to mutable. WalletDigest is used as a message and hence its lifetime is not so obvious. As result using mutable public field is unsafe. The Scala 2.13 has https://www.scala-lang.org/api/current/scala/collection/immutable/VectorMap.html for this use case. To keep it simple, I suggest to use Seq[(EncodedTokenId, Long)], which is both immutable and good for serialization. And I think this will not be a performance problem until we frequently have more than 1000 tokens in a wallet digest.

kushti: @aslesarenko but why not to revert ?
it worked, and changing WalletDigest is not related to the PR

kushti:
@aslesarenko  made immutable Seq in wallet digest, added a test for order preservation . Not sure still that such transition is needed, and hopefully after switching to 2.13 we can use some VectorMap instance

aslesarenko (approved):

Issue #1209, opened 2020-10-07
Improve network utilization during headers sync


aslesarenko:
The following is read/write activity of the network connections, which shows the data flow is not smooth and have a lot waiting pauses.
<img width="1077" alt="image" src="https://user-images.githubusercontent.com/654077/95317688-2f848100-089e-11eb-9638-e45ecdc80c63.png">

pragmaxim:
This is due to retrieving bulks of up to 400 modifiers (by default) from peers and then parsing all of them in a single actor message : 
https://github.com/ergoplatform/ergo/blob/e6086e23ecd45f1e01a3e4c0344f003cec1a5b11/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala#L104
then this bulk is validated and applied also within a single actor message : 
https://github.com/hyperledger-labs/Scorex/blob/30f3bea5ddb660f479964b7879912cebc4ee467e/src/main/scala/scorex/core/NodeViewHolder.scala#L344

So this clogs the message throughput for a while with just a single thread being busy.

PR #1247, opened 2020-11-25
PR for Sync Protocol 2.0 

### LINKED ISSUE #1226: Sync protocol 2.0

kushti:
Close #1226


pragmaxim:
:wave: @kushti what stopped you from shipping this? I'd be glad to continue, possibly tackling #1209

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -93,6 +143,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     val typeId = data.typeId
     val modifiers = data.modifiers
     log.info(s"Got ${modifiers.size} modifiers of type $typeId from remote connected peer: $remote")
+    log.info("Modifier ids: " + modifiers.map(_._1))
```
pragmaxim, line 146: `modifiers.keys.mkString(",")`

greenhat (approved):

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -159,6 +226,46 @@ trait ErgoHistoryReader
       }
     }
 
+  /**
+    *
+    * Calculating continuation from common header which will be sent to another node
+    * if comparison status is YOUNGER of FORK, for sync message V2.
```
pragmaxim, line 232:
```suggestion
    * if comparison status is YOUNGER or FORK, for sync message V2.
```

PR #1264, opened 2021-01-20
Tests for parsing networking messages against test vectors

kushti:
This PR contains :

* `assert()` swapped with `require()` as the latter is non-elidable
* unused parameter  `nextBlockTimestampOpt` is removed from `HeadersProcessor.requiredDifficultyAfter()` 

Different test vectors and simple parsers: 
 
* invalid PoW solution validation test for manually checked test vector
* handshake parsing test for a test vector (bytes got from a 3.3.6 mainnet node)
* SyncInfo networking message parsing test (can be used as a simple spec)
* INV networking message parsing test (can be used as a simple spec)

Demo applications:
* to generate address (`AdressGenerationDemo`)
* transaction JSON printing added to `CreateTransactionDemo`


pragmaxim (commented):
Looks great, I'd improve the spacing between the `get-some-bytes` and `assert` lines so it is more readable, there are strange spaces.

src/test/scala/org/ergoplatform/network/DecodingUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+  def getULong(buf: ByteBuffer): Long = {
+    var result: Long = 0
+    var shift = 0
+    while (shift < 64) {
+      val b = getByte(buf)
+      result = result | ((b & 0x7F).toLong << shift)
+      if ((b & 0x80) == 0) return result
+      shift += 7
+    }
+    ???
```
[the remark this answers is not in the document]
kushti: done

ergo-wallet/src/test/java/org/ergoplatform/wallet/AdressGenerationDemo.java
```diff
@@ -0,0 +1,58 @@
...
+import org.ergoplatform.wallet.mnemonic.Mnemonic;
+import org.ergoplatform.wallet.secrets.DerivationPath;
+import org.ergoplatform.wallet.secrets.ExtendedPublicKey;
+import org.ergoplatform.wallet.secrets.ExtendedSecretKey;
+import scala.Option;
+
+/**
+ * This runnable example is showing how to derive change address and other addresses according to EIP-3
+ */
+public class AdressGenerationDemo {
```
greenhat, line 14:
```suggestion
public class AddressGenerationDemo {
```

src/test/scala/org/ergoplatform/network/DecodingUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+    * Read signed byte
+    */
+  def getByte(buf: ByteBuffer): Byte = buf.get
+
+  /**
+    * Reading up to 64 bits VLQ-encoded value. The value is considered to be signed.
+    * Can be read less than 64 bytes!
+    * See https://en.wikipedia.org/wiki/Variable-length_quantity
+    */
+  def getULong(buf: ByteBuffer): Long = {
```
greenhat, line 23: Why not call `VLQReader.getULong()`? Otherwise, it adds an overhead to keep these two implementations in sync.
This question is also relevant to other methods in this trait.

greenhat (changes requested):

src/test/scala/org/ergoplatform/network/DecodingUtils.scala
kushti: The tests are done in particular as examples for external application developers. Thus it was made intentionally to collect all the needed functions in one file to avoid need for traversing different classes.

greenhat (approved):

PR #1265, opened 2021-01-24
ErgoMiner - race condition and calcN() fixes

kushti:
In this PR:

* a race condition fixed in ErgoMiner. It was about a possibility for two block candidates to be generated via internal trigger and also external request. Lead to improper solutions found sometimes when external miner polling the node often (e.g. every 10 ms).
* calcN() had a bug with integer overflows towards old NIncreasementHeightMax which were unnoticed. In this PR NIncreasementHeightMax reduced to 4,198,400. This will be consensus issue 14 years after! (Means 4.0.0 & 4.0.1 nodes will diverge from 4.0.2). Should not be a big issue though.
* *knownPeers* set to ["213.239.193.208:9020"] for the testnet (as old 88... server is dead now)
* minor fixes in documentation

### DIFF
--- a/src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
+++ b/src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
@@ -281,23 +281,22 @@ class ErgoMiner(ergoSettings: ErgoSettings,
 
     // Send cached candidate if it is available and (non-empty) list of transactions to include hasn't been changed
     case PrepareCandidate(txsToInclude) =>
-      if (cachedFor(txsToInclude)) {
-        val candBlockFuture = candidateOpt
+      val candBlockFuture = if (cachedFor(txsToInclude)) {
+        candidateOpt
           .map(_.externalVersion)
           .fold[Future[WorkMessage]](
           Future.failed(new Exception("Failed to create candidate")))(Future.successful)
-        sender() ! candBlockFuture
       } else {
         log.info("Generating new candidate requested by external miner")
         val readersR = (readersHolderRef ? GetReaders).mapTo[Readers]
-        val candBlockFuture = readersR.flatMap {
+        readersR.flatMap {
           case Readers(h, s: UtxoStateReader, m, _) =>
             Future.fromTry(generateCandidate(h, m, s, Some(txsToInclude)).map(_.externalVersion))
           case _ =>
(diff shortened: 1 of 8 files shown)


src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -268,35 +270,33 @@ class ErgoMiner(ergoSettings: ErgoSettings,
...
-    case PrepareCandidate(txsToInclude) =>
-      if (cachedFor(txsToInclude)) {
-        val candBlockFuture = candidateOpt
-          .map(_.externalVersion)
-          .fold[Future[WorkMessage]](
-          Future.failed(new Exception("Failed to create candidate")))(Future.successful)
-        sender() ! candBlockFuture
+    case PrepareCandidate(txsToInclude, reply) =>
+      val candF: Future[CandidateCache] = if (candidateGenerating) {
+        Future.failed(new Exception("Candidate generation is in progress"))
```
greenhat, line 279:
```suggestion
        Future.failed(new Exception("Skipping candidate generation, one is already in progress"))
```

greenhat (approved):

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -268,35 +270,33 @@ class ErgoMiner(ergoSettings: ErgoSettings,
...
+              Future.failed(new Exception("Invalid readers state, mining is possible in UTXO mode only"))
+          }
         }
-        sender() ! candBlockFuture
+        f.onComplete(_ => candidateGenerating = false)
+        f
+      }
+      if (reply) {
+        sender() ! candF.map(_.externalVersion)
       }
```
aslesarenko, line 300: I don't like the idea of passing Future as a message in general. 
Here, more specifically, cannot imagine what the state it is when the sender() receives the Future which is still being processed in parallel somewhere in readersHolderRef and "generateCandidate" is not even started.
This part needs major rethinking, IMO.

aslesarenko (dismissed):

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
kushti: @aslesarenko that's typical for our API though

PR #1285, opened 2021-02-08
Candidate for version 4.0.5

kushti:
Candidate for version 4.0.5.

* #1281 : description for wallet/boxes/collect added 
* #1289 : the node is re-announcing new blocks received from other peers 
* #1288: API method to get Merkle proof for a transaction added
* more log output for wallet failures
* fix for #1294

### DIFF
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -1414,11 +1414,10 @@ components:
         levels:
           type: array
           items:
-            type: string
-            description: internal node with direction byte prepended
-            example: '018b7ae20a4acd23e3f1bf38671ce97103ad96d8f1c780b5e5e865e4873ae16337'
-          example: ['018b7ae20a4acd23e3f1bf38671ce97103ad96d8f1c780b5e5e865e4873ae16337']
-
+            type: array
+            description: pairs of hash and side of hash
+            example: ['018b7ae20a4acd23e3f1bf38671ce97103ad96d8f1c780b5e5e865e4873ae16337', 0]
+          example: [['018b7ae20a4acd23e3f1bf38671ce97103ad96d8f1c780b5e5e865e4873ae16337', 0]]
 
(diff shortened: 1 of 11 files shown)


src/main/scala/org/ergoplatform/http/api/BlocksApiRoute.scala
```diff
@@ -63,6 +66,21 @@ case class BlocksApiRoute(viewHolderRef: ActorRef, readersHolder: ActorRef, ergo
...
     getHistory.map(_.modifierById(modifierId))
 
+  private def getProofForTx(headerId: ModifierId, txId: ModifierId): Future[Option[MerkleProof[Digest32]]] =
+    getModifierById(headerId).flatMap {
+      case Some(header: Header) =>
+        val blockTxsId = BlockSection.computeId(
+          BlockTransactions.modifierTypeId,
+          headerId,
+          header.transactionsRoot.asInstanceOf[Array[Byte]]
+      )
```
aslesarenko, line 76:
```suggestion
        )
```

src/main/resources/api/openapi.yaml
```diff
@@ -2129,6 +2146,38 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiError'
 
+  /blocks/{headerId}/proofFor/{txId}:
```
aslesarenko, line 2149: Is it possible to get proof by using `txId` alone? This would simplify development of applications.  txId is unique and in most cases it is enough to refer to the specific transaction, but not here. Thus every future application will have to solve the same problem of finding headerId for a given txId.

kushti: @aslesarenko it is not possible as the node is not storing transactions index, and no plans to add it for minimal settings.

aslesarenko (approved):

greenhat (approved):

PR #1286, opened 2021-02-08
Fix for config issues

kushti:
In this PR:

* apiKeyHash is set to null in reference.conf to avoid "Bad API Key" message when the node is unable to parse user-provided config  
* if *ergo.directory* is set, but *ergo.wallet.secretStorage.secretDir* is not, the node is setting ergo.wallet.secretStorage.secretDir = ergo.directory + "/wallet/keystore" like in reference.conf (previously, ergo.wallet.secretStorage.secretDir = default ergo.directory from reference.conf + "/wallet/keystore"). Previous behavior led to many confusing issues, especially with PowerShell

### DIFF
--- a/src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
+++ b/src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
@@ -2,7 +2,8 @@ package org.ergoplatform.settings
 
 import java.io.{File, FileOutputStream}
 import java.nio.channels.Channels
-import com.typesafe.config.{Config, ConfigFactory}
+
+import com.typesafe.config.{Config, ConfigFactory, ConfigValueFactory}
 import net.ceedubs.ficus.Ficus._
 import net.ceedubs.ficus.readers.ArbitraryTypeReader._
 import org.ergoplatform.mining.groupElemFromBytes
@@ -12,6 +13,7 @@ import scorex.core.settings.{ScorexSettings, SettingsReaders}
 import scorex.util.ScorexLogging
 import scorex.util.encode.Base16
 import sigmastate.basics.DLogProtocol.ProveDlog
+
 import scala.util.Try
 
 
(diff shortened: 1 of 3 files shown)


greenhat (approved):

oskin1:
ApiKeyHash was intentionally set to invalid value in base conf in order to prevent users from leaving wallet methods without protection by default. In case apiKeyHash = null wallet auth is disabled so users can unintentionally leave it as is not even being aware of security vulnerability.

src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
```diff
@@ -82,6 +84,47 @@ object ErgoSettings extends ScorexLogging
     )
   }
 
+  // Helper method to read user-provided `configFile` with network-specific `fallbackConfig`
+  // to be used before
```
aslesarenko, line 88: What is means "to be used before"?

aslesarenko (approved):
LGTM, but please see comment

kushti:
> ApiKeyHash was intentionally set to invalid value in base conf in order to prevent users from leaving wallet methods without protection by default. In case apiKeyHash = null wallet auth is disabled so users can unintentionally leave it as is not even being aware of security vulnerability.


Added ``` require(ergoSettings.scorexSettings.restApi.apiKeyHash.isDefined, "API key hash must be set")``` check in ErgoApp. It is cleaner to fail than to have invalid API key.

src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
kushti: @aslesarenko improved the comment

oskin1:
> > ApiKeyHash was intentionally set to invalid value in base conf in order to prevent users from leaving wallet methods without protection by default. In case apiKeyHash = null wallet auth is disabled so users can unintentionally leave it as is not even being aware of security vulnerability.
> 
> Added ` require(ergoSettings.scorexSettings.restApi.apiKeyHash.isDefined, "API key hash must be set")` check in ErgoApp. It is cleaner to fail than to have invalid API key.

Good. But this new requirement would make testing a bit more complicated. Maybe we should add some flag like `scorex.restApi.authEnabled` which is set to `true` by default. Then if `authEnabled` apiKeyHash must be set, but if it intentionally set to `false` by a user then apiKeyHash is ignored.

kushti:
> ApiKeyHash was intentionally set to invalid value in base conf in order to prevent users from leaving wallet methods without protection by default. In case apiKeyHash = null wallet auth is disabled so users can unintentionally leave it as is not even being aware of security vulnerability.

No, 

> > > ApiKeyHash was intentionally set to invalid value in base conf in order to prevent users from leaving wallet methods without protection by default. In case apiKeyHash = null wallet auth is disabled so users can unintentionally leave it as is not even being aware of security vulnerability.
> > 
> > 
> > Added ` require(ergoSettings.scorexSettings.restApi.apiKeyHash.isDefined, "API key hash must be set")` check in ErgoApp. It is cleaner to fail than to have invalid API key.
> 
> Good. But this new requirement would make testing a bit more complicated. Maybe we should add some flag like `scorex.restApi.authEnabled` which is set to `true` by default. Then if `authEnabled` apiKeyHash must be set, but if it intentionally set to `false` by a user then apiKeyHash is ignored.

What kind of testing? Why some value can't be provided in testing config?

oskin1:
> > ApiKeyHash was intentionally set to invalid value in base conf in order to prevent users from leaving wallet methods without protection by default. In case apiKeyHash = null wallet auth is disabled so users can unintentionally leave it as is not even being aware of security vulnerability.
> 
> No,
> 
> > > > ApiKeyHash was intentionally set to invalid value in base conf in order to prevent users from leaving wallet methods without protection by default. In case apiKeyHash = null wallet auth is disabled so users can unintentionally leave it as is not even being aware of security vulnerability.
> > > 
> > > 
> > > Added ` require(ergoSettings.scorexSettings.restApi.apiKeyHash.isDefined, "API key hash must be set")` check in ErgoApp. It is cleaner to fail than to have invalid API key.
> > 
> > 
> > Good. But this new requirement would make testing a bit more complicated. Maybe we should add some flag like `scorex.restApi.authEnabled` which is set to `true` by default. Then if `authEnabled` apiKeyHash must be set, but if it intentionally set to `false` by a user then apiKeyHash is ignored.
> 
> What kind of testing? Why some value can't be provided in testing config?

I mean it is more convenient to disable wallet auth while testing.

PR #1288, opened 2021-02-11
API method to get Merkle proof for transaction added

### LINKED ISSUE #886: API method to get transaction membership proof against Merkle tree digest
Implement API method to get a transaction membership proof against transactions Merkle tree digest included into a block header.

semyonoskin:
Close #886


kushti (changes requested):
openapi.yaml description is missed

PR #1289, opened 2021-02-11
Re-announcing new blocks 

kushti:
In this PR, the node is re-announcing new blocks received.

### DIFF
--- a/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
+++ b/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
@@ -1,6 +1,7 @@
 package org.ergoplatform.network
 
 import akka.actor.{ActorRef, ActorRefFactory, Props}
+import org.ergoplatform.modifiers.history.Header
 import org.ergoplatform.modifiers.mempool.ErgoTransaction
 import org.ergoplatform.modifiers.{ErgoFullBlock, ErgoPersistentModifier}
 import org.ergoplatform.network.ErgoNodeViewSynchronizer.CheckModifiersToDownload
@@ -22,7 +23,7 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
 /**
-  * Tweaks on top of Scorex' NodeViewSynchronizer made for optimizing Ergo network
+  * Tweaks on top of Scorex' NodeViewSynchronizer made to optimize Ergo network
   */
 class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
                                viewHolderRef: ActorRef,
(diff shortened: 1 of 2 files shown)


src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -77,6 +82,130 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     networkControllerRef ! SendToNetwork(msg, SendToRandom)
   }
 
+  /**
+    * Helper method which is deciding whether chain is likely nearly of fully synchronized with the network
```
greenhat, line 86:
```suggestion
    * Helper method which is deciding whether chain is likely nearly or fully synchronized with the network
```

greenhat (approved):

oskin1 (approved):

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -77,6 +82,130 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+    */
+  private def validateAndSetStatus(remote: ConnectedPeer, pmod: ErgoPersistentModifier): Boolean = {
+    historyReaderOpt match {
+      case Some(hr) =>
+        hr.applicableTry(pmod) match {
+          case Failure(e) if e.isInstanceOf[MalformedModifierError] =>
+            log.warn(s"Modifier ${pmod.encodedId} is permanently invalid", e)
+            deliveryTracker.setInvalid(pmod.id)
+            penalizeMisbehavingPeer(remote)
+            false
```
aslesarenko, line 197: What if it is Failure with different type of exception?

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -77,6 +82,130 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+          case _ =>
+            deliveryTracker.setReceived(pmod.id, remote)
+            true
+        }
+      case None =>
+        log.error("Got modifier while history reader is not ready")
+        deliveryTracker.setReceived(pmod.id, remote)
```
aslesarenko, line 204: So in this case it got into Received state without being checked by applicableTry? So the `received` set may contain mods which are not applicable.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -77,6 +82,130 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+        val parsed: Iterable[ErgoPersistentModifier] = parseModifiers(requestedModifiers, serializer, remote)
+        val valid = parsed.filter(validateAndSetStatus(remote, _))
+        if (valid.nonEmpty) viewHolderRef ! ModifiersFromRemote[ErgoPersistentModifier](valid)
+
+        // If chain is synced or almost synced, announce new modifiers received.
+        // Helping to push new blocks around the network faster.
+        if (chainAlmostDownloaded) {
+          val toAnnounce = valid.map(_.id).toSeq
+          log.info(s"Announcing recent modifiers: ${toAnnounce.size}, type: $typeId}")
+          val msg = Message(invSpec, Right(InvData(typeId, toAnnounce)), None)
```
aslesarenko, line 128: I wonder, why the source is None?

kushti: None means there's no external source for the message (so it is sent from the node).

PR #1296, opened 2021-02-18
Docker touchup

gtklocker:
Most importantly:
Fixes an issue where the container would not receive the SIGTERM which
docker stop sends and would always have to timeout and be SIGKILLed,
potentially resulting in data corruption.

Additionally, I took the opportunity to tune it to take advantage of docker build caching and clean it up. You should see a nice improvement in building times.


kushti (approved):

Y9807:
Привет надеюсь кто-то сможет мне помочь настроить добычу из русскоговорящих

Issue #1121, opened 2020-05-21
Make a test for ErgoWalletActor.generateTransactionWithOutputs


kushti:
generateTransactionWithOutputs is big and logic is not trivial, so a test for it would be very helpful

pragmaxim:
:wave: I'm taking this one ... I'm gonna have to turn it into state machine, extracting the business logic from the actor which should be fairly easy.

While reading it I noticed of few bugs that I can fix on the way : 
https://github.com/ergoplatform/ergo/blob/6ad1132ffd1278403d51504cf818f1d660699b3d/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala#L787
this should be : 
```
    if (selectionResult.changeBoxes.nonEmpty) {
      require(changeAddressOpt.isDefined, "Does not have change address to send change to")
    }
```
Another strange thing is that when creating output boxes , we use `stateContext.currentHeight` height for the box but `registry.fetchDigest().height` for `change` box, so one `UnsignedErgoTransaction` then contains output boxes of different height ... I guess it should be unified to either one of these.

PR #1311, opened 2021-03-07
wallet state machine, refactoring, tests

pragmaxim:
:wave: I was looking at https://github.com/ergoplatform/ergo/issues/1121 and I realized that for any kind of testing :
1. the `WalletActor`'s state must be extracted and passed through `context.become(walletInit(newState))`
2. this required a serious refactoring but I did my best to preserve all the contracts and business logic ... still the changes might seem radical as most of the methods cause a state transition `oldState -> newState` 

I know that this part is very sensitive + it had not many tests, so it might seem risky to release this. Maybe we could add more tests but as I'm new to Ergo, I'm not good at writing specs yet. I started covering it with tests but I will need much more time to learn the internal working...


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWallet.scala
```diff
@@ -32,7 +32,7 @@ class ErgoWallet(historyReader: ErgoHistoryReader, settings: ErgoSettings)
   override type NVCT = this.type
 
   override val walletActor: ActorRef = {
-    val props = Props(classOf[ErgoWalletActor], settings, boxSelector, historyReader)
+    val props = Props(classOf[ErgoWalletActor], settings, new ErgoWalletServiceImpl, boxSelector, historyReader)
```
pragmaxim, line 35: `ErgoWalletService` could be just `object` as it has no state, but this way we can mock it and test the logic of the Actor, there is still some business logic, for instance Recursive scanning that must be done by message passing to not block the actor ...

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
```diff
@@ -0,0 +1,268 @@
...
+                                         parameters: Parameters,
+                                         publicKeyAddresses: Seq[P2PKAddress]): Try[Seq[ErgoBoxCandidate]] =
+    Traverse[List].sequence {
+      requests.toList
+        .map {
+          case PaymentRequest(address, value, assets, registers) =>
+            Success(new ErgoBoxCandidate(value, address.script, fullHeight, assets.toColl, registers))
+          case AssetIssueRequest(addressOpt, valueOpt, amount, name, description, decimals, registers) =>
+            // Check that auxiliary registers do not try to rewrite registers R0...R6
+            if (registers.exists(_.exists(_._1.number < 7))) {
```
pragmaxim, line 141: I changed this from `if (registers.exists(_.forall(_._1.number < 7)))` as I assume we don't want any 0-7 register to be overriden. Is that correct or ?

src/it/resources/devnetTemplate.conf
```diff
@@ -55,6 +55,6 @@ scorex {
   }
   restApi {
     bindAddress = "0.0.0.0:9051"
-    apiKeyHash = null
+    apiKeyHash = "324dcf027dd4a30a932c441f365a25e86b173defa4b8e58948253471b81b72cf"
```
pragmaxim, line 58: This is related to https://github.com/ergoplatform/ergo/issues/1316 ... when all the configs get merged, `apiKeyHash`  eventually turns into empty string which is sort of a bug ... Anyway integration tests of wallet require authentication with api key so I added it to be able to authenticate

kushti (changes requested):
Please add ScalaDoc for new classes and their public methods.

Please do some functional testing. I am getting wallet being locked after unlock (via panel).

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
kushti: Oh I guess you're right. Would be good to cover this with a test.

l15k4: I covered that already, I was just not sure if the AssetIssueRequest aux registers really cannot override any of the R0-R6 https://github.com/ergoplatform/ergo/pull/1311/files#diff-fca9ab9b22f435644bbc31e76ac9b4e4fb37e61555dd05e51a1c3763ad8f9ff1R90-R93

pragmaxim: I covered that already, I was just not sure if the AssetIssueRequest aux registers really cannot override any of the R0-R6 https://github.com/ergoplatform/ergo/pull/1311/files#diff-fca9ab9b22f435644bbc31e76ac9b4e4fb37e61555dd05e51a1c3763ad8f9ff1R90-R93

src/test/scala/org/ergoplatform/local/MempoolAuditorSpec.scala
```diff
@@ -80,7 +80,7 @@ class MempoolAuditorSpec extends AnyFlatSpec with NodeViewTestOps with ErgoTestH
 
     getPoolSize shouldBe 1 // first tx removed from pool during node view update
 
-    scorex.core.utils.untilTimeout(cleanupDuration * 3, 100.millis) {
+    scorex.core.utils.untilTimeout(cleanupDuration * 4, 100.millis) {
```
pragmaxim, line 83: @kushti You probably know but this test often fails due to busy Travis. It did not help though

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
kushti: Yes, that was an intention (aux registers shouldn't override R0-R6)

src/it/resources/devnetTemplate.conf
kushti: Looks good

kushti (approved):

PR #1317, opened 2021-03-14
Pagination added to wallet transactions/transactionsByScanId method

### LINKED ISSUE #1271: Add pagination to scan-related transactions API
Potentially scan-related transactions method can return too many transactions to fit into one response, so pagination needs to be added

semyonoskin:
close #1271

### DIFF
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -3080,6 +3080,24 @@ paths:
   /wallet/transactions:
     parameters:
       - in: query
+        name: offset
+        required: false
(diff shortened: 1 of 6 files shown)


kushti (changes requested):
I think there's some misunderstanding in this implementation. A client can limit response  size via {min/max}InclusionHeight and {min/max}Confirmations parameters. What is needed, is to avoid reading all the transactions from the wallet for that (as it can be slow in the presence of many transactions e.g. in case of an exchange wallet). For that, additional by-height index is needed.

PR #1320, opened 2021-03-26
Different wallet optimizations

kushti:
In this PR: 

*  filtering height is taken into account in the DB layer when transactions are read from it
* *WalletRegistryBenchmark* to track most significant read operations

Issue #1316, opened 2021-03-13
non-deterministic configuration loading 


pragmaxim:
While writing integration tests for https://github.com/ergoplatform/ergo/issues/1316 I noticed, that If you print the config with `cfg.root().get("scorex").render(ConfigRenderOptions.concise())`, then this
contains still `apiKeyHash = null` : 
https://github.com/ergoplatform/ergo/blob/master/src/main/scala/org/ergoplatform/settings/ErgoSettings.scala#L92

but this contains empty string : 
https://github.com/ergoplatform/ergo/blob/master/src/main/scala/org/ergoplatform/settings/ErgoSettings.scala#L108 

and as a result this boots fine as `apiKeyHash` is defined but empty string : 
https://github.com/ergoplatform/ergo/blob/master/src/main/scala/org/ergoplatform/ErgoApp.scala#L38

imho this : 
```
    val firstFallBack = fallbackConfig.map(ConfigFactory.parseFile).getOrElse(ConfigFactory.defaultApplication())
    val fullConfig = ConfigFactory
      .defaultOverrides()
      .withFallback(cfg)
      .withFallback(firstFallBack)
      .withFallback(ConfigFactory.defaultReference())
      .resolve()
```
introduces non-deterministic behavior, obviously there are at least 2 `apiKeyHash` keys on classpath : `null` and empty string and in this case `null` value causes fallback to empty string value.

I can play with it but I would suggest to **always** apply `defaultApplication` and to remove `defaultReference`... This way it should be at least consistent regardless of what arguments come.
But as security is involved here, we should be careful about not loading some default values from scorex reference files.

pragmaxim:
@kushti fixed in https://github.com/ergoplatform/ergo/pull/1328

PR #1324, opened 2021-04-05
Candidate for 4.0.10

kushti:
Candidate for 4.0.10. It contains: 

* #1264 - tests (and also reference parsers) for p2p network messages
* #1319 - default amount in payment send form raised from 0.001 to 0.01. Also, a bug with passwords caching fixed. 
* #1320 - filtering height is taken into account in the DB layer when transactions are read from it (critical for wallets with high load, e.g. exchanges)
* #1328 - fix for non-deterministic configurations loading (issue #1316 )
* #1335 - starting external mining with just pubkey from wallet
* #1341 - reproducible builds

Contributors: @deadit , @kushti , @notelos ,  @pragmaxim


src/test/scala/org/ergoplatform/network/DecodingUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+    var result: Long = 0
+    var shift = 0
+    while (shift < 64) {
+      val b = getByte(buf)
+      result = result | ((b & 0x7F).toLong << shift)
+      if ((b & 0x80) == 0) return result
+      shift += 7
+    }
+    throw new IllegalStateException("Trying to read long, but more bytes than needed found")
+  }
```
aslesarenko, line 33: This duplicates code already present in scorex-util, which is in Ergo's dependencies. https://github.com/ScorexFoundation/scorex-util/blob/master/src/main/scala/scorex/util/serialization/VLQReader.scala

src/test/scala/org/ergoplatform/network/TransactionSpecification.scala
```diff
@@ -0,0 +1,80 @@
...
+import scorex.util.encode.Base16
+
+class TransactionSpecification extends ErgoPropertyTest with DecodingUtils {
+
+  property("transaction specification parser") {
+    /** Header flag to indicate that constant segregation should be applied. */
+    val ConstantSegregationFlag: Byte = 0x10
+
+    /** Header flag to indicate that whole size of ErgoTree should be saved before tree content. */
+    val SizeFlag: Byte = 0x08
```
aslesarenko, line 15: These flags are defined in ErgoTree from Sigma. It make sense to refer to either sigma definition or to ErgoTree Specification document.

src/test/scala/org/ergoplatform/network/TransactionSpecification.scala
```diff
@@ -0,0 +1,80 @@
...
+    val bb = ByteBuffer.wrap(txBytes)
+
+    val inputsCount = getULong(bb) // must be within  0..0xFFFF range
+    inputsCount shouldBe 2
+
+    val firstInputId = getBytes(bb, 32)
+    Base16.encode(firstInputId) shouldBe "430e80ca31a25400e77dac0ad14c1cd39cb09dc3f7c1c384dce9aef19b604e27"
+
+    val firstSigLen = getULong(bb).toInt // must be within  0..0xFFFF range
+    val firstSig = getBytes(bb, firstSigLen)
```
aslesarenko, line 31: I suggest to use original serializers from scorex-util (instead of DecodingUtils), which are used in both Sigma and Ergo. In this case this tests will also serve as test vectors for those serializers (to catch regression bugs)

pragmaxim (approved):
:+1:

deadit (approved):
All is good at panel

pragmaxim:
@kushti hot fix for a node without initialized wallet https://github.com/ergoplatform/ergo/pull/1342

PR #1328, opened 2021-04-09
#1316 non-deterministic-configs-loading

### LINKED ISSUE #1316: non-deterministic configuration loading

pragmaxim:
Fixes #1316 

Story : ergo should not have `reference.conf` as 
1. it is application, not library (`ergo-wallet` is a library)
2. it shares `scorex {` scope with with `reference.conf` from scorex library, that's not recommended -> non-deterministic loading
---------------------------------------------------------------
1. The tests pass, `org.ergoplatform.nodeView.mempool.ErgoMemPoolSpec` was failing even before, I cannot find out why.
2. I tested that on `mainnet` and it seems to work fine
3. I made sure that only `ergo-wallet` is published and used from `ergo-appkit`, so nothing is using `"org.ergoplatform" %% "ergo"`, it is not even published, which means the change from `reference.conf` -> `application.conf` should be legitimate.


src/it/resources/devnetTemplate.conf
```diff
@@ -55,6 +55,6 @@ scorex {
   }
   restApi {
     bindAddress = "0.0.0.0:9051"
-    apiKeyHash = "324dcf027dd4a30a932c441f365a25e86b173defa4b8e58948253471b81b72cf"
+    apiKeyHash = null
```
pragmaxim, line 58: Reverting that as it was before, as it works now.

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -117,14 +117,14 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool, private[mempool] val sta
           case utxo: UtxoState =>
             // Allow proceeded transaction to spend outputs of pooled transactions.
             utxo.withTransactions(getAll).validate(tx).fold(
-              new ErgoMemPool(pool.invalidate(tx), stats) -> ProcessingOutcome.Invalidated(_),
+              ex => new ErgoMemPool(pool.invalidate(tx), stats) -> ProcessingOutcome.Invalidated(ex),
```
pragmaxim, line 120: I was troubleshooting failing ErgoMempool test ... I found this much more readable :shrug:

src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
```diff
@@ -109,6 +109,7 @@ object ErgoSettings extends ScorexLogging
       .defaultOverrides()
       .withFallback(cfg)
       .withFallback(firstFallBack)
+      .withFallback(ConfigFactory.defaultApplication())
```
pragmaxim, line 112: having `reference.conf` on ergo's classpath with  `scorex {` scope conflicts with `reference.conf` from scorex framework ... typesafe config merges configs from these files sort of as classloader loads it from the classpath (non-deterministically, it depends on underlying JVM type) ... I guess if we want ergo's scorex configs to override scorex's scorex configs, we should use `application.conf`  instead

kushti (approved):

PR #1335, opened 2021-04-17
Starting external mining with just pubkey from wallet, scanning mining

pragmaxim:


src/test/resources/application.conf
```diff
@@ -25,7 +25,7 @@ ergo {
     minimalSuffix = 10
 
     # Is the node is doing mining
-    mining = false
+    mining = true
```
pragmaxim, line 28: @kushti a lot of tests were failing because they use the mining reward as Input to make transactions ...

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -88,16 +88,17 @@ class ErgoMiner(ergoSettings: ErgoSettings,
...
 
   // internal (CPU) miner threads, if internal mining is chosen
   private val miningThreads: mutable.Buffer[ActorRef] = new ArrayBuffer[ActorRef]()
 
   override def preStart(): Unit = {
-    // in external miner mode key from wallet is used if `publicKeyOpt` is not set
-    if ((publicKeyOpt.isEmpty && externalMinerMode) || (secretKeyOpt.isEmpty && !externalMinerMode)) {
+    if (secretKeyOpt.isEmpty && !externalMinerMode) {
+      self ! QueryWallet(secret = true)
+    } else if (publicKeyOpt.isEmpty) { // mining pubKey is needed in both modes
```
pragmaxim, line 99: I removed `externalMinerMode` condition here as PK is needed in both modes, that's why all `ErgoMinerSpec` tests were failing, PK was absent and no candidate was generated ...

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -239,9 +253,9 @@ class ErgoMiner(ergoSettings: ErgoSettings,
   private def generateCandidate(h: ErgoHistoryReader,
                                 m: ErgoMemPoolReader,
                                 s: UtxoStateReader,
-                                txsToIncludeOpt: Option[Seq[ErgoTransaction]]): Try[CandidateCache] = {
+                                txsToInclude: Seq[ErgoTransaction]): Try[CandidateCache] = {
     //mandatory transactions to include into next block taken from the previous candidate
-    val txsToInclude = txsToIncludeOpt.getOrElse(candidateOpt.map(_.txsToInclude).getOrElse(Seq.empty)).filter { tx =>
+    val unspentTxsToInclude = txsToInclude.filter { tx =>
```
pragmaxim, line 258: `Option` was legacy here, it was always `Some`

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -458,7 +472,7 @@ class ErgoMiner(ergoSettings: ErgoSettings,
     log.info("Requesting candidate")
     candidateOpt = None
     candidateGenerating = false
-    self ! PrepareCandidate(candidateOpt.map(_.txsToInclude).getOrElse(Seq.empty), reply = false)
+    self ! PrepareCandidate(Seq.empty, reply = false)
```
pragmaxim, line 475: We set `candidateOpt = None` so no need for that construct.

src/test/resources/application.conf
kushti: Ok, but then add mining = false to mainnet.conf to avoid running miner for the existing mainnet nodes (for the testnet maybe that's okay to have mining switched on by default)

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
kushti: True, thanks

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
kushti: Good catch!

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
kushti: Good catch!

kushti (approved):

kushti:
@pragmaxim approved, but please add mining = false to mainnet.conf to avoid running miner for the existing mainnet nodes (for the testnet maybe that's okay to have mining switched on by default)

src/test/resources/application.conf
pragmaxim: I checked that already, in `test/it` scope, all configs are explicitly set to `mining = false` already except for `devnetTemplate.conf` with `mining = true`.

Issue #1326, opened 2021-04-08
API /blocks?limit=1 returns []


c8e4d2a:
version: 4.0.8

steps to reproduce:
```
curl http://localhost:9053/blocks?limit=1
```

expected behavior:
api call returns a set with the size 1

PR #1337, opened 2021-04-22
Fix for #1326: /blocks?limit=N now returns N block ids, not N-1

kushti:
As the genesis block has height = 1, and offset value by default was 0, /blocks?limit=N returned N-1 block ids. Fixed in this PR, now /blocks?limit=N now returns N block ids.


pragmaxim (approved):
:+1:

greenhat (approved):

PR #1353, opened 2021-05-17
TransactionBuilder can bypass creating feeOut box

aslesarenko:
This is necessary to allow user-define fee outputs.
Related Appkit issue https://github.com/ergoplatform/ergo-appkit/issues/77

### DIFF
--- a/ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala
+++ b/ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala
@@ -115,7 +115,7 @@ object TransactionBuilder {
 
     require(!changeGoesToFee || (changeAmt == 0 || createFeeOutput.isDefined),
       s"""When change=$changeAmt < minChangeValue=$minChangeValue it is added to miner's fee,
-        |it this case createFeeOutput should be defined
+        |in this case createFeeOutput should be defined
         |""".stripMargin)
 
     val feeOutOpt = createFeeOutput.map { fee =>
--- a/ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala
+++ b/ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala
@@ -49,17 +49,17 @@ object TransactionBuilder {
     require(inputs.distinct.size == inputs.size, s"There should be no duplicate inputs")
   }
 
-
   /** Creates unsigned transaction from given inputs and outputs adding outputs with miner's fee and change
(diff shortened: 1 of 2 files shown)


ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala
```diff
@@ -110,17 +111,24 @@ object TransactionBuilder {
...
-    val feeOut = new ErgoBoxCandidate(
-      actualFee,
-      ErgoScriptPredef.feeProposition(minerRewardDelay),
-      currentHeight
-    )
+    val changeGoesToFee = changeAmt < minChangeValue && !changeBoxesHaveTokens
+
+    require(!changeGoesToFee || (changeAmt == 0 || createFeeOutput.isDefined),
+      s"""When change=$changeAmt < minChangeValue=$minChangeValue it is added to miner's fee,
+        |it this case createFeeOutput should be defined
```
greenhat, line 118:
```suggestion
        |in this case createFeeOutput should be defined
```

greenhat (approved):

ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala
```diff
@@ -68,7 +68,7 @@ object TransactionBuilder {
     dataInputs: IndexedSeq[DataInput],
     outputCandidates: Seq[ErgoBoxCandidate],
     currentHeight: Int,
-    feeAmount: Long,
+    createFeeOutput: Option[Long],
```
kushti, line 71: feeAmountOpt: Option[Long]

"create..." sounds like a function or a boolean parameter  name

kushti (approved):

ergo-wallet/src/main/scala/org/ergoplatform/wallet/transactions/TransactionBuilder.scala
aslesarenko: yeah, but feeAmountOpt hides the semantics of why it is optional, i.e. that when it is None, then new feeOut is not created, and then it is Some, then it is created. That is why "create" in the name.
I suggest to leave it with "create", because parameter is hard to confuse with method.

scalahub:
Doesn't the node already allow "exact" transactions? Referring to  to this endpoint `/wallet/transaction/generate`. From the documentation:

     ... You may specify boxes to spend by providing them in "inputsRaw". Please note you need to have strict 
     equality between input and output total amounts of Ergs in this case. If you want wallet to pick up the 
     boxes, leave "inputsRaw" empty.

So doesn't this mean that the transaction must be "exact" when the `inputsRaw` is not empty?

PR #1355, opened 2021-05-18
Bucketing modifiers coming from peers

pragmaxim:
Improving logic of modifier downloading for better efficiency. 
- this change would be backward compatible with older peer versions
- it fully syncs with default `xmx`
- resource utilization is higher and more evenly distributed in time, ie. no peeks
- it constraints modifier batch sizes (min/max) to be downloaded from peers
- it utilizes all available peers, Unknown/Fork only as a fallback when Even/Older is not available

Check the buckets of headers and later `102/108` modifiers : 
https://gist.githubusercontent.com/pragmaxim/a0d7aadc541b29c1b8dc919df563462e/raw/40d7fbee9070f5a1139914ee16f413927f46c91a/sync.log


src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -56,31 +59,115 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
-  override protected def requestDownload(modifierTypeId: ModifierTypeId, modifierIds: Seq[ModifierId]): Unit = {
-    deliveryTracker.setRequested(modifierIds, modifierTypeId, None)
-    val msg = Message(requestModifierSpec, Right(InvData(modifierTypeId, modifierIds)), None)
-    networkControllerRef ! SendToNetwork(msg, SendToRandom)
-  }
+  protected def requestDownload(maxElements: Int, minElementsPerBucket: Int, maxElementsPerBucket: Int)
+                               (getPeers: => Option[(PeerSyncState, Iterable[ConnectedPeer])])
+                               (fetchMax: Int => Map[ModifierTypeId, Seq[ModifierId]]): Unit =
+    getPeers
+      .foreach { case (peerStatus, peers) =>
```
kushti, line 154: Why to ask every status for modifier ids ?

pragmaxim: `getPeers` returns `Option` of peers of certain `PeerSyncState`

kushti (approved):

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -56,31 +59,124 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     context.system.scheduler.scheduleAtFixedRate(toDownloadCheckInterval, toDownloadCheckInterval, self, CheckModifiersToDownload)
   }
 
+  private def downloadRequired(historyReader: ErgoHistory)(id: ModifierId): Boolean =
+    deliveryTracker.status(id, Seq(historyReader)) == ModifiersStatus.Unknown
```
aslesarenko, line 63: @pragmaxim Seq constructor is quite slow comparing to Array, which can safely be used here, because the parameter is of type Seq, and the array will be wrapped into WrappedArray.
Please read this document and try to use those recommendations https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/perf-style-guide.md

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -56,31 +59,124 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+  }
 
-        toDownload.groupBy(_._1).foreach(ids => requestDownload(ids._1, ids._2.map(_._2)))
+  /**
+    * Headers should be downloaded from an Older node, it is triggered by received sync message from an older node
+    * @param callingPeer that can be used to download headers, it must be Older
+    * @return available peers to download headers from together with the state/origin of the peer
+    */
+  private def getPeersForDownloadingHeaders(callingPeer: ConnectedPeer): (PeerSyncState, Iterable[ConnectedPeer]) =
+    statusTracker.peersByStatus.get(Older).map(PeerSyncState.Older -> _).getOrElse(PeerSyncState.OlderCalling -> Iterable(callingPeer))
```
aslesarenko, line 132: Same as with Seq, Iterable constructor here uses the same builder (backed by List) and can be replaced with Array

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -56,31 +59,124 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+
+  /**
+    * Other persistent modifiers besides headers should be downloaded from either Older or Equal node, with fallback to Unknown or Fork
+    * @return available peers to download persistent modifiers from together with the state/origin of the peer
+    */
+  private def getPeersForDownloadingBlocks: Option[(PeerSyncState, Iterable[ConnectedPeer])] = {
+    val peersByStatus = statusTracker.peersByStatus
+    Option(peersByStatus.getOrElse(Older, Iterable.empty) ++ peersByStatus.getOrElse(Equal, Iterable.empty))
```
aslesarenko, line 140: Here Iterable.empty can be replaced by WrappedArray.empty which allows to avoid allocations, and way much faster.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -36,20 +37,26 @@ trait ToDownloadProcessor extends BasicReaders with ScorexLogging {
...
+      if (acc.values.exists(_.lengthCompare(howManyPerType) >= 0)) {
+        acc.mapValues(_.take(howManyPerType)).view.force
       } else {
         val headersAtThisHeight = headerIdsAtHeight(height).flatMap(id => typedModifierById[Header](id))
 
         if (headersAtThisHeight.nonEmpty) {
           val toDownload = headersAtThisHeight.flatMap(requiredModifiersForHeader).filter(m => condition(m._2))
-          continuation(height + 1, acc ++ toDownload)
+          // add new modifiers to download to accumulator
+          val newAcc = toDownload.foldLeft(acc) { case (newAcc, (mType, mId)) => newAcc.adjust(mType)(_.fold(Seq(mId))(_ :+ mId)) }
```
aslesarenko, line 58: Looks like using `_ :+ mId` in a loop makes this quadratic i.e. O(toDownload.size ^ 2).
What is the typical size of toDownload?
Consider using some builder here and make both the logic more explicit and performance better. 
It is not functional though, but I would prefer performance, since this is system level code.

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
pragmaxim: Ah, sorry, frankly I haven't used `Seq()` anywhere in past 10 years and I thought it was backed by `Vector`, I've always gone for concrete constructor in the past or Builders, but they don't have `size/length` information.

There is just 5-20 elements in it, I guess it would deserve a rewrite to something more efficient ...

pragmaxim: @kushti @aslesarenko guys let me do this in next iteration, I just changed it to `Vector` from `List` for now, for instance `HeadersProcessor` is full of `Seq()` and `Seq.empty` ... I did not even know about this constructor

pragmaxim:
I just tested this version and it works nicely, it's very fast now.
Screenshot from 2021-05-30 16-47-03

aslesarenko (approved):
LGTM

Issue #1309, opened 2021-03-05
/mining/candidate returns 500


neal-zhu:
Hi, we call /mining/candidate api every 50ms, and we get some error response like:
```
{\n  \"error\" : 500,\n  \"reason\" : \"internal.error\",\n  \"detail\" : \"Skipping candidate generation, one is already in progress\"\n}
{\n  \"error\" : 500,\n  \"reason\" : \"internal.error\",\n  \"detail\" : \"Can not generate block candidate: chain not synced (maybe last block not fully applied yet\"\n}
```
And we found our node sometime does not sync with main-net which cause many orphan blocks.
Could you guys help us with this?

pragmaxim:
@neal-zhu this should be fixed in the release after `4.0.10` ... I will update this issue when it is

pragmaxim:
@neal-zhu hi, it addressed in 4.0.13

xmrstickers:
initialize the wallet and unlock it through the 9053 admin panel to avoid this issue if you are experiencing it post 4.0.13

Issue #914, opened 2019-09-23
CPU and memory consumption when many miners are connected


kushti:
It is reported that when there are many miners (like 100) connected, a protocol client consumes a lot of CPU (400%) and memory. We need to do profiling to find possible resource leaks.

pragmaxim:
I think I know why this happens, we use unbounded mailboxes and 100 miners querying 20 reqs/s might produce 200 reqs/s and I believe that at time of `PrepareCandidate` which takes quite a while, the requests queue up ... The solution might be in introducing some sort of backpressure to slow miners down.

I guess in general this might be resolved by internal `retries` in combination with some Bounded mailbox to avoid DDOS attack.

pragmaxim:
@kushti I think I found the real cause of this. Miners compete for block candidate and we have CPU heavy operation in `onComplete` https://github.com/ergoplatform/ergo/blob/04c1424066255cfc4aa6bbe0917686b860065d49/src/main/scala/org/ergoplatform/mining/ErgoMiner.scala#L318

I believe that we sohuld eliminate the `ask` pattern ` (readersHolderRef ? GetReaders).mapTo[Readers]` and simply subscribe for the NodeView changes and keep it as local state.

PR #1364, opened 2021-05-25
backpressure miners

pragmaxim:
It addresses : 
https://github.com/ergoplatform/ergo/issues/1309
https://github.com/ergoplatform/ergo/issues/914

and should be backward compatible with miners. I will do more tests.


kushti (approved):

pragmaxim:
@kushti Btw I was a bit worried here that `mailbox-capacity = 1000` is not enough as preparing the candidate takes a long time during which messages could enqueue up to thousands, if we release this, we might consider `mailbox-capacity = 5000` in case of those nodes with 100 miners

Issue #1386, opened 2021-06-07
Node can't sync


nevata:
![image](https://user-images.githubusercontent.com/34930450/120958689-d05b7100-c78a-11eb-8723-b66cee3f7e4a.png)

nevata:
ergo@wyatt-ubuntu:~/ergo_node$ java -jar ergo-4.0.11.jar --mainnet -c ergo.conf
13:56:26.157 INFO  [main] org.ergoplatform.ErgoApp - Running with args: Args(Some(ergo.conf),Some(MainNet))
13:56:26.217 INFO  [main] o.e.settings.ErgoSettings$ - Running in mainnet network mode
13:56:28.377 INFO  [main] org.ergoplatform.ErgoApp - Working directory: /home/ergo/ergo_node/.ergo
13:56:28.378 INFO  [main] org.ergoplatform.ErgoApp - Secret directory: /home/ergo/ergo_node/.ergo/wallet/keystore
13:56:29.355 ERROR [ctor.default-dispatcher-6] s.c.n.NetworkController - Declared address validation failed:
 /58.247.92.146:9020 not match any of the listening address: Set(/10.10.3.203:9020, /127.0.0.1:9020)
 or Gateway WAN address: None
13:56:29.365 INFO  [ctor.default-dispatcher-6] s.c.n.NetworkController - Declared address: Some(/58.247.92.146:9020)
13:56:29.379 INFO  [ctor.default-dispatcher-8] o.e.mining.ErgoMiner - Trying to use public key from wallet for mining
13:56:29.472 INFO  [tor.default-dispatcher-14] s.c.n.NetworkController - Registering handlers for List((1,GetPeers message), (2,Peers message))
13:56:29.489 INFO  [ctor.default-dispatcher-8] s.c.n.NetworkController - Successfully bound to the port 9020
13:56:29.609 INFO  [tor.default-dispatcher-13] s.c.u.NetworkTimeProvider - New offset adjusted: -29
13:56:29.614 WARN  [ergoref-api-dispatcher-9] o.e.n.ErgoReadersHolder - Got GetReaders request in state (None,None,None,None)
13:56:29.707 INFO  [tor.default-dispatcher-10] s.c.n.NetworkController - Registering handlers for List((55,Inv), (22,RequestModifier), (33,Modifier), (65,Sync))
[...]

PR #1391, opened 2021-06-15
avoiding silent wallet failure

pragmaxim:
Related to https://github.com/ergoplatform/ergo/issues/1386

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -190,7 +190,13 @@ class ErgoWalletActor(settings: ErgoSettings,
           historyReader.bestFullBlockAt(blockHeight) match {
             case Some(block) =>
               log.info(s"Wallet is going to scan a block ${block.id} in the past at height ${block.height}")
-              ergoWalletService.scanBlockUpdate(state, block)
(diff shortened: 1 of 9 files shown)


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -190,7 +190,13 @@ class ErgoWalletActor(settings: ErgoSettings,
           historyReader.bestFullBlockAt(blockHeight) match {
             case Some(block) =>
               log.info(s"Wallet is going to scan a block ${block.id} in the past at height ${block.height}")
-              ergoWalletService.scanBlockUpdate(state, block)
+              ergoWalletService.scanBlockUpdate(state, block) match {
+                case Failure(ex) =>
+                  log.error("Scanning past block update failed", ex)
+                  state
```
pragmaxim, line 196: @kushti I'm wondering whether we should stop scanning in the past if this happens or continue :shrug: We stopped here before this pull request in case of a failure.

kushti: @pragmaxim I guess if we stop, there should be some new wallet status introduced (Failed) and the status should be passed into the UI. Sounds like a topic for another issue

kushti (approved):

PR #1412, opened 2021-07-26
Regenerate block candidate periodically

pragmaxim:


pragmaxim:
I cannot explain the compilation errors, the last commit was not related to it and cross compilation works fine locally :shrug:

kushti:
> @kushti I cannot explain the compilation errors, the last commit was not related to it and cross compilation works fine locally shrug I tried to clean the sbt cache but it did not help ...

fixed, some merging error I guess

pragmaxim:
@kushti This would be good to merge into `master` as there is merging overhead, wdyt ?

PR #1414, opened 2021-07-30
handle db errors at higher level

pragmaxim:
See https://github.com/hyperledger-labs/Scorex/pull/397

The error propagation is done by `Try` so there might be some `map` / `flatMap` overhead, I could probably reduce it by throwing and wrapping into `Try` at higher level to spare couple of instantiations per update.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
@@ -8,7 +8,7 @@ import scorex.core.utils.ScorexEncoding
 import scorex.db.{ByteArrayWrapper, LDBKVStore}
 import scorex.util.{ModifierId, ScorexLogging, idToBytes}
 
-import scala.util.{Failure, Success}
+import scala.util.{Failure, Success, Try}
 
 /**
   * Storage for Ergo history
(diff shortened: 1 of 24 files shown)


src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
```diff
@@ -58,23 +58,27 @@ class HistoryStorage(indexStore: LDBKVStore, objectsStore: LDBKVStore, config: C
...
-    if (indexesToInsert.nonEmpty) {
-      indexesToInsert.foreach(kv => indexCache.put(kv._1, kv._2))
-      indexStore.insert(indexesToInsert.map { case (k, v) => k.data -> v })
+    ).flatMap { _ =>
+      objectsToInsert.foreach(o => modifiersCache.put(o.id, o))
+      if (indexesToInsert.nonEmpty) {
+        indexStore.insert(indexesToInsert.map { case (k, v) => k.data -> v }).map { _ =>
+          indexesToInsert.foreach(kv => indexCache.put(kv._1, kv._2))
+          ()
+        }
```
pragmaxim, line 70: I changed this so that cache is modified only if DB does not throw error.

kushti (approved):

PR #1427, opened 2021-08-12
coordinated shutdown

### LINKED ISSUE #1425: Implement akka CoordinatedShutdown
While reviewing the code, I noticed there are places that just `System.exit` which might cause some DB corruptions. I have good experience with akka's CoordinatedShutdown which is great for closing resources on shutdown etc.

Also all actors should have supervision strategy to log exceptions that are thrown from the Actor thread.

pragmaxim:
Closes #1425  ... tests fail because it depends on newest scorex

### DIFF
--- a/src/main/scala/org/ergoplatform/ErgoApp.scala
+++ b/src/main/scala/org/ergoplatform/ErgoApp.scala
@@ -191,11 +191,19 @@ object ErgoApp extends ScorexLogging {
       help("help").text("prints this usage text")
   }
 
+  /** Internal failure causing shutdown */
   case object InternalShutdown extends CoordinatedShutdown.Reason
+
+  /** Intentional user invoked remote shutdown */
   case object RemoteShutdown extends CoordinatedShutdown.Reason
+
+  /** Exception that triggers proper system shutdown */
   case class CriticalSystemException(message: String) extends Exception(message)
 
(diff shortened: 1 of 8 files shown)


src/main/scala/org/ergoplatform/ErgoApp.scala
```diff
@@ -166,28 +165,7 @@ class ErgoApp(args: Args) extends ScorexLogging {
...
 
     Http().newServerAt(bindAddress.getAddress.getHostAddress, bindAddress.getPort).bindFlow(httpService.compositeRoute)
-
-    //on unexpected shutdown
-    Runtime.getRuntime.addShutdownHook(new Thread() {
-      override def run() {
-        log.error("Unexpected shutdown")
-        stopAll()
-      }
-    })
```
aslesarenko, line 176: Why this hook is not necessary?

pragmaxim: I ported it to a CoordinatedShutdown task which is triggered on this hook implicitly ... One would have to explicitly disable it : 
```
akka.coordinated-shutdown.run-by-jvm-shutdown-hook=off
```

kushti:
@pragmaxim please add ScalaDoc for new (or touched) classes and methods, and add comments to the new code

kushti (approved):

PR #1442, opened 2021-09-10
smaller tokens count per box limit

pragmaxim:


src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -285,7 +284,7 @@ object ErgoTransaction extends ApiCodecs with ScorexLogging with ScorexEncoding
   def extractAssets(boxes: IndexedSeq[ErgoBoxCandidate]): Try[(Map[ByteArrayWrapper, Long], Int)] = Try {
     val map: mutable.Map[ByteArrayWrapper, Long] = mutable.Map[ByteArrayWrapper, Long]()
     val assetsNum = boxes.foldLeft(0) { case (acc, box) =>
-      require(box.additionalTokens.length <= ErgoTransaction.MaxAssetsPerBox, "too many assets in one box")
+      require(box.additionalTokens.length <= MaxAssetsPerBox, "too many assets in one box")
```
kushti, line 287: This change is causing hard-fork, as old nodes and ones with this code will diverge. 

Please leave verification protocol (the Ergo protocol) intact

src/main/scala/org/ergoplatform/settings/ValidationRules.scala
```diff
@@ -48,7 +49,7 @@ object ValidationRules {
     txInputsUnique -> RuleStatus(s => fatal(s"There should be no duplicate inputs. $s"),
       Seq(classOf[ErgoTransaction]),
       mayBeDisabled = false),
-    txAssetsInOneBox -> RuleStatus(s => fatal(s"A number of tokens within a box should not exceed ${ErgoTransaction.MaxAssetsPerBox}" +
+    txAssetsInOneBox -> RuleStatus(s => fatal(s"A number of tokens within a box should not exceed $MaxAssetsPerBox" +
```
kushti, line 52: Please leave the message intact as well,

kushti (changes requested):

Xaeda:
Hi, just an observation I had - if tokens have metadata and are "full" NFT tokens, then 100 per box is still too large. We settled on 10 per box as a 'safe' limit, but I am considering forcing the Ergnomes code to check the proposed output box size given on the response from transaction/generate, and then if greater than 4096, calculate how many boxes would be needed to store the tokens, then update the transaction/send call to create the required number of boxes for the 'change' tokens. Would this be a solid approach here too, instead of trying to find a 'sweet spot'?

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
pragmaxim: Fixed, I just removed the object reference `ErgoTransaction.` as we are withing the object scope ...

pragmaxim:
@HazeyOneKenobi @kushti  I'm not sure if this is only about the change boxes, if so, we could maybe find the sweet spot dynamically in the box selector by checking the size of tokens perhaps?

Issue #1443, opened 2021-09-10
Non-atomic update of HistoryStorage on new modifier received


kushti:
Investigating an issue with a node being stuck, I realized that sometimes node's database can report that modifier (header) id us there but modifier itself is not:

modifierById(ModifierId @@ "96f68654faa2c26381c25426c3fc634ba9a31eb04127ad39c85f7454bf157e17") // None
contains(ModifierId @@ "96f68654faa2c26381c25426c3fc634ba9a31eb04127ad39c85f7454bf157e17") // true

while these methods must be consistent

pragmaxim:
@kushti Can you please parse Node's logs for `Failed to parse modifier` ? 
https://github.com/ergoplatform/ergo/blob/e6086e23ecd45f1e01a3e4c0344f003cec1a5b11/src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala#L43

As believe that this `HistoryModifierSerializer.parseBytesTry(bytes)` is the source of unnecessary problems. Common logic says that if it successfully serializes, it should also succeed in deserialization. Or is it some backward compatibility thing?

kushti:
effect fixed in #1494 (on node restart)

kushti:
found that corresponding BlockTransactions section is invalid in a database I have

pragmaxim:
@kushti do we consider this closed by your fix #1494 ?

PR #1446, opened 2021-09-14
consistent HistoryStorage.contains

pragmaxim:
Possible solution to https://github.com/ergoplatform/ergo/issues/1443 ... Unfortunately without logs it is really hard to discover the real reason, this is a possible one ... Downside is a bit of a performance downgrade.

**UPDATE** this is just about logging a case which would corrupt the state.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
@@ -165,6 +165,7 @@ trait ErgoHistory
         }
       case None =>
         //No headers become invalid. Just valid this modifier as invalid
+        log.warn(s"Modifier ${modifier.encodedId} of type ${modifier.modifierTypeId} is missing corresponding header")
         historyStorage.insert(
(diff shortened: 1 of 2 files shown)


kushti:
Still,

 println("contains : " + historyReaderOpt.get.contains(ModifierId @@ "96f68654faa2c26381c25426c3fc634ba9a31eb04127ad39c85f7454bf157e17"))
println("modifierById: " + historyReaderOpt.get.modifierById(ModifierId @@ "96f68654faa2c26381c25426c3fc634ba9a31eb04127ad39c85f7454bf157e17"))
 
returns 
true
None

as 

```
override def modifierById(id: ModifierId): Option[ErgoPersistentModifier] =
    if (isSemanticallyValid(id) != ModifierSemanticValidity.Invalid) {
      historyStorage.modifierById(id)
    } else {
      None
    }
```

and the modifier is ModifierSemanticValidity.Invalid

However, this behavior is potentially more or less ok, the node is trying to store invalid modifier ids in order to avoid re-processing and possible related spam attacks. However, not clear how modifier became sematically invalid, at least we need to add logging and maybe also write reason to the database.

PR #1456, opened 2021-09-27
/script/p2sh API method got back, minor code refactoring 

kushti:
In this PR:
* P2SH address support back (script/p2sh)
* a bit of unused code removed


src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -101,6 +101,138 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
+                           outAssets: Map[ByteBuffer, Long],
+                           outAssetsNum: Int,
+                           boxesToSpend: IndexedSeq[ErgoBox],
+                           stateContext: ErgoStateContext): ValidationResult[Long] = {
+    // Cost limit per block
+    val maxCost = stateContext.currentParameters.maxBlockCost
+
+    ErgoBoxAssetExtractor.extractAssets(boxesToSpend) match {
+      case Success((inAssets, inAssetsNum)) =>
+        lazy val newAssetId = ByteArrayWrapper(inputs.head.boxId)
```
pragmaxim, line 208: Using `ByteBuffer.wrap` here would be consistent with the asset-cost PR of mine

pragmaxim: This is a bug, as the `ErgoBoxAssetExtractor.extractAssets` returns `ByteBuffer`, so you need to do `ByteBuffer.wrap` I think

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -101,6 +101,138 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
+                         (implicit verifier: ErgoInterpreter): ValidationResult[Long] = {
+
+    // Cost limit per block
+    val maxCost = stateContext.currentParameters.maxBlockCost
+
+    val input = inputs(inputIndex)
+
+    // Just in case, should always be true if client implementation is correct.
+    if(!box.id.sameElements(input.boxId)) {
+      log.error("Critical client error: box is not inputs(inputIndex)")
```
greenhat, line 134: I wonder what is the reason to "downgrade" this check from `validEquals` which fails the whole validation to `log.error`?

greenhat (approved):

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
aslesarenko: I don't think it is safe to change the validation logic. What is the motivation?

aslesarenko (commented):
Looks good, except one place.

ccellado (approved):

Issue #1410, opened 2021-07-20
Mainnet explorer link for testnet node


kushti:
"When running on testnet, and you click the address (to bring it up in explorer), it links you to the address in the mainnet explorer"

PR #1459, opened 2021-09-29
Add networkType to /info call

ApexTheory:
Regarding https://github.com/ergoplatform/ergo/issues/1410

I want to add a `networkType` lookup on the [ergo-node-interface](https://github.com/ergoplatform/ergo-node-interface/issues) project so that panel hyperlinks point to the correct explorer.  I didn't see any existing way to check the node's `networkType` via API, so I figured a good place to add it is in `GET /info`.

This change would be used in a subsequent PR in the [ergo-node-interface](https://github.com/ergoplatform/ergo-node-interface/issues) repository.


ApexTheory:
Hi @kushti , I have made changes on the node-interface project here https://github.com/ergoplatform/ergo-node-interface/pull/37.  If this change is approved, can we merge so that @deadit can review and test the frontend changes?

Thanks!

PR #1460, opened 2021-09-30
keep invalidated TXs in expiring bloomfilters

pragmaxim:


pragmaxim:
Done in https://github.com/ergoplatform/ergo/pull/1468

Issue #1461, opened 2021-10-01
Remember recently included transactions 


kushti:
It could be the case a transaction asked from a peer got included into a block and processed before delivery. For avoiding banning peers for sending invalid transactions, we need to store included transactions for last 2-3 blocks (into a Bloom filter?) and check on tx delivery that the tx is not included. I guess bitcoind is doing that.

pragmaxim:
I think I'd have to introduce a new Event that would transfer transactions from applied blocks from `ErgoNodeViewHolder` to `ErgoNodeViewSynchronizer` where it would become part of bloom filters .... I did not think of other place to keep the bloom filters in ...

pragmaxim:
@kushti I don't get the point of bloom filter here since the amount of transactions in a few last blocks is very low

PR #1464, opened 2021-10-04
ignore already applied txs from peers

pragmaxim:
Tackles https://github.com/ergoplatform/ergo/issues/1461

Solution : cache ids of transactions that were applied to blocks and then use it to filter out ids coming from peers.


pragmaxim:
@kushti I checked logs of a node that booted from scratch and was running for a while and no peer sent such a transaction  :shrug: However I did what you suggested, now we penalize peer for spammy transactions, but NOT for those that were spammy due to being already applied.

kushti:
> @kushti I checked logs of a node that booted from scratch and was running for a while and no peer sent such a transaction shrug However I did what you suggested, now we penalize peer for spammy transactions, but NOT for those that were spammy due to being already applied.

Check for messages like:

`Spam attempt: peer ConnectedPeer(ConnectionId(remote=/192.248.174.58:9030, local=/10.0.0.10:60016, direction=Outgoing)) has sent a non-requested modifiers of type 102 with ids: Set(d87656ccaaeef3eb0cec69462ade0d578229b4dae548a6799dd9adcc9d6b8bfe)`

pragmaxim:
I tried to test it but I did not actually see any `Got spammy transactions` messages, so I was not able to reproduce it.

Issue #1465, opened 2021-10-09
Rework DeliveryTracker internal structures with expiring Bloom filters


kushti:
Rework DeliveryTracker structures (at least *invalid*, which is a HashSet) to use expiring Bloom filters.

Issue #607, opened 2019-01-18
Use Bloom or Cuckoo filters for mempool blacklist


kushti:
Currently, blacklist in the mempool is organized as a dictionary of (transaction_id -> timestamp). It is more efficient to replace it with Bloom (or Cuckoo) filter, or construction with periodic cleaning(based on two filters?)

PR #1468, opened 2021-10-13
Delivery tracker with bloom filters

pragmaxim:
Tackles https://github.com/ergoplatform/ergo/issues/1465 and https://github.com/ergoplatform/ergo/issues/607

Size-limited FIFO collection of BloomFilters with a hash based time-expiring cache in front,
so that lower number of elements is tested for presence accurately and huge number of elements only approximately.
Bloom filters do not have means of element expiration, so a collection of BloomFilters that gradually expire is a viable alternative.
Ie. we expire whole bloom filters instead of expiring elements and we check all bloom filters for element presence

**Broken test is fixed directly in v4.0.15**


src/main/scala/org/ergoplatform/nodeView/mempool/ExpiringApproximateCache.scala
```diff
@@ -0,0 +1,126 @@
+package org.ergoplatform.nodeView.mempool
+
+import com.google.common.hash.{BloomFilter, Funnels}
+import java.nio.charset.Charset
+import scala.collection.immutable.TreeMap
+import scala.concurrent.duration.FiniteDuration
+
+sealed trait ApproximateCacheLike[T] {
```
kushti, line 8: Please do Scaladoc for the trait

src/main/resources/application.conf
```diff
@@ -53,8 +53,17 @@ ergo {
     # Maximum number of unconfirmed transactions node can accept
     mempoolCapacity = 1000
 
-    # Maximum number of invalidated transactions node can save to decline them immediately without validation
-    blacklistCapacity = 10000
+    # Maximum number of invalid modifiers in bloom filters
+    invalidModifiersBloomFilterCapacity = 10000000
+
+    # fraction of 1 that represents a rate of expiration as modifiers are coming, only values from open interval (0,1) are valid
```
kushti, line 59: Probably better description is needed, better with an example

src/main/scala/org/ergoplatform/nodeView/mempool/ExpiringApproximateCache.scala
```diff
@@ -0,0 +1,126 @@
...
+        case Some((_, headBf))
+            if headBf.approximateElementCount() < bloomFilterApproxElemCount =>
+          headBf.put(elem)
+          this
+        case Some((idx, _)) =>
+          val newIdx = idx + 1
+          val bf     = createNewFilter
+          bf.put(elem)
+          val newFilters =
+            if (bloomFilterQueue.size >= bloomFilterQueueSize)
```
kushti, line 72: Please use braces for if... else

PR #1483, opened 2021-10-26
Rework of processRemoteModifiers in ErgoNodeViewHolder

kushti:
In this PR, headers got from another PR applied to blockchain and its state bypassing cache if possible, and in efficient way


src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -254,6 +255,74 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
...
+          val sorted = mods.sortBy(_.asInstanceOf[Header].height)
+
+          val applied0 = if (sorted.head.asInstanceOf[Header].height == history().headersHeight + 1) {
+
+            // we apply sorted headers while headers sequence is not broken
+            val appliedBuffer = mutable.Buffer[Header]()
+            var expectedHeight = history().headersHeight + 1
+            var linkBroken = false
+
+            cfor(0)(_ < sorted.length, _ + 1) { idx =>
```
pragmaxim, line 288: can we use `fastFor` here ? 
```
import spire.syntax.fastFor._

// print numbers 1 through 10
fastFor(0)(_ < 10, _ + 1) { i =>
  println(i)
}
```

pragmaxim (approved):

Issue #1392, opened 2021-06-20
Sign a custom message [Feature request]


nitram147:
Back in the days (like in 2015), I've seen a "sign message" feature in the Bitcoin Core wallet implementation.

Implementing something like that could be beneficial for the ErgoWallet.

It could be used for example for signing a message like:
>  I'm John Doe [ID 123, living in ABC] and I own the keys for this address, so I'm also the owner of all NFTs belonging to this address.

So the NFT holder can prove that he own a key for address to which some NFT belongs to. This can be useful in data notarization, decentralized name service and many other possible applications.

Maybe it would be better to use ECDSA instead of Schnorr (more implementations available - so it would be easier to build software for signature verification). The same "k" parameter problem during signing different messages (thus leaking a private key) with ECDSA could be avoided for example with a process proposed in this RFC [https://datatracker.ietf.org/doc/html/rfc6979#section-3.2](https://datatracker.ietf.org/doc/html/rfc6979#section-3.2)

kushti:
Related issue #1174

kushti:
I think verification for Ergo signatures should be implemented in the first place, based on https://github.com/ScorexFoundation/sigmastate-interpreter/pull/685 .

WinterTFG:
from signastate.interpreter, are signMessage, and verifySignature enough?

Can possibly provide signatures and verifications to token messages via API:
GET /wallet/sign/{tokenName} - return signature of token message 
POST /wallet/verify/{tokenName} - verifies token signature provided by requestor (against message in token)
  {"signature": "..."}

Questions:
- Is it expected that a witness could request verification by the owner to decrypt a message to a hash for comparison?
- Do these methods need to have other nodes help with consensus?  Perhaps the sigma methods already account for this?

kushti:
> @kushti from signastate.interpreter, are signMessage, and verifySignature enough?
> 
> Can possibly provide signatures and verifications to token messages via API: GET /wallet/sign/{tokenName} - return signature of token message POST /wallet/verify/{tokenName} - verifies token signature provided by requestor (against message in token) {"signature": "..."}
> 
> Questions:
> 
>     * Is it expected that a witness could request verification by the owner to decrypt a message to a hash for comparison?
> 
>     * Do these methods need to have other nodes help with consensus?  Perhaps the sigma methods already account for this?

I guess enough, by why token name? Let the message be arbitrary and allow client to know its semantics

[...]

PR #1484, opened 2021-10-26
sign and verify added to api route

vikingphoenix:
ref: issue #1392 

The additions were tested against 4.0.15 working branch, but updated today against 4.0.16 after rebuilding from new fork, but 4.0.16 doesn't seem to build on it's own currently.  This is just to show you current progress.


vikingphoenix:
I still need help with this.  

reduceToCrypto seems to be derived from base class properly, but I'm having trouble calling from secretsProver (or even proverOpt.head).  

I also spent some time trying to just use the secretsProver.publicKeys, should this work?  I was able to call the API but when I verified it would not equate the hashes.

PR #1487, opened 2021-10-31
transactionhintsbag bug fix

sahandzou:
Fixing a bug in TransactionHintsBag case class. for the addHintsForInput method, it has a bug in the case that if index != 0.


kushti (approved):

kushti (approved):

kushti:
@sahandzou thanks for the contribution!

Issue #1018, opened 2019-12-29
Use SecretString for passwords and other secrets instead of String


aslesarenko:
Security practices should be followed to protect password storage during ErgoTool
operations. 
For example, password characters need to be stored in mutable buffers which are erased
as soon as possible so that they are not leaked to Java Garbage Collector and to the OS
when the running ErgoTool process terminates.

This requires refactoring in ergo-wallet code.
Required also for Appkit.

UPDATE:
Consider move [SecretString](https://github.com/aslesarenko/ergo-appkit/blob/d82048e477e09abdc9ecc6542ecdd2fd0bfbb2c6/common/src/main/java/org/ergoplatform/appkit/SecretString.java) class from Appkit to ergo-wallet and using it instead of String for passwords, mnemonic and other sensitive information.

PR #1497, opened 2021-11-14
#1018 add SecretString

ccellado:
https://github.com/ergoplatform/ergo/issues/1018

Added SecretString to mnemonic. 

Are there any other places to replace the string with?


ccellado:
Well, those errors doesn't look related here. Or are they?

kushti (approved):
LGTM, maybe you can consider replacing ```mnemonicPassOpt.fold[Option[SecretString]](None)(x => Option(SecretString.create(x)))``` with a shortcut

Issue #1173, opened 2020-08-18
Unify recursive deletion of folder 


kushti:
FileUtils in Ergo and FileUtils in Scorex testing scope do have in common a method for recursive deletion of folder contents. Move a unified method to scorex (main scope) and re-use it in Ergo.

kushti:
currently, we have 3 FileUtils:

*scorex.testkit.utils.FileUtils*
*org.ergoplatform.wallet.utils.FileUtils*
*org.ergoplatform.utils.FileUtils*

it would be good to merge them into single *org.ergoplatform.wallet.utils.FileUtils*

ApexTheory:
Starting on this issue.

PR #1499, opened 2021-11-16
Unify FileUtils

ApexTheory:
Close [1173](https://github.com/ergoplatform/ergo/issues/1173)


kushti (approved):
This code: 
```
  val basePath: Path = java.nio.file.Files.createTempDirectory(s"scorex-${System.nanoTime()}")

  sys.addShutdownHook {
    deleteRecursive(basePath.toFile)
  }
```

is not needed in Ergo, test-specific only. I think you need to create ```trait TestFileUtils extends FileUtils``` in src/test and move the test-specific code above to this trait (to be used around tests)

kushti (approved):

Issue #1504, opened 2021-11-17
Introduce maxTransactionCost limit for the mempool 


kushti:
Introduce max transaction cost setting in the node setting, and check it in the mempool transaction validation (peer which is sending costly transaction to be considered as spammy one)

pragmaxim:
Is too high tx complexity also a reason for considering peer as spammy one ?

PR #1506, opened 2021-11-18
max transaction cost in mempool

pragmaxim:
Addresses https://github.com/ergoplatform/ergo/issues/1504 ... peer which is sending costly transaction should be considered spammy one


src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -710,7 +711,8 @@ object CandidateGenerator extends ScorexLogging {
             stateWithTxs.validateWithCost(
               tx,
               Some(upcomingContext),
-              maxTransactionComplexity
+              maxTransactionComplexity,
+              maxTransactionCost
             ) match {
```
pragmaxim, line 716: I think that letting a transaction with cost over limit into a candidate is also wrong

kushti: The limit is for mempool, to protect from it spam, and prevent broadcasting too heavy transactions. For blocks, still there's another limit (cost per block) only, and miner can include transactions which exceed maxTransactionCost directly.

kushti:
@pragmaxim merged #1511 into this PR, please do new tests for too heavy transactions instead of ones removed (in #1511)

pragmaxim:
@kushti test changed from `complexity` -> `cost`

kushti:
@pragmaxim replaced tx cost limit in miner (CandidateGenerator) with block cost limit. Miner should have ability to include transactions not propagated directly into the blocks (as verification will be passed anyway). Maybe block cost limit is not very precise though ?

pragmaxim:
@kushti not sure I understand : 
>>> Maybe block cost limit is not very precise though ?

kushti:
> @kushti not sure I understand :
> 
> > > > Maybe block cost limit is not very precise though ?

I meant maybe better to have some gap, for security and also fee collecting tx

pragmaxim:
@kushti I'm a bit lost in this PR ... do you mean hard code it like this :  `maxBlockCost + someExtraReserveCost` ?

pragmaxim:
@kushti please check last 2 commits, I wanted to finish this PR finally

kushti:
> @kushti please check last 2 commits, bug-fix + extraFeeCost ... passing context is important as otherwise it is read from DB

cost in transaction assembly is computational cost (similarly to gas in Ethereum), unlikely it can be correlated to tx fee

pragmaxim:
> cost in transaction assembly is computational cost (similarly to gas in Ethereum), unlikely it can be correlated to tx fee

yeah I misinterpreted that, it is totally unrelated :facepalm:

src/main/resources/mainnet.conf
```diff
@@ -34,6 +34,9 @@ ergo {
 
     # List with hex-encoded identifiers of transactions banned from getting into memory pool
     blacklistedTransactions = []
+
+    # maximum cost of transaction for it to be propagated
+    maxTransactionCost = 5000000
```
aslesarenko, line 39: Looks too high.

src/main/resources/testnet.conf
```diff
@@ -8,10 +8,15 @@ ergo {
...
     headerChainDiff = 800
+
     # having more internal miner threads helps to reproduce real use cases, like 100 GPU miners
     internalMinersCount = 1
+
     # GPU miners are currently polling at 100ms interval
     internalMinerPollingInterval = 200ms
+
+    # maximum cost of transaction for it to be propagated
+    maxTransactionCost = 800000
```
aslesarenko, line 19: Why each conf file has its own value for this parameter?

src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
```diff
@@ -38,22 +39,23 @@ trait UtxoStateReader extends ErgoStateReader with TransactionValidation {
...
     val verifier = interpreterOpt.getOrElse(ErgoInterpreter(context.currentParameters))
 
     tx.statelessValidity().flatMap { _ =>
       val boxesToSpend = tx.inputs.flatMap(i => boxById(i.boxId))
-      val txComplexity = boxesToSpend.map(_.ergoTree.complexity).sum
-      if (txComplexity > complexityLimit) {
-        throw new Exception(s"Transaction $tx has too high complexity $txComplexity")
-      }
       tx.statefulValidity(
-        boxesToSpend,
```
aslesarenko, line 54: validateStateful internally uses MaxBlockCost as the limit. Now, when we have costLimit, it should be passed in validateStateful and take place of MaxBlockCost limit.

src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
```diff
@@ -38,22 +39,23 @@ trait UtxoStateReader extends ErgoStateReader with TransactionValidation {
...
-        tx.dataInputs.flatMap(i => boxById(i.boxId)),
-        context)(verifier)
+          boxesToSpend,
+          tx.dataInputs.flatMap(i => boxById(i.boxId)),
+          context)(verifier).flatMap {
+            case txCost if txCost > costLimit =>
```
aslesarenko, line 54: At the time we check this limit, the node may already have spent full MaxBlockCost of computations. If the ratio between MaxBlockCost and MaxTxCost become higher, then this is increasingly may be an attack vector. See my previous comment.

aslesarenko (changes requested):
Looks good overall, but see comments.

src/main/resources/mainnet.conf
kushti: Block limit is 7M, and 5M is non-discriminative to most known existing transactions while also limiting them.

src/main/resources/testnet.conf
kushti: Block limit is 1M in the testnet, 7M in the mainnet. Thus max tx cost values are different as well.

src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
kushti: see my reply below

src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
kushti: good catch, made a new version of *validateWithCost* which is passing accumulated cost considering the limit and tests, please check

@aslesarenko made 
```case Failure(mme: MalformedModifierError) if mme.message.contains("CostLimitException")``` as no structured information to extract CostLimitException available

File: src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala [hunk 6]
```diff
@@ -38,22 +40,30 @@ trait UtxoStateReader extends ErgoStateReader with TransactionValidation {
...
-        throw new Exception(s"Transaction $tx has too high complexity $txComplexity")
-      }
       tx.statefulValidity(
-        boxesToSpend,
-        tx.dataInputs.flatMap(i => boxById(i.boxId)),
-        context)(verifier)
+          boxesToSpend,
+          tx.dataInputs.flatMap(i => boxById(i.boxId)),
+          context,
+          accCost)(verifier).map(_ - accCost) match {
```
aslesarenko, line 58: The value passed to statefulValidity shouldn't be greater than MaxTxCost limit. 
accCost, as it is computed above, is not bound by MaxTxCost.
So, changes in validateStateful are also necessary.

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
kushti: Cost is bounded by MaxBlockCost, so we set accCost as there's only txCost room to reach it. Please check updated tests.

oskin1 (approved):

PR #1522, opened 2021-12-06
Using interpreter during candidate generation like in verification, more logging on tx verification

kushti:
In this PR:
* Interpreter is reused in candidate generation like in block verification
* Logging context and stacktrace when tx validation failed


src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -701,16 +701,18 @@ object CandidateGenerator extends ScorexLogging {
           if (!inputsNotSpent(tx, stateWithTxs) || doublespend(current, tx)) {
             //mark transaction as invalid if it tries to do double-spending or trying to spend outputs not present
             //do these checks before validating the scripts to save time
+            log.debug(s"Transaction ${tx.id} double-spending or spending non-existing inputs")
             loop(mempoolTxs.tail, acc, lastFeeTx, invalidTxs :+ tx.id)
           } else {
-            implicit val verifier: ErgoInterpreter = ErgoInterpreter(
+            val verifier: ErgoInterpreter = ErgoInterpreter(
```
aslesarenko, line 707: This verifier is created for each tx from mempoolTxs. 
In ErgoState.execTransactions it is created once for all transactions in the block.
It is impossible to do the same here (because we don't know which txs will be added to the block).
However, once the candidate is created (https://github.com/ergoplatform/ergo/blob/30c28eb66660eb0874cc10fd6ebadb00d4b62786/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala#L507) it can be verified again with ErgoState.execTransactions, i.e. with exactly the same procedure.

kushti: Moved verifier outside the loop. 

ErgoState.execTransactions is called as part of local block processing. Calling it for candidate is possible, but could slow down miners a bit.

aslesarenko: > Moved verifier outside the loop.

I'm not sure it will work well. If the number of transactions in the pool is large, then all of the script will accumulate inside IRContext's internal data structures. 
So in this case resetContext is necessary before each transaction, same as in ErgoState.execTransactions

kushti: resetContext is called for each transaction in validateStateful (which is in turn is called in statefulValidity which is called inside stateWithTxs.validateWithCost), pretty much as in ErgoState.execTransactions

aslesarenko: ah, ok, missed this.

greenhat (approved):

src/main/scala/scorex/core/transaction/state/StateFeature.scala
```diff
@@ -8,10 +8,6 @@ import scala.util.Try
 trait StateFeature
 
 trait TransactionValidation extends StateFeature {
-  def isValid(tx: ErgoTransaction): Boolean = validate(tx).isSuccess
-
-  def filterValid(txs: Seq[ErgoTransaction]): Seq[ErgoTransaction] = txs.filter(isValid)
-
   def validate(tx: ErgoTransaction): Try[Unit]
```
aslesarenko, line 11: Please write ScalaDocs for the class and the method.

kushti: Done in 4.0.17 branch

PR #1530, opened 2021-12-14
fixing negative number of modifiers to download while syncing

pragmaxim:
1. DeliveryTracker does not differentiate between modifier types so we cannot treat headers differently than blocks for instance ...  We need to fetch more headers than blocks in parallel as it is much faster... So far it worked only because while syncing, headers are fetched first, then blocks.
2. invalid modifiers bug fixed
3. improving logging, weird states would log warning now

### DIFF
--- a/src/main/scala/scorex/core/network/DeliveryTracker.scala
+++ b/src/main/scala/scorex/core/network/DeliveryTracker.scala
@@ -14,6 +14,8 @@ import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
 import scala.util.{Failure, Try}
+import DeliveryTracker._
+import org.ergoplatform.modifiers.history.header.Header
 
 /**
   * This class tracks modifier statuses.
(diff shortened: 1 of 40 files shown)


src/main/scala/scorex/core/network/DeliveryTracker.scala
```diff
@@ -223,17 +299,36 @@ object DeliveryTracker {
...
       deliveryTimeout,
       maxDeliveryChecks,
-      ExpiringApproximateCache.empty(bloomFilterCapacity, bloomFilterExpirationRate, frontCacheSize, frontCacheExpiration),
+      settings.cacheSettings.network,
+      settings.scorexSettings.network.desiredInvObjects,
       nvsRef
     )
   }
+
+  implicit class MapPimp[K, V](underlying: mutable.Map[K, V]) {
```
kushti, line 312: This class is done in ToDownloadProcessor already (with just adjust function there), please do a common implementation.

kushti (commented):
LGTM, but please address comments.

Re-targeted against 4.0.17

src/main/scala/scorex/core/network/DeliveryTracker.scala
pragmaxim: This is `mutable.Map` and `ToDownloadProcessor` has immutable ... I thought I would do that after rewriting `DeliveryTracker` in an immutable way.

kushti (approved):

PR #1533, opened 2021-12-15
modifier deduplication while syncing [WIP]

pragmaxim:
Revisiting syncing process, trying to find how come we fetch some blocks twice.

I know that Sets are not used in Ergo mostly due to performance reasons, but it is a good defense in a complex system against duplication ...


pragmaxim:
It was not the cause of problems, Sets are slow, let's believe all inputs are unique.

PR #1538, opened 2021-12-20
Improve SecretString

ccellado:
Added SecretString from ergo-appkit to some critical parts.

Should be a different release, as `erase()` method could introduce unexpected behaviour.


ccellado:
Added erased flag and throw exceptions after erasure

ccellado:
Strangely the tests failed in the nodeView code.

Local `sbt test` ran flawlessly with this pr

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
```diff
@@ -1,10 +1,15 @@
 package org.ergoplatform.wallet.interface4j;
 
+import sigmastate.serialization.SelectFieldSerializer;
+
 import java.util.Arrays;
 
 /**
+ * Modified copy from https://github.com/aslesarenko/ergo-appkit SecretString class
```
aslesarenko, line 8:
```suggestion
 * Modified copy from https://github.com/ergoplatform/ergo-appkit SecretString class
```

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
```diff
@@ -27,16 +38,23 @@
```
aslesarenko, line 38: All method, not only getData, must raise an exception. 
Imagine, instead of erase we did actual de-allocation and cleanup of the memory.
Calling isEmpty would mean segmentation fault.
We cannot de-allocate, but we can make all methods throw exception.

The only method which would be allowed on erased object is `isErased`, which can be added and be used
to detect the status of SecretString.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/SecretStorage.scala
```diff
@@ -16,6 +20,9 @@ trait SecretStorage {
 
   def lock(): Unit
 
+  /**
```
aslesarenko, line 23: This is not related to SecretString, but since this class is touched, please add ScalaDocs for all methods.
A lot of quesitons can be addressed in docs:
- should secretFile be open, and how
- what are the implications of isLocked = true
- what means secret = None
- how state should change on unlock and lock

aslesarenko (changes requested):

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/SecretStorage.scala
ccellado: Added docs, please review

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
```diff
@@ -13,40 +18,66 @@
...
+    }
 
     /**
      * Extracts secret characters as an array.
+     * Throws an exception if {@link SecretString#_erased} flag is true
      */
     public char[] getData() {
-        return _data;
+        if (_erased) {
+            throw new RuntimeException("SecretString already erased");
```
aslesarenko, line 51: Throwing the same exception from all methods should be extracted into a single method with some meaningful name.

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
aslesarenko: @ccellado What about adding `isErased` method?
I can imaging it can be useful in situation when you cannot statically know whether the secret was erased or not by the previous code.
So you have an option of either creating a new secret string, or just reusing existing, in the case it was not erased.

aslesarenko (changes requested):

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
ccellado: What's the point of extracting the throw? It muddies the code readability in my opinion.
If the reason is to be sure the text of the error thrown is the same, it could be extracted to a const.

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
ccellado: Forgot to add. Fixed!

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
aslesarenko: The point is that once error condition is detected, some _logical unit of work_ has to be done. 
Throwing is part of this unit of work, now happen to be the only part, but something else can be done as well, like collecting some data. 
You can also think of the SecretString to move kind of error state (yes you don't save this state now, but anyway).
It is better to have all these encapsulated.

The readability depends on the name you give the method.

aslesarenko (commented):
See comments

ergo-wallet/src/main/java/org/ergoplatform/wallet/interface4j/SecretString.java
ccellado: Added method, please check

aslesarenko (approved):

kushti (approved):

Issue #1322, opened 2021-03-29
Add "Burn Tokens" functionality to the API and panel


kushti:
Can be done via a new type of request, AssetBurnRequest (in addition to AssetIssueRequest and PaymentRequest)

ccellado:
I'll work on that

glasgowm148:
closed with #1540

PR #1540, opened 2021-12-23
Add BurnTokens API

ccellado:
#1322


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
```diff
@@ -268,8 +268,18 @@ trait ErgoWalletSupport extends ScorexLogging {
...
     require(inputBoxes.nonEmpty, "There must be at least one input box")
 
+    //filter burnTokens requests
+    val requestsWithoutBurnTokens = requests.filterNot(_.isInstanceOf[BurnTokensRequest])
+    val burnTokens = {
+      val burnRequests = requests.flatMap {
+        case req: BurnTokensRequest => req.assetsToBurn
+        case _ => None
+      }
+      if (burnRequests.isEmpty) Map.empty[ModifierId, Long] else TransactionBuilder.collTokensToMap(burnRequests.toColl)
```
pragmaxim, line 278: @ccellado this would be much cleaner with `val (requestsWithBurnTokens, requestsWithoutBurnTokens) = requests.partition(_.isInstanceOf[BurnTokensRequest])`

kushti (changes requested):
Please add assets burn request description to openapi.yaml

see

```
            anyOf:
              - $ref: '#/components/schemas/PaymentRequest'
              - $ref: '#/components/schemas/AssetIssueRequest'
```

there

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
pragmaxim: Just `val (requestsWithBurnTokens, requestsWithoutBurnTokens)` and then : 
```
requestsWithBurnTokens.map(_.asInstanceOf[BurnTokensRequest])
```
I suggested this as your version is filtering the `requests` twice. OCD

ccellado: Now this makes sense. Fixed it, thanks

kushti (changes requested):
It seems request encoder/decoder are missed along with their usage, see ```TransactionRequestEncoder``` and *AssetIssueRequest should be serialized to json* and *PaymentRequest should be serialized to json* tests in ```JsonSerializationSpec```

kushti (approved):

Issue #1524, opened 2021-12-07
Make tests and optimizations for ErgoState.execTransactions


kushti:
* make tests for ErgoState.execTransactions in ErgoStateSpecification
* rework ErgoState.execTransactions to avoid .toList -> .toIndexedSeq transformations and *lazy val boxes* construction, to reduce memory footprint

PR #1549, opened 2021-12-29
ErgoState.execTransactions perf improvements and tests

pragmaxim:
Solves https://github.com/ergoplatform/ergo/issues/1524 

Getting rid of too many immutable collections copying. Also I replaced Cats `sequence` with tail recursive function.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
@@ -26,8 +26,9 @@ import sigmastate.basics.DLogProtocol.ProveDlog
 import sigmastate.serialization.ValueSerializer
 
 import scala.annotation.tailrec
+import scala.collection.immutable.VectorBuilder
 import scala.collection.mutable
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 /**
   * Implementation of minimal state concept in Scorex. Minimal state (or just state from now) is some data structure
@@ -76,39 +77,31 @@ object ErgoState extends ScorexLogging {
 
   /**
     * Tries to validate and execute transactions.
-    *
+    * @param transactions to be validated and executed
+    * @param currentStateContext to be used for tx execution
+    * @param checkBoxExistence function to provide ErgoBox by BoxId
     * @return Result of transactions execution with total cost inside
     */
   def execTransactions(transactions: Seq[ErgoTransaction],
                        currentStateContext: ErgoStateContext)
                       (checkBoxExistence: ErgoBox.BoxId => Try[ErgoBox]): ValidationResult[Long] = {
-    import cats.implicits._
     val verifier: ErgoInterpreter = ErgoInterpreter(currentStateContext.currentParameters)
 
     @tailrec
-    def execTx(txs: List[ErgoTransaction], accCostTry: ValidationResult[Long]): ValidationResult[Long] = (txs, accCostTry) match {
-      case (tx :: tail, r: Valid[Long]) =>
-        val boxesToSpendTry: Try[List[ErgoBox]] = tx.inputs.toList
-          .map(in => checkBoxExistence(in.boxId))
-          .sequence
-
-        lazy val dataBoxesTry: Try[List[ErgoBox]] = tx.dataInputs.toList
-          .map(in => checkBoxExistence(in.boxId))
-          .sequence
-
-        lazy val boxes: Try[(List[ErgoBox], List[ErgoBox])] = dataBoxesTry.flatMap(db => boxesToSpendTry.map(bs => (db, bs)))
-
-        val vs = tx.validateStateless()
-          .validateNoFailure(txBoxesToSpend, boxesToSpendTry)
-          .validateNoFailure(txDataBoxes, dataBoxesTry)
-          .payload[Long](r.value)
-          .validateTry(boxes, e => ModifierValidator.fatal("Missed data boxes", e)) { case (_, (dataBoxes, toSpend)) =>
-            tx.validateStateful(toSpend.toIndexedSeq, dataBoxes.toIndexedSeq, currentStateContext, r.value)(verifier).result
-          }
-
-        execTx(tail, vs)
-      case _ =>
-        accCostTry
+    def collectBoxesById(
+                 remainingBoxIds: Iterator[ErgoBox.BoxId],
+                 resultingBoxes: Try[VectorBuilder[ErgoBox]] = Success(new VectorBuilder[ErgoBox])
+               ): Try[IndexedSeq[ErgoBox]] = {
+      if (!remainingBoxIds.hasNext) {
+        resultingBoxes.map(_.result())
+      } else {
+        checkBoxExistence(remainingBoxIds.next()) match {
+          case Success(box) =>
+            collectBoxesById(remainingBoxIds, resultingBoxes.map(_ += box))
+          case Failure(ex) =>
+            Failure(ex)
+        }
+      }
     }
 
     // Skip v1 block transactions validation if corresponding setting is on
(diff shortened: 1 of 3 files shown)


src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -76,37 +76,50 @@ object ErgoState extends ScorexLogging {
...
           .validateNoFailure(txBoxesToSpend, boxesToSpendTry)
           .validateNoFailure(txDataBoxes, dataBoxesTry)
           .payload[Long](r.value)
           .validateTry(boxes, e => ModifierValidator.fatal("Missed data boxes", e)) { case (_, (dataBoxes, toSpend)) =>
-            tx.validateStateful(toSpend.toIndexedSeq, dataBoxes.toIndexedSeq, currentStateContext, r.value)(verifier).result
+            tx.validateStateful(toSpend, dataBoxes, currentStateContext, r.value)(verifier).result
           }
 
-        execTx(tail, vs)
+        execTx(txs.tail, vs)
```
aslesarenko, line 122: The `tail` is only fast when txs is a List, which is not good idea in a first place.

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -76,37 +76,50 @@ object ErgoState extends ScorexLogging {
...
+            collectBoxesById(remainingBoxIds, resultingBoxes.map(_ += box))
+          case Failure(ex) =>
+            Failure(ex)
+        }
+      }
+    }
 
-        lazy val boxes: Try[(List[ErgoBox], List[ErgoBox])] = dataBoxesTry.flatMap(db => boxesToSpendTry.map(bs => (db, bs)))
+    @tailrec
+    def execTx(allTxs: Seq[ErgoTransaction], accCostTry: ValidationResult[Long]): ValidationResult[Long] = (allTxs, accCostTry) match {
```
aslesarenko, line 107: Tuple allocation `(allTxs, accCostTry)` for every allTxs

src/test/scala/org/ergoplatform/nodeView/state/ErgoStateSpecification.scala
```diff
@@ -141,4 +143,36 @@ class ErgoStateSpecification extends ErgoPropertyTest {
...
+
+    // tx validation should kick in and detect block height violation
+    val invalidTx = invalidErgoTransactionGen.sample.get
+    assert(
+      ErgoState.execTransactions(txs :+ invalidTx, stateContext)(id => Try(boxes.getOrElse(ByteArrayWrapper(id), invalidTx.outputs.head)))
+        .errors.head.message.startsWith("Transaction outputs should have creationHeight not exceeding block height.")
+    )
+
+    // no transactions are valid
+    assert(ErgoState.execTransactions(Seq.empty, stateContext)(id => Try(boxes(ByteArrayWrapper(id)))).isValid)
```
aslesarenko, line 176: Please add tests and test vectors for cost computation.

aslesarenko (changes requested):
I suggest to first add tests and test vectors for costs in a separate PR.
Then make optimizations on top of that tests and make sure they don't break them.

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -76,37 +76,50 @@ object ErgoState extends ScorexLogging {
...
+        resultingBoxes.map(_.result())
+      } else {
+        checkBoxExistence(remainingBoxIds.next()) match {
+          case Success(box) =>
+            collectBoxesById(remainingBoxIds, resultingBoxes.map(_ += box))
+          case Failure(ex) =>
+            Failure(ex)
+        }
+      }
+    }
```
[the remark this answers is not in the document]
pragmaxim: It does not loop through after `Failure` as the recursive function is not called in case of first `Failure`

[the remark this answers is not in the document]
pragmaxim: `Try` allocation is necessary I believe as obtaining the `Box` from UtxoState can simply fail or it might not be there ...

src/test/scala/org/ergoplatform/nodeView/state/ErgoStateSpecification.scala
pragmaxim: Ok, I wanted to but the costs are all randomly generated in the helper methods, I have to figure out how to do it so it is not mess

pragmaxim:
I addressed all concerns except the `TailRec` for collecting boxes, `Try` is kind of inevitable here, the collection is `Vector`, so while loop was very ugly and it does not have any added value here.

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -76,47 +77,55 @@ object ErgoState extends ScorexLogging {
...
+    def collectBoxesById(
+                 remainingBoxIds: Iterator[ErgoBox.BoxId],
+                 resultingBoxes: Try[VectorBuilder[ErgoBox]] = Success(new VectorBuilder[ErgoBox])
+               ): Try[IndexedSeq[ErgoBox]] = {
+      if (!remainingBoxIds.hasNext) {
+        resultingBoxes.map(_.result())
+      } else {
+        checkBoxExistence(remainingBoxIds.next()) match {
+          case Success(box) =>
+            collectBoxesById(remainingBoxIds, resultingBoxes.map(_ += box))
```
aslesarenko, line 100: ArrayBuilder will be much faster than VectorBuilder, especially in this use case, where you can use arrBuilder.sizeHint(tx.inputs.length) to do single allocations. Then appending with += will be super fast.
Just compare what happens on += in VectorBuilder and ArrayBuilder.

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -76,47 +77,55 @@ object ErgoState extends ScorexLogging {
...
           currentStateContext.ergoSettings.nodeSettings.skipV1TransactionsValidation) {
       Valid(0L)
     } else {
-      execTx(transactions.toList, Valid[Long](0L))
+      import spire.syntax.all.cfor
+      var costResult: ValidationResult[Long] = Valid[Long](0L)
+      cfor(0)(_ < transactions.length && costResult.isValid, _ + 1) { i =>
+        val validCostResult = costResult.asInstanceOf[Valid[Long]]
+        val tx = transactions(i)
+        val boxesToSpendTry: Try[IndexedSeq[ErgoBox]] = collectBoxesById(tx.inputs.iterator.map(_.boxId))
```
aslesarenko, line 117: Here you can take advantage of knowing val n = tx.inputs.length and save O(log(n)) allocations in builder by using sizeHint method.

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
```diff
@@ -76,47 +77,55 @@ object ErgoState extends ScorexLogging {
...
+        lazy val dataBoxesTry: Try[IndexedSeq[ErgoBox]] = collectBoxesById(tx.dataInputs.iterator.map(_.boxId))
+        lazy val boxes: Try[(IndexedSeq[ErgoBox], IndexedSeq[ErgoBox])] = dataBoxesTry.flatMap(db => boxesToSpendTry.map(bs => (db, bs)))
+        costResult = tx.validateStateless()
+          .validateNoFailure(txBoxesToSpend, boxesToSpendTry)
+          .validateNoFailure(txDataBoxes, dataBoxesTry)
+          .payload[Long](validCostResult.value)
+          .validateTry(boxes, e => ModifierValidator.fatal("Missed data boxes", e)) { case (_, (dataBoxes, toSpend)) =>
+            tx.validateStateful(toSpend, dataBoxes, currentStateContext, validCostResult.value)(verifier).result
```
aslesarenko, line 125: The currentStateContext is not update after each tx, is it ok?

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
pragmaxim: yes but `tx.inputs` and tx.dataInputs` have different types so `.map(_.boxId)` here was used for readability ...

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
pragmaxim: it is, just a few lines above `costResult = tx.validateStateless()`

src/main/scala/org/ergoplatform/nodeView/state/ErgoState.scala
pragmaxim: Maybe we should introduce new interface for `Input` and `DataInput` with `dev boxId: BoxId` ? We could then with any input that has `boxId` ... WYT ?

aslesarenko: collectBoxesById implicitly creates an empty builder here.  However, since we know the size, we can use sizeHint of the builder, to pre-allocate space.

pragmaxim: but `ArrayBuilder` does not take size hints, one can either `new mutable.ArrayBuilder.ofRef[ErgoBox]()` or `mutable.ArrayBuilder.make()`

pragmaxim: I see, strange that they did not make it part of the constructor

aslesarenko (approved):

kushti (approved):

PR #1561, opened 2022-01-04
Networking layer optimizations for 4.0.20

kushti:
* minModifiersPerBucket & maxModifiersPerBucket values increased for better bandwidth utilization 
* maxDeliveryChecks decreased for the mainnet (to ignore nodes failing to deliver quicker)
* less INFO output
* if node is storing all the full blocks (blocksToKeep = -1), downloading to be started immediately now, not after headers synchronization
* `.get` removed in ErgoNodeViewHolder on `vault().rollback(idToVersion(progressInfo.branchPoint.get)).get`
* wallet scan to be started only when chain is about to sync
* peer eviction happens now only if there are 5 connections at least

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -414,7 +414,7 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
                   vault().rollback(idToVersion(progressInfo.branchPoint.get)).get
                 } else vault()
 
-                if(newHistory.fullBlockHeight == newHistory.headersHeight) {
+                if((newHistory.headersHeight - newHistory.fullBlockHeight) < 20) {
                   blocksApplied.foreach(newVault.scanPersistent)
                 }
 
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -409,16 +409,18 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
                 val newMemPool = updateMemPool(progressInfo.toRemove, blocksApplied, memoryPool(), newMinState)
 
                 @SuppressWarnings(Array("org.wartremover.warts.OptionPartial"))
+                val v = vault()
                 val newVault = if (progressInfo.chainSwitchingNeeded) {
-                  val v = vault()
                   v.rollback(idToVersion(progressInfo.branchPoint.get)) match {
                     case Success(nv) => nv
                     case Failure(e) => log.warn("Wallet rollback failed: ", e); v
                   }
                 } else {
-                  vault()
+                  v
                 }
 
+                // we assume that wallet scan may be started if fullblocks-chain is no more
+                // than 20 blocks behind headers-chain
                 val almostSyncedGap = 20
 
                 if((newHistory.headersHeight - newHistory.fullBlockHeight) < almostSyncedGap) {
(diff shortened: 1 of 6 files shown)


src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -408,12 +408,22 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
...
+                    case Failure(e) => log.warn("Wallet rollback failed: ", e); v
+                  }
+                } else {
+                  vault()
+                }
+
+                val almostSyncedGap = 20
+
+                if((newHistory.headersHeight - newHistory.fullBlockHeight) < almostSyncedGap) {
+                  blocksApplied.foreach(newVault.scanPersistent)
```
pragmaxim, line 425: but does not it mean that the previous blocks will be missing in Vault? In comparison to previous way

greenhat (approved):

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -241,10 +241,12 @@ class NetworkController(settings: NetworkSettings,
     * It is needed to prevent eclipsing (https://www.usenix.org/system/files/conference/usenixsecurity15/sec15-paper-heilman.pdf)
     */
   private def scheduleEvictRandomConnections(): Unit = {
+   val evictionThreshold = 5
```
knizhnik, line 244: Should it be moved to settings?

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -408,12 +408,22 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
...
+                  val v = vault()
+                  v.rollback(idToVersion(progressInfo.branchPoint.get)) match {
+                    case Success(nv) => nv
+                    case Failure(e) => log.warn("Wallet rollback failed: ", e); v
+                  }
+                } else {
+                  vault()
+                }
+
+                val almostSyncedGap = 20
```
knizhnik, line 422: Should not this be made a configuration parameter?

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -408,12 +408,22 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
               case Success(newMinState) =>
                 val newMemPool = updateMemPool(progressInfo.toRemove, blocksApplied, memoryPool(), newMinState)
 
-                //we consider that vault always able to perform a rollback needed
                 @SuppressWarnings(Array("org.wartremover.warts.OptionPartial"))
                 val newVault = if (progressInfo.chainSwitchingNeeded) {
-                  vault().rollback(idToVersion(progressInfo.branchPoint.get)).get
-                } else vault()
-                blocksApplied.foreach(newVault.scanPersistent)
+                  val v = vault()
```
knizhnik, line 413: This assignment can be moved before newVault to avoid vault() call in else branch.

knizhnik (approved):

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
kushti: Done

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
kushti: Hard question, it makes sense if someone will need to change it. IDK if this is the case

src/main/scala/scorex/core/network/NetworkController.scala
kushti: Any need to change it ?

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
kushti: No, the wallet will automatically scan previous blocks by sending ScanInThePast messages to self

PR #1563, opened 2022-01-08
avoid using global LaunchParameters throughout the codebase

pragmaxim:
There is a lot of Randomness in tests which causes non-deterministic failures, especially with generators that run a test many times. When we have complete control over `parameters` then we can change it or override it on a single place and troubleshoot all tests together, ie. avoid fixing one and breaking others.

With this change, `parameters` are propagated as an instance wherever it is needed, instead of just calling `LaunchParameters` global object from everywhere.

I tried to change the randomly generated numbers that are used to generate boxes, transactions and blocks, but after day of trying, not being able to make ALL test pass with it, I figured the only way is to increase some limits like `maxBlockCost` or `minValuePerByte` just on a two places.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletStorage.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletStorage.scala
@@ -3,14 +3,14 @@ package org.ergoplatform.nodeView.wallet.persistence
 import com.google.common.primitives.{Ints, Shorts}
 import org.ergoplatform.nodeView.state.{ErgoStateContext, ErgoStateContextSerializer}
 import org.ergoplatform.nodeView.wallet.scanning.{Scan, ScanRequest, ScanSerializer}
-import org.ergoplatform.settings.{Constants, ErgoSettings}
+import org.ergoplatform.settings.{Constants, ErgoSettings, Parameters}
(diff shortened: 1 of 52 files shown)


src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletStorage.scala
```diff
@@ -106,10 +106,10 @@ final class WalletStorage(store: LDBKVStore, settings: ErgoSettings)
     * Read state context from the database
     * @return state context read
     */
-  def readStateContext: ErgoStateContext = store
+  def readStateContext(parameters: Parameters): ErgoStateContext = store
```
kushti, line 109: Likely it will be cleaner to serialize parameters (and we have serializer already in "object ParametersSerializer"), and during deserialization check if parameters there, and fallback to LaunchParameters if so (so old versions of the node storage will be supported).

kushti (approved):
Please check ScalaDoc, fix arguments where not updated

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletStorage.scala
pragmaxim: Frankly I don't understand that as we are passing `LaunchParameters` from `ErgoApp`. It is all `LaunchParameters`. Do you mean trying to serialize `parameters` at startup in `ErgoApp` to check if they are valid?

PR #1570, opened 2022-01-11
Candidate for 4.0.20.2: Memory pool improvements

kushti:
This PR covering following issues: 

1) broadcasting is not randomized, always broadcasting the same transactions set. also it is not checking whether inputs of transactions are in utxo set
2) transaction got invalidated if no inputs in the mempool or utxo set. but then due to (1) chained transactions can come out-of-order and be invalidated

In this PR, transaction with no inputs presented in the utxo set or pool only declined, not invalidated.

Also, only transactions with inputs in the utxo set broadcasted. And broadcast now randomized

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
@@ -52,7 +52,7 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool, private[mempool] val sta
       scala.util.Random.nextInt(max)
     }
 
-    cfor(start)(_ < start + limit, _ + 1) { idx =>
+    cfor(start)(_ < Math.min(start + limit, total), _ + 1) { idx =>
       val tx = txSeq.apply(idx)
       result += tx
     }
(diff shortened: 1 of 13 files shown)


src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -39,6 +41,25 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool, private[mempool] val sta
 
   override def take(limit: Int): Iterable[ErgoTransaction] = pool.orderedTransactions.values.take(limit)
 
+  def random(limit: Int): Iterable[ErgoTransaction] = {
+    val result = mutable.ArrayBuffer[ErgoTransaction]()
+    val txSeq = pool.orderedTransactions.values.toSeq
```
pragmaxim, line 46: `TreeMap(1 -> 1).values.toSeq` produces `Stream` ... this is better `TreeMap(1 -> 1).values.to[Vector]`

kushti: done in https://github.com/ergoplatform/ergo/pull/1570/commits/d7f185b2dee59d75b4cf0abd79a346afabf50276

oskin1 (approved):

PR #1573, opened 2022-01-13
1568 check node for being stuck

pragmaxim:
@kushti I can do a second iteration in case we need more info to be printed

`ErgoNodeViewSynchronizer` schedules `HealthCheck` to `ErgoNodeViewHolder` which checks if chain is stuck, if so, returns why to `ErgoNodeViewSynchronizer` which then prints it all out together with state of `DeliveryTracker` and `SyncTracker`


src/main/resources/application.conf
```diff
@@ -165,6 +165,9 @@ ergo {
     # Base16 representation of genesis state roothash
     genesisStateDigestHex = "cb63aa99a3060f341781d8662b58bf18b9ad258db4fe88d09f8f71cb668cad4502"
 
+    # Difference between NOW and timestamp of the latest chain update. This helps to discover syncing issues.
+    acceptableChainUpdateDelay = 5m
```
pragmaxim, line 169: Should this be part of chain settings or node settings?

kushti: Node settings, as this setting is node-specific (can be different on different nodes)

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -612,12 +626,54 @@ object ErgoNodeViewHolder {
...
+
+    if (chainUpdateDelayed || blockUpdateDelayed) {
+      val bestFullBlockOpt =
+        history.bestFullBlockOpt
+          .filter(_.id != lastMod.id)
+          .fold("")(fb => s"\n best full block: $fb")
+      val repairNeeded = ErgoHistory.repairIfNeeded(history)
+      ChainIsStuck(s"Chain not modified for $chainUpdateDelay ms, headers-height: $headersHeight, " +
+        s"block-height $blockHeight, chain synced: $chainSynced, repair needed: $repairNeeded, " +
+        s"last modifier applied: $lastMod $bestFullBlockOpt")
```
pragmaxim, line 672: I thought it would be useful to print out last persistent modifier applied to history, possibly bestFullBlock if they are not the same modifiers

kushti (approved):

PR #1575, opened 2022-01-14
fix: invalid yaml in openapi spec

ross-weir:


src/main/resources/api/openapi.yaml
```diff
@@ -1718,10 +1716,12 @@ components:
             - $ref: '#/components/schemas/Timestamp'
         headersScore:
           type: integer
+          format: int64
```
kushti, line 1719: could be out of 64 bits, so better to stay on format being skipped I guess

src/main/resources/api/openapi.yaml
```diff
@@ -1718,10 +1716,12 @@ components:
...
           description: Can be 'null' if no headers is applied since node launch
           nullable: true
         fullBlocksScore:
           type: integer
+          format: int64
```
kushti, line 1724: could be out of 64 as well

src/main/resources/api/openapi.yaml
```diff
@@ -1701,7 +1699,7 @@ components:
           example: 327
         difficulty:
           type: integer
-          format: int32
+          format: int64
```
kushti, line 1702: better to skip format, can be out of 64 bits

kushti (changes requested):

src/main/resources/api/openapi.yaml
ross-weir: Removed & added a note to description

kushti (approved):

PR #1576, opened 2022-01-14
Candidate for version 4.0.22

kushti:
* #1540
* #1569
* #1579
* #1584
* #1590
* #1593
* #1594

Contributors: @aslesarenko , @ccellado , @deadit , @kushti,  @pragmaxim , @ross-weir


aslesarenko (approved):

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -198,12 +198,12 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
 
-    if(globalDiff > 250) {
+    if(globalDiff > 100) {
       globalSyncGot = newGlobal
 
       val diff = syncTracker.updateLastSyncGetTime(remote)
-      if (diff > 1000 ) {
-        // process sync if sent in more than 1 second after previous sync
+      if (diff > 600 ) {
+        // process sync if sent in more than 600 ms after previous sync
```
aslesarenko, line 206: I'm a bit concerned with this constants. I suspect there is a lot of hidden wisdom (and rounds of tunning) to come up with these values. It is better to document the reasons motivation about why they are changed.
It will be really helpful to read them in 1-2 years.

aslesarenko (approved):

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
kushti: Good point, done in https://github.com/ergoplatform/ergo/pull/1576/commits/675d0df5f2ec6ac04ccb880841481e9a701b1ba6

PR #1579, opened 2022-01-16
Avoid using reduceToCrypto in ScriptApiRoute

aslesarenko:
The new implementation uses passed ErgoTree and interpret it the same way as TX validation and candidate assembly code.

In particulare, caching and optimized evaluation is performed.


src/test/scala/org/ergoplatform/http/routes/ScriptApiRouteSpec.scala
```diff
@@ -56,7 +56,7 @@ class ScriptApiRouteSpec extends AnyFlatSpec
       val cost = json.hcursor.downField("cost").as[Int].right.get
       value shouldEqual -45
       condition shouldEqual true
-      cost shouldEqual 40
+      cost shouldEqual 50
```
aslesarenko, line 59: This new cost is correct, because this is how the tree is actually validated as part of transaction.

kushti (approved):

PR #1580, opened 2022-01-17
fix: increase akka max uri length

ross-weir:
Large ergo trees are unable to be converted to addresses using the default uri limit of 2k via `utils/ergoTreeToAddress`.


kushti:
@ross-weir thanks for contributions! Please feel free to reach me over Ergo Discord https://discord.gg/kj7s7nb to get bounties

Issue #1577, opened 2022-01-14
SyncTracker should unite Statuses and lastSyncSentTime


pragmaxim:
See https://github.com/ergoplatform/ergo/blob/0af9dd9d8846d672c1e2a77f8ab29963fa5acd1e/src/main/scala/org/ergoplatform/network/ErgoSyncTracker.scala#L64

It is hard to reason about and possibly source of bugs as we have 2 separate Lists of peers that are mutated. Let's have a single collection.

PR #1583, opened 2022-01-18
1577 unifying collections in sync tracker

pragmaxim:
Solves https://github.com/ergoplatform/ergo/issues/1577


src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -295,7 +295,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
         log.debug(s"$remote has equal header-chain")
     }
 
-    if ((oldStatus != status) || syncTracker.isOutdated(remote) || status == Older || status == Fork) {
+    if ((oldStatus != status) || syncTracker.notSyncedOrOutdated(remote) || status == Older || status == Fork) {
```
pragmaxim, line 298: `isOutdated` returned `true` in these cases (missing, notSynced, outdated), so I renamed the method and changed the implementation so it is more obvious.

src/main/scala/org/ergoplatform/network/ErgoSyncTracker.scala
```diff
@@ -22,34 +21,37 @@ final case class ErgoSyncTracker(system: ActorSystem,
...
+      peerOpt
+        .flatMap(_.lastSyncSentTime)
+        .exists(syncTime => (timeProvider.time() - syncTime).millis > SyncThreshold)
+    notSyncedOrMissing || outdated
   }
 
   def updateStatus(peer: ConnectedPeer, status: HistoryComparisonResult, height: Option[Height]): Unit = {
     val seniorsBefore = numOfSeniors()
-    statuses += peer -> ErgoPeerStatus(peer, status, height.getOrElse(ErgoHistory.EmptyHistoryHeight))
+    statuses += peer -> ErgoPeerStatus(peer, status, height.getOrElse(ErgoHistory.EmptyHistoryHeight), None, None)
```
kushti, line 54: Why None, None here for sync update time ? Sync times being erased on e.g. height update then

pragmaxim: Yes, you are right. fixed

kushti (approved):

PR #1586, opened 2022-01-20
Candidate for 4.0.21.1: fix for UTXO set transformation proofs causing p2p problems

kushti:
It was considered in p2p layer that any block section is no more than some limit, namely, 2MB. While this is true for transactions (where limit can be changed by miners though and so exceed 2MB) and extension section (10 KB max), there is no check for UTXO set transformations proofs (most of the nodes generating and not downloading). 

It was the case that one transaction made 5K UTXOs https://explorer.ergoplatform.com/en/transactions/4268e2bf73cb9b8e7e36ed04da8fe6759dd723f164b3ca2ff43c3e406c9c6995 . Then a light client (working in "digest" mode) caused problems for public nodes storing UTXO set, as they did not considered violation of the rule and so stuck in endless loop.

This release relaxing the issue by making full nodes not stuck if they some have a block section more than allowed block section size. In this case the node is still sending it. On client side, there is some relaxation of the limit. However, to protect light clients fully we need to introduce consensus-level checks for UTXO set transformation proofs in a next soft-fork, or miners can vote on raising cost of an output.  

Also, too much of syncing traffic appeared in the logs. This release tries to cool down syncing.

### DIFF
--- a/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
+++ b/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
@@ -145,7 +145,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     val syncV2Version = Version(4, 0, 16)
     remote.peerInfo.exists(_.peerSpec.protocolVersion >= syncV2Version)
   }
-
+  
   /**
     * Send synchronization statuses to neighbour peers
     *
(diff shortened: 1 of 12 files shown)


src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -219,12 +239,20 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
       case Younger | Fork =>
-        // send extension (up to 400 header ids) to a peer which chain is less developed or forked
-        val ext = hr.continuationIds(syncInfo, size = 400)
-        if (ext.isEmpty) log.warn("Extension is empty while comparison is younger")
-        log.debug(s"Sending extension of length ${ext.length}")
-        log.debug(s"Extension ids: ${idsToString(ext)}")
-        sendExtension(remote, ext)
+        val newExtSend = timeProvider.time()
+
+        if(newExtSend - globalExtSend > 250) {
```
oskin1, line 244: Why `250`? Please make a constant instead of a magic number.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -277,12 +305,20 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
       case Younger =>
-        // send extension (up to 400 header ids) to a peer which chain is less developed or forked
-        val ext = hr.continuationIds(syncInfo, size = 400)
-        if (ext.isEmpty) log.warn("Extension is empty while comparison is younger")
-        log.debug(s"Sending extension of length ${ext.length}")
-        log.debug(s"Extension ids: ${idsToString(ext)}")
-        sendExtension(remote, ext)
+        val newExtSend = timeProvider.time()
+
+        if (newExtSend - globalExtSend > 250) {
```
oskin1, line 310: Same here.

pragmaxim (approved):

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -189,16 +190,35 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
     * Process sync message `syncInfo` got from neighbour peer `remote`
     */
   protected def processSync(hr: ErgoHistory, syncInfo: ErgoSyncInfo, remote: ConnectedPeer): Unit = {
-    syncInfo match {
-      case syncV1: ErgoSyncInfoV1 => processSyncV1(hr, syncV1, remote)
-      case syncV2: ErgoSyncInfoV2 => processSyncV2(hr, syncV2, remote)
+    val newGlobal = timeProvider.time()
+    val globalDiff = newGlobal - globalSyncGot
+
+    if(globalDiff > 250) {
```
oskin1, line 201: Why 250? Please make a constant instead of a magic number.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -189,16 +190,35 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+      globalSyncGot = newGlobal
+
+      val diff = syncTracker.updateLastSyncGetTime(remote)
+      if (diff > 1000 ) {
```
oskin1, line 205: Why 1000?

Issue #1582, opened 2022-01-18
Make DeliveryTracker REST endpoint


pragmaxim:
Similar to : 
https://github.com/ergoplatform/ergo/blob/0af9dd9d8846d672c1e2a77f8ab29963fa5acd1e/src/main/scala/org/ergoplatform/http/api/ErgoPeersApiRoute.scala#L39

PR #1589, opened 2022-01-20
1582 delivery tracker rest api endpoint

pragmaxim:
Solves https://github.com/ergoplatform/ergo/issues/1582


kushti (changes requested):
Please add new API method description to openapi.yaml

pragmaxim:
@kushti I'm thinking whether to change those `.getOrElse("N/A")` to simply "not required" fields

kushti (approved):

PR #1597, opened 2022-01-27
Minor memory optimizations

kushti:
In this PR:

* StateChangeOperation hierarchy removed, along with conversions from StateChangeOperation descendants into tree ops, now tree ops created directly
* Seq() replaced in many places 
* Seq(args) replaced in some places with Array(args)
* ErgoMempool.extract fee now taking transaction outputs directly instead of calling ErgoState.boxChanges(Array(tx))._2, the latter option is much heavier (for a tx with many outputs) 

also:
* unused ByteArrayWrapper.fromLong removed
* extension field removed from Input and DataInput schemas in openapi.yaml (as decoder is not actually expecting it)


pragmaxim (approved):

greenhat (approved):

avldb/src/main/scala/scorex/crypto/authds/avltree/batch/VersionedLDBAVLStorage.scala
```diff
@@ -73,7 +73,7 @@ class VersionedLDBAVLStorage[D <: Digest](store: LDBVersionedStore,
         case _: ProverLeaf[D] => Seq(pair)
       }
     } else {
-      Seq()
```
knizhnik, line 76: Is it mandatory to use Seq.empty here? In some places we are using Nil instead of Seq(), in some places - Seq.empty

src/main/resources/api/openapi.yaml
```diff
@@ -31,12 +31,6 @@ components:
           $ref: '#/components/schemas/TransactionBoxId'
         spendingProof:
           $ref: '#/components/schemas/SpendingProof'
-        extension:
```
knizhnik, line 34: Is it really related to this PR?

PR #1601, opened 2022-01-28
Limit transaction size

knizhnik:
refer #1595


src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -446,14 +446,20 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
     modifiers.flatMap { case (id, bytes) =>
-      serializer.parseBytesTry(bytes) match {
-        case Success(mod) if id == mod.id =>
-          Some(mod)
-        case _ =>
-          // Penalize peer and do nothing - it will be switched to correct state on CheckDelivery
-          penalizeMisbehavingPeer(remote)
-          log.warn(s"Failed to parse modifier with declared id ${encoder.encodeId(id)} from ${remote.toString}")
-          None
+      if (bytes.size > settings.nodeSettings.maxTransactionSize) {
```
kushti, line 449: here you filter out not just transactions, but block sections (such as block transactions orutxo set transformations proofs), for which it is normal to be above the threshold.

Proper filtering would be inside 

```
    Constants.modifierSerializers.get(typeId) match {
      case Some(serializer: ScorexSerializer[ErgoTransaction]@unchecked) if typeId == Transaction.ModifierTypeId =>
        // parse all transactions and send them to node view holder
        val parsed: Iterable[ErgoTransaction] = parseModifiers(requestedModifiers, serializer, remote)
```

kushti (changes requested):

kushti (approved):

PR #1611, opened 2022-02-02
requiredDifficultyAfter optimizations

kushti:
In this PR, difficulty recalculation done only when readjustment happens (1st block of an epoch).  Otherwise, parent's difficulty used. Allows to save thousands of headers DB fetches per header which is not starting an epoch (so for 1023 out of 1024 headers in an epoch)

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
@@ -267,18 +267,25 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
       // Set difficulty for version 2 activation height (where specific difficulty is needed due to PoW change)
       settings.chainSettings.initialDifficultyVersion2
     } else {
-      //todo: it is slow to read thousands headers from database for each header
-      //todo; consider caching here
-      //todo: https://github.com/ergoplatform/ergo/issues/872
       val parentHeight = parent.height
-      val heights = difficultyCalculator.previousHeadersRequiredForRecalculation(parentHeight + 1)
-        .ensuring(_.last == parentHeight)
-      if (heights.lengthCompare(1) == 0) {
-        difficultyCalculator.calculate(Seq(parent))
+
+      // Difficulty readjustment happens on the first block of an epoch (height % epochLength == 1)
+      if(parentHeight % settings.chainSettings.epochLength == 0) {
+        //todo: it is slow to read thousands headers from database for each header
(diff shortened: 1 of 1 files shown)


src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -267,18 +267,25 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
...
+      if(parentHeight % settings.chainSettings.epochLength == 0) {
+        //todo: it is slow to read thousands headers from database for each header
+        //todo; consider caching here
+        //todo: https://github.com/ergoplatform/ergo/issues/872
+        val heights = difficultyCalculator.previousHeadersRequiredForRecalculation(parentHeight + 1)
+          .ensuring(_.last == parentHeight)
+        if (heights.lengthCompare(1) == 0) {
+          difficultyCalculator.calculate(Seq(parent))
+        } else {
+          val chain = headerChainBack(heights.max - heights.min + 1, parent, _ => false)
```
pragmaxim, line 282: hard to understand why the `limit` here is `heights.max - heights.min + 1`

kushti: This is the old code, and I guess for best headers-chain, we need to avoid headerChainBack at all, just read headers at needed heights

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -267,18 +267,25 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
...
+
+      // Difficulty readjustment happens on the first block of an epoch (height % epochLength == 1)
+      if(parentHeight % settings.chainSettings.epochLength == 0) {
+        //todo: it is slow to read thousands headers from database for each header
+        //todo; consider caching here
+        //todo: https://github.com/ergoplatform/ergo/issues/872
+        val heights = difficultyCalculator.previousHeadersRequiredForRecalculation(parentHeight + 1)
+          .ensuring(_.last == parentHeight)
+        if (heights.lengthCompare(1) == 0) {
+          difficultyCalculator.calculate(Seq(parent))
```
knizhnik, line 280: Is Seq->Array replacement appreciate  here?

kushti: Fixed, in #1618

PR #1613, opened 2022-02-02
Regenerate UTXO set transformation proofs if proofs hash is different from announced one 

kushti:
In this PR: 
* checks for UTXO set transformation proofs (ADProofs) checks reordered, now proof (digest) is checked first, then proof bytes
* if proof bytes are different from announced ones, re-check is done. Please check new comment in the code.


src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
```diff
@@ -77,15 +77,13 @@ class UtxoState(override val persistentProver: PersistentBatchAVLProver[Digest32
...
-        val mods = ErgoState.stateChanges(transactions).operations
-        val resultTry = Traverse[List].sequence(mods.map(persistentProver.performOneOperation).toList).map(_ => ())
-        ModifierValidator(stateContext.validationSettings)
-          .validateNoFailure(fbOperationFailed, resultTry)
-          .validateEquals(fbDigestIncorrect, expectedDigest, persistentProver.digest)
-          .result
-          .toTry
-      }
+      val mods = ErgoState.stateChanges(transactions).operations
+      val resultTry = Traverse[List].sequence(mods.map(persistentProver.performOneOperation).toList).map(_ => ())
```
aslesarenko, line 81: Why it is now allowed to perform operations in persistentProver concurrently?

src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
```diff
@@ -101,37 +99,61 @@ class UtxoState(override val persistentProver: PersistentBatchAVLProver[Digest32
...
+                * We are trying to generate proof again now.
+                */
+
+              persistentProver.rollback(inRoot)
+                .ensuring(java.util.Arrays.equals(persistentProver.digest, inRoot))
+
+              val mods = ErgoState.stateChanges(fb.blockTransactions.txs).operations
+              mods.foreach(persistentProver.performOneOperation)
+
+              proofBytes = persistentProver.generateProofAndUpdateStorage(meta)
```
aslesarenko, line 144: Shouldn't meta be also recomputed again?

src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
kushti: The only place `applyTransactions` is called from (aside of tests) is `applyModifier` function, where external lock on `persistentProver` already applied

src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
kushti: No, as it is not dependent on proofs hash.

aslesarenko (approved):

PR #1616, opened 2022-02-02
Sending extension to a peer when fork is detected 

kushti:
In this PR, extension sent to sync protocol V2 peers being forked (so on the same height but with different tip). It allows to resolve forks more quickly as the node is sending continuation eimmediately in case of fork detected when analyzing sync message.


src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -327,15 +327,20 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
         log.debug(s"Extension ids: ${idsToString(ext)}")
         sendExtension(remote, ext)
 
       case Fork =>
-        log.info(s"Fork detected with peer $remote, its sync message $syncInfo")
+        // send extension (up to 400 header ids) to a peer which chain is forked
+        val ext = hr.continuationIds(syncInfo, size = 400)
+        if (ext.isEmpty) log.warn("Extension is empty while comparison is fork")
+        log.debug(s"Sending extension of length ${ext.length} to forked peer $remote")
+        log.debug(s"Extension ids: ${idsToString(ext)}")
```
pragmaxim, line 342: Shouldn't it be `INFO` level since this should not happen that often, right?

pragmaxim (approved):

aslesarenko (approved):
@kushti Please add explanation to the PR on why this makes fork resolution faster.

kushti:
> @kushti Please add explanation to the PR on why this makes fork resolution faster.

done

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
kushti: "Extension ids: " message is quite long

PR #1625, opened 2022-02-08
Add extra check to `NipopowProof::hasValidConnections`

kettlebell:
Found this issue when simulating chains for `sigma-rust` nipopow code.


gtklocker (approved):
Makes sense. @kushti do we explicitly not include a level 0 on the interlink?

kushti:
> Makes sense. @kushti do we explicitly not include a level 0 on the interlink?

@gtklocker yes, seems so, from ```updateInterlinks``` function

kushti (approved):

Issue #1627, opened 2022-02-08
Incorrect BIP32 key derivation for some mnemonic phrases


greenhat:
### How to reproduce

Using `race relax argue hair sorry riot there spirit ready fetch food hedgehog hybrid mobile pretty` mnemonic and derive a secret key for `m/44'/429'/0'/0/0` path and then get the address(public key) from this derived secret key.
Expected (according to BIP32): 
`9eYMpbGgBf42bCcnB2nG3wQdqPzpCCw5eB1YaWUUen9uCaW3wwm` 
tested in Yoroi, Nautilus, SAFEW wallets, [bitcoinjs/bip32](https://github.com/bitcoinjs/bip32), sigma-rust
Got: 
`9ewv8sxJ1jfr6j3WUSbGPMTVx3TZgcJKdnjKCbJWhiJp5U62uhP` 
in Ergo node wallet, Ergo official mobile wallets

### Cause
For the above mnemonic hardened intermediate(`m/44'/429'`) child private key `childKey` BigInt value serializes to 31 bytes:

https://github.com/ergoplatform/ergo/blob/60bcda536bace5231ddc100e44097f059af80a72/ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/ExtendedSecretKey.scala#L73

which is then used for the next path:
https://github.com/ergoplatform/ergo/blob/60bcda536bace5231ddc100e44097f059af80a72/ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/ExtendedSecretKey.scala#L61

instead of required 32 bytes according to (BIP32)[https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--private-child-key]. See `ser256(kpar)` used for hardened child key in the first bullet point:

> If so (hardened child): let I = HMAC-SHA512(Key = c<sub>par</sub>, Data = 0x00 || ser<sub>256</sub>(k<sub>par</sub>) || ser<sub>32</sub>(i)).

where `ser256` should be:

> ser<sub>256</sub>(p): serializes the integer p as a 32-byte sequence, most significant byte first.

### Fix
Use previous(incorrect) derivation for existing wallets (JSON key file loading), fixed key derivation for new wallets, and explicitly ask for key derivation type to be used for the wallets restored from mnemonic.

### What does it mean for your wallet?
Your mnemonic phrase will still produce the same wallet if restored in the same or compatible app (see "How to reproduce" section).
[...]

PR #1628, opened 2022-02-08
add test case for incorrect BIP32 private child key derivation;

greenhat:
See description in #1627


greenhat:
Close in favor of #1636

Lu3kyplutusjoymoney (approved):

PR #1638, opened 2022-02-14
i1620 reset delivery tracker

pragmaxim:
To my understanding, it is not necessary to restart the actor. As `CheckModifiersToDownload` is repeatedly executed but returns No Modifiers with current `DeliveryTracker` state, we just need to `reset` the state and try to re-download missing modifiers.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -578,10 +578,10 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
 
   protected def handleHealthCheck: Receive = {
     case IsChainHealthy =>
-      log.debug(s"Check that chain is healthy, progress is $chainProgress")
+      log.info(s"Check that chain is healthy, progress is $chainProgress")
       val healthCheckReply = chainProgress.map { progress =>
         ErgoNodeViewHolder.checkChainIsHealthy(progress, history(), timeProvider, settings)
-      }.getOrElse(ChainIsHealthy)
+      }.getOrElse(ChainIsStuck("Node already stuck when started"))
       sender() ! healthCheckReply
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -578,10 +578,10 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
 
   protected def handleHealthCheck: Receive = {
     case IsChainHealthy =>
-      log.debug(s"Check that chain is healthy, progress is $chainProgress")
+      log.info(s"Check that chain is healthy, progress is $chainProgress")
       val healthCheckReply = chainProgress.map { progress =>
         ErgoNodeViewHolder.checkChainIsHealthy(progress, history(), timeProvider, settings)
-      }.getOrElse(ChainIsHealthy)
+      }.getOrElse(ChainIsStuck("Node already stuck when started"))
```
pragmaxim, line 584: After `acceptableChainUpdateDelay` the `chainProgress` must be Some, otherwise we are already stuck at the very beginning.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -653,24 +653,15 @@ object ErgoNodeViewHolder {
...
       val bestFullBlockOpt =
         history.bestFullBlockOpt
           .filter(_.id != lastMod.id)
           .fold("")(fb => s"\n best full block: $fb")
-      val repairNeeded = if(blockUpdateDelayed) {
-        ErgoHistory.repairIfNeeded(history)
-      } else {
-        false
-      }
+      val repairNeeded = ErgoHistory.repairIfNeeded(history)
```
pragmaxim, line 664: RepairIfNeeded can be called unconditionally as there is a strict condition inside for reparing.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -653,24 +653,15 @@ object ErgoNodeViewHolder {
     val chainUpdateDelay = timeProvider.time() - lastUpdate
     val acceptableChainUpdateDelay = settings.nodeSettings.acceptableChainUpdateDelay
     def chainUpdateDelayed = chainUpdateDelay > acceptableChainUpdateDelay.toMillis
-    def blockUpdateDelayed =
-      history.bestFullBlockOpt
-        .map(b => timeProvider.time() - b.header.timestamp)
-        .exists(blockUpdateDelay => blockUpdateDelay > acceptableChainUpdateDelay.toMillis)
-
     def chainSynced =
```
pragmaxim, line 656: We can rely on `ChainProgress` only, there is no need for checking `bestFullBlock` as it only complicates things, we'd have to differentiate between "Are we synced" and "Are we still syncing"

pragmaxim:
@kushti This is now fully and properly tested ... while working on DigestMode stuck, I went through all the cases that can happen and it work as expected.

kushti (approved):

PR #1643, opened 2022-02-21
Optional checkpointing

kushti:
This PR is generalizing optional checkpointing introduced after the hard-fork. In this PR, it is possible to set arbitrary checkpoint. Scripts in blocks before the checkpoint will not be validated, correctness of UTXO set is checked anyway. Thus bootstrapping can be significantly improved with no security issues.


src/main/resources/mainnet.conf
```diff
@@ -30,7 +30,10 @@ ergo {
     # Skip validation of transactions in the mainnet before block 417,792 (in v1 blocks).
```
greenhat, line 30:
```suggestion
    # Skip validation of transactions in the mainnet before the given block (height)
```

src/main/resources/mainnet.conf
```diff
@@ -30,7 +30,10 @@ ergo {
...
     # Block 417,792 is checkpointed by the protocol (so its UTXO set as well).
```
greenhat, line 31:
```suggestion
    # The given block is checkpointed by the protocol (so its UTXO set as well).
```

src/main/scala/org/ergoplatform/settings/ValidationRules.scala
```diff
@@ -140,7 +140,7 @@ object ValidationRules {
     hdrVotesUnknown -> RuleStatus(s => fatal(s"First header of an epoch should not contain a vote for unknown parameter. $s"),
       Seq(classOf[Header]),
       mayBeDisabled = true),
-    hdrCheckpointV2 -> RuleStatus(s => fatal(s"First version 2 header on the mainnet on height 417,729 should have " +
+    hdrCheckpoint -> RuleStatus(s => fatal(s"First version 2 header on the mainnet on height 417,729 should have " +
```
greenhat, line 143: Should the hard-coded height be changed to the one set in the config file?

greenhat (changes requested):

src/main/resources/application.conf
```diff
@@ -82,10 +82,7 @@ ergo {
     # default value is 100 blocks ~= 200 minutes
     headerChainDiff = 100
 
-    # Skip validation of transactions in the mainnet before block 417,792 (in v1 blocks).
-    # Block 417,792 is checkpointed by the protocol (so its UTXO set as well).
-    # The node still applying transactions to UTXO set and so checks UTXO set digests for each block.
-    skipV1TransactionsValidation = false
+    checkpoint = null
```
aslesarenko, line 85: Please add a description.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -331,19 +331,21 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
...
     /**
       * Helper method to check v2 checkpoint (first v2 block)
       */
-    private def checkpointV2Condition(header: Header): Boolean = {
-      if (header.height == 417792 && settings.networkType.isMainNet) {
-        header.id == "0ba60a7db44877aade553beb05200f7d67b586945418d733e455840d283e0508"
-      } else {
-        true
-      }
+    private def checkpointCondition(header: Header): Boolean = {
```
aslesarenko, line 341: Please add comment that the method returns true only when Id at the checkpoint height is not as expected.

src/main/resources/mainnet.conf
kushti: Changed comment in another way

src/main/resources/mainnet.conf
kushti: Changed comment in another way

src/main/scala/org/ergoplatform/settings/ValidationRules.scala
kushti: Changed comment

src/main/resources/application.conf
kushti: Done

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
kushti: done

greenhat (approved):

aslesarenko (approved):

Issue #1644, opened 2022-02-22
Wallet: DefaultBoxSelector ignoring needed boxes when needed for change


MrStahlfelge:
Steps to reproduce:
Call `DefaultBoxSelector.select(inputBoxes, amountToSpend, targetAssets)` with

* inputBoxes: 
 box 1 with amount 1001000000 nanoerg and a token
 box 2 with amount 1000000 nanoerg
* amoundToSpend: 1001000000 (exactly matching box 1)
* targetAssets: empty

Exptected behaviour:
returning both inputboxes (as both are needed for amountToSpent and change box for the token not to be sent)

Actual behaviour:
throwing `NotEnoughCoinsForChangeError`

Related test case: ergoplatform/ergo-appkit#141

kushti:
Fixed in #1645

PR #1645, opened 2022-02-22
Fix for #1644: DefaultBoxSelector should collect needed inputs when needed for change in presence of assets

kushti:
This PR is fixing impossibility to collect inputs  in presence of assets (#1644)

From the issue : 
"Steps to reproduce:
Call DefaultBoxSelector.select(inputBoxes, amountToSpend, targetAssets) with

    inputBoxes:
    box 1 with amount 1001000000 nanoerg and a token
    box 2 with amount 1000000 nanoerg
    amoundToSpend: 1001000000 (exactly matching box 1)
    targetAssets: empty

Exptected behaviour:
returning both inputboxes (as both are needed for amountToSpent and change box for the token not to be sent)

Actual behaviour:
throwing NotEnoughCoinsForChangeError"
 
 This PR contains test for situation described (failed initially), as well as a fix (and updated in old tests, as sometimes box selector now works differently)

### DIFF
--- a/ergo-wallet/src/main/scala/org/ergoplatform/wallet/boxes/DefaultBoxSelector.scala
+++ b/ergo-wallet/src/main/scala/org/ergoplatform/wallet/boxes/DefaultBoxSelector.scala
@@ -40,10 +40,20 @@ object DefaultBoxSelector extends BoxSelector {
       res += unspentBox
     }
 
-    def balanceMet = currentBalance >= targetBalance
+    def balanceMet: Boolean = {
+      val diff = currentBalance - targetBalance
+      val diffThreshold = if (currentAssets.isEmpty) {
+        0
+      } else {
+        MinBoxValue * (currentAssets.size / MaxAssetsPerBox + 1)
+      }
+      diff >= diffThreshold
+    }
 
-    def assetsMet = targetAssets.forall {
-      case (id, targetAmt) => currentAssets.getOrElse(id, 0L) >= targetAmt
+    def assetsMet: Boolean = {
+      targetAssets.forall {
+        case (id, targetAmt) => currentAssets.getOrElse(id, 0L) >= targetAmt
+      }
     }
 
     @tailrec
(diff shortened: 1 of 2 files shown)


pragmaxim (approved):

ergo-wallet/src/test/scala/org/ergoplatform/wallet/boxes/DefaultBoxSelectorSpec.scala
```diff
@@ -218,4 +215,28 @@ class DefaultBoxSelectorSpec extends AnyPropSpec with Matchers with EitherValues
...
+    val ergValue = 10 * MinBoxValue
+
+    val box1 = testBox(ergValue, TrueLeaf, StartHeight, tokenData)
+    val uBox1 = TrackedBox(parentTx, 0, Some(100), box1, Set(PaymentsScanId))
+    val box2 = testBox(ergValue, TrueLeaf, StartHeight)
+    val uBox2 = TrackedBox(parentTx, 0, Some(100), box2, Set(PaymentsScanId))
+
+    val s1 = select(Iterator(uBox1, uBox2), noFilter, ergValue, Map.empty)
+    s1 shouldBe 'right
+    s1.right.get.changeBoxes.size shouldBe 1
```
aslesarenko, line 232: Suggest to also check that there is the token in the change box.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/boxes/DefaultBoxSelector.scala
```diff
@@ -40,10 +40,30 @@ object DefaultBoxSelector extends BoxSelector {
...
 
-    def balanceMet = currentBalance >= targetBalance
+    /**
+      * Helper functions which checks whether enough ERGs collected
+      */
+    def balanceMet: Boolean = {
+      val diff = currentBalance - targetBalance
+
+      // We estimate how many ERG needed for assets in change boxes
+      val assetsDiff = currentAssets.size - targetAssets.size
```
MrStahlfelge, line 50: This estimation is not accurate in general. It works for the given test case, but this slightly changed one is still failing:

* inputBoxes:
box 1 with amount 1001000000 nanoerg and token A of value 2
box 2 with amount 1000000 nanoerg
* amoundToSpend: 1001000000 (exactly matching box 1)
* targetAssets: token A of value 1

MrStahlfelge (changes requested):

ergo-wallet/src/test/scala/org/ergoplatform/wallet/boxes/DefaultBoxSelectorSpec.scala
kushti: ok, done

PR #1646, opened 2022-02-23
Auto-burn tokens

ccellado:
Give user the opportunity to choose a set auto-burning of tokens, whitelisting some, or ignoring that feature at all


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
```diff
@@ -95,6 +95,31 @@ trait ErgoWalletSupport extends ScorexLogging {
...
+  private def mergeBurnWhitelistTokens(state: ErgoWalletState,
+                                       inputBoxes: Seq[TrackedBox],
+                                       burnTokensRequestMap: TokensMap): TokensMap = {
+    val input = inputBoxes.flatMap(_.tokens)
+    state.walletVars.settings.walletSettings.tokensWhitelist match {
+      case Some(x: Seq[String]) if x.isEmpty =>
+        AssetUtils.mergeAssets(
+          TransactionBuilder.collTokensToMap(
+            input
+              .map(tMap => (IdUtils.decodedTokenId(tMap._1), tMap._2)).toColl
```
pragmaxim, line 108: why not `{ case (id, amt) =>` ?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
```diff
@@ -95,6 +95,31 @@ trait ErgoWalletSupport extends ScorexLogging {
...
+          ),
+          burnTokensRequestMap)
+      case Some(x: Seq[String]) => AssetUtils.mergeAssets(
+        TransactionBuilder.collTokensToMap(
+          input
+            .filterNot(tMap => x.contains(tMap._1))
+            .map(tMap => (IdUtils.decodedTokenId(tMap._1), tMap._2)).toColl
```
pragmaxim, line 115: why not `{ case (id, amt) =>` ?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletSupport.scala
ccellado: Thanks! That's better

kushti (approved):

Issue #1648, opened 2022-02-26
Transactions with estimated execution cost bigger than 1000000 are rejected since 4.0.23


Luivatra:
A transaction that can be submitted and executed successfully on 4.0.22 fails with "Failed to sign boxes due to Estimated execution cost 1036435 exceeds the limit 1000000" on 4.0.23

This seems to ignore the maxTransactionCost parameter in the config file.

Luivatra:
Also hitting the same issue on 4.0.22 on a slightly larger transaction. Not really sure what the exact issue is.

pragmaxim:
Problem is that we use the default `MaxBlockCost = 1 000 000` in transaction/send : 
```
epochLength = 1024
MaxBlockCostIncrease -> 16 * 1024 
which basically means 696 160 * 16 = 11 138 560 
so the current MaxBlockCost should be 11 138 560 
```

kushti:
> Problem is that we use the default `MaxBlockCost = 1 000 000` in transaction/send :
> 
> ```
> epochLength = 1024
> MaxBlockCostIncrease -> 16 * 1024 
> which basically means 696 160 * 16 = 11 138 560 
> so the current MaxBlockCost should be 11 138 560 
> ```

Not every epoch cost increased, only if miners are voting for

PR #1649, opened 2022-02-28
Fix for #1648

kushti:


pragmaxim (approved):
Looks good, I'm glad I finally understood this part

Issue #1632, opened 2022-02-10
Bootstrapping in "digest" mode with large suffix often stucks on the way


kushti:
For  a config like:

```
ergo {
  directory = ${ergo.directory}"/.ergo"
  node {
    stateType = "digest"
    blocksToKeep = 100000
    mining = false
  }
}
```

sync stucks on the way (during full block application, headers synced), reportedly and also on my machine as well.

pragmaxim:
@kushti I could use more info as I'm new to this. It is not about Nipopow suffix but `suffix: IndexedSeq[ErgoPersistentModifier]` right?

kushti:
> @kushti I could use more info as I'm new to this. It is not about Nipopow suffix but `suffix: IndexedSeq[ErgoPersistentModifier]` right?

Nipopow suffix is different beast. In digest mode, all the headers still downloaded, but instead of building and holding UTXO set proofs of correctness of its transformations (ADProofs) downloaded and checked. Then it is possible to perform full validation for only suffix of the blockchain.

pragmaxim:
Ahh, so when `blocksToKeep` suffix is too large, like `100000` ...ok thanks

pragmaxim:
I reproduced it, stuck at `667613` ... looks like syncing problem.

pragmaxim:
The block is different from others by many more modifiers to apply, see `going to apply 933 modifiers` : 
```
23:54:23.138 INFO  [tor.default-dispatcher-11] o.e.n.h.ErgoHistory$$anon$3 - Full block ada64036803188230c4fbec5d6e0b3e9c2b848dbe09f69ec14bd7f24361021ff appended, going to apply 933 modifiers. New best block is 0983dacbff14c3beb4bb2da242220a3e03e8be40c2260c4b366edfab4bf17f4b with height 667613 updates block 9b17e2a20f911760dc95384a2705f09e51f7a7e8762fb9abaf1070e6a94009c9 with height 666680                                 
```
@kushti Can you see anything suspicious about that block? https://explorer.ergoplatform.com/en/blocks/0983dacbff14c3beb4bb2da242220a3e03e8be40c2260c4b366edfab4bf17f4b

kushti:
[...]

PR #1650, opened 2022-03-02
i1632 Fixing digest mode stuck

pragmaxim:
Should fix https://github.com/ergoplatform/ergo/issues/1632 when adopted by the network ...

@kushti see the second commit, we forgot to `4 *` the message size limit on Sender side ... so that those messages has `msgCount=0` which on Receiver side looks like 0 modifiers ... The logic behind `val count = if (size <= maxMessageSize) c + 1 else c` is weird (it probably exists to silently disregard this message, but that is quite dangerous). We should clearly define what to do when we hit TooLarge message during Serialization and Deserialization ...

### DIFF
--- a/src/main/scala/scorex/core/network/message/BasicMessagesRepo.scala
+++ b/src/main/scala/scorex/core/network/message/BasicMessagesRepo.scala
@@ -130,6 +130,8 @@ class ModifiersSpec(maxMessageSize: Int) extends MessageSpecV1[ModifiersData] wi
 
   import ModifiersSpec._
 
+  private val maxMsgSizeWithReserve = maxMessageSize * 4 // due to big ADProofs
+
   override val messageCode: MessageCode = MessageCode
   override val messageName: String = MessageName
 
@@ -143,7 +145,7 @@ class ModifiersSpec(maxMessageSize: Int) extends MessageSpecV1[ModifiersData] wi
 
     val (msgCount, msgSize) = modifiers.foldLeft((0, HeaderLength)) { case ((c, s), (_, modifier)) =>
       val size = s + NodeViewModifier.ModifierIdSize + 4 + modifier.length
-      val count = if (size <= maxMessageSize) c + 1 else c
+      val count = if (size <= maxMsgSizeWithReserve) c + 1 else c
       count -> size
     }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/scorex/core/network/message/BasicMessagesRepo.scala
```diff
@@ -143,7 +145,7 @@ class ModifiersSpec(maxMessageSize: Int) extends MessageSpecV1[ModifiersData] wi
 
     val (msgCount, msgSize) = modifiers.foldLeft((0, HeaderLength)) { case ((c, s), (_, modifier)) =>
       val size = s + NodeViewModifier.ModifierIdSize + 4 + modifier.length
-      val count = if (size <= maxMessageSize) c + 1 else c
+      val count = if (size <= maxMsgSizeWithReserve) c + 1 else c
```
pragmaxim, line 148: I don't like this much, we mess with modifier count information instead of Failing Fast here ... I mean if the load is too big, shouldn't we fail?

src/main/scala/scorex/core/network/message/BasicMessagesRepo.scala
```diff
@@ -156,21 +158,22 @@ class ModifiersSpec(maxMessageSize: Int) extends MessageSpecV1[ModifiersData] wi
...
-      log.warn(s"Message with modifiers ${modifiers.keySet} has size $msgSize exceeding limit $maxMessageSize.")
+    if (msgSize > maxMsgSizeWithReserve) {
+      log.warn(s"Message with modifiers ${modifiers.keySet} has size $msgSize exceeding limit $maxMsgSizeWithReserve.")
     }
   }
 
   override def parse(r: Reader): ModifiersData = {
     val typeId = ModifierTypeId @@ r.getByte() // 1 byte
     val count = r.getUInt().toIntExact // 8 bytes
+    require(count > 0, s"Illegal message with 0 modifiers of type $typeId")
```
pragmaxim, line 169: The same goes here, we get the Bytes but as `count` was 0 due to exceeding the maxSize, we ignore it ... so it is better to Fail Fast imo.

pragmaxim:
@kushti I assume that the `broken modifiers in 4.0.17 or 4.0.18` are not concerning headers but only block modifiers.

kushti (approved):

src/main/scala/scorex/core/network/message/BasicMessagesRepo.scala
kushti: Agree, at least we need to log. Failing on serialization should be investigated for possible side effects.

PR #1655, opened 2022-03-18
tidy openapi spec

ross-weir:
- Remove inline responses/requests to make spec more friendly to codegen/docgen
- Add servers field
- Fix couple of small issues in schema defs

### DIFF
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -12,6 +12,12 @@ info:
     name: CC0 1.0 Universal
     url: https://raw.githubusercontent.com/ergoplatform/ergo/master/LICENSE
 
+servers:
+  - url: http://127.0.0.1:9053
+    description: Local full node
+  - url: http://213.239.193.208:9053
+    description: Known public node
+
 components:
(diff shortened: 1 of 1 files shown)


src/main/resources/api/openapi.yaml
```diff
@@ -397,6 +428,13 @@ components:
           items:
             description: Pair of TokenId and Long
             type: array
+            items:
+              description: TokenId or Long
+              anyOf:
+                - description: TokenId
+                  type: string
```
ross-weir, line 435: Not sure if the types in this pair are right

kushti (approved):

pragmaxim:
@ross-weir What was the reasoning behind adding the `servers` field ? It causes problems like this https://github.com/ergoplatform/ergo/pull/1671 or when people are running node on a different IP address.

ross-weir:
> @ross-weir What was the reasoning behind adding the `servers` field ? It causes problems like this #1671 or when people are running node on a different IP address.
> 
> See #1672

Tools like `openapi-generator` complain if it's missing. Making a PR to fix

Issue #1652, opened 2022-03-08
Help decentralization: Add an information which peers are open to get requests on HTTP API


MrStahlfelge:
### Background

Light clients like the ergo wallet app need to connect to node's REST API in order to get box information or submit transactions.

### Problem

Ergo is a peer2peer network and nodes can be offline or malfunctioning. This shouldn't be a problem, a client could connect to another node and continue working. However, a client don't know which peers of the network it could connect to. Even when a connection to a perfectly working node is established, clients can't fetch information of other nodes to connect to. There is an existing Peers API that reports all peers the node is connected to. This information is very valuable to help decentralization. However, for applications like SPV wallets, it is important to know if the node is not only open for p2p connection, but also for HTTP REST API requests - and on which port or URL. This information is not available on Rest API.

Currently, peers/connected returns a block like this:
```
{
    "address": "/<ip>:9030",
    "lastMessage": 1646725996336,
    "lastHandshake": 1646719641723,
    "name": "ergo-mainnet-4.0.23",
    "connectionType": "Outgoing"
}
```

It could be guessed and tried if the HTTP API is running on `http://<ip>:9053` for mainnet or `http://<ip>:9052` for testnet, but this is some kind of trial&error. Additionally, some node's HTTP API might be even exposed via a https://domain.name/ergoapi url that could not be found this way.

### Solution

I suggest to extend the node in the following way:

1. Add a config entry for REST API. The node administator can enter here the HTTP address the node's REST API is reachable from outside, it should be able to hold a user-entered http(s) domain address, if available. The entry is an opt-in entry, if not filled by admin the REST API is seen as closed for the public.
2. The information of the REST API should be broadcasted to the connected nodes
3. Add a new property to peers API with the HTTP REST API address.
4. Sorting should be flipped for peers/connected - at the moment, the peers with the oldest lastMessage timestamp come first.
 
This way, a light client can build a database of nodes to connect to from the Peers API and can suggest alternatives to the user in case the node to connect to is not reachable.

Needed for ergoplatform/ergo-wallet-app#20

greenhat:
I concur. We're facing the same problem in the SPV node API (https://github.com/ergoplatform/sigma-rust/issues/544).

satsen:
Maybe programs should use the normal connection instead? I don't see the point of having this other than consuming more bandwidth

MrStahlfelge:
[...]

PR #1662, opened 2022-03-22
I1652 peers rest api address

pragmaxim:
See https://github.com/ergoplatform/ergo/issues/1652

I gave a lot of thought into backward compatibility, `PeerSpec` is using hand-written writer/reader and new `restApiUrl` optional field is at the end, with a little hack that allows for reading older `PeerSpec` bytes by newer peer version ... And old peers just won't see the extra `PeerSpec` bytes at the end ... 

I noticed that if anything goes wrong with loading peers from PeerDatabase, the node gets stuck, so I added a fallback to loading peers only from network for better user experience + log ERROR.


MrStahlfelge:
Thank you for the implementation! It looks like what was requested, unfortunately I cannot test the implementation.
What I think should be added somewhere is a documentation to the new configuration entry. Where is such documentation added for now?

Also adding @greenhat to the reviewers as he needs the feature as well.

pragmaxim:
@MrStahlfelge the `application.conf` is documented, I will add it here : 
https://github.com/ergoplatform/ergo/wiki/Set-up-a-full-node#update-config-file-with-api-key-hash

basically you just add `scorex.restApi.publicUrl = "https://example.com"` to the `ergo.conf` if the node exposes it ...

greenhat:
Hey! @pragmaxim thank you for taking care of this! Looks good!
I'm wondering if we could lower the friction for the node operators and allow the hostname to be set to "0.0.0.0" and treat it on the clients as "use the same IP as `address` field in peers/connected". It should make life easier for the node operators that have dynamic white IPs. On the second thought that might add more confusion, so it might not be a good idea after all.

MrStahlfelge:
Problem here would be the port, but the node should know the port it uses and could fill it in correctly.

pragmaxim:
I leave this up to you guys, I haven't done any client apps talking to node yet ... If we should just switch the port on the node's address, then the config would be something like `publicRestApiEnabled = true` ... But @MrStahlfelge mentioned some concerns with wifis and firewalls ... ?

MrStahlfelge:
http://....:9053 is blocked on many company networks or public wifis, people regularly contact me about it. So a hostname url with https support is fortunate to have and should be able to be reported by the api.

pragmaxim:
@MrStahlfelge @greenhat My thinking also was that this address should be probably a domain name or static IP address as it is called by a user-facing application and therefor it should be reliably accessible ... whereas the addresses from binding to `0.0.0.0` are often dynamic and not persistent in time. This problem resolves in peer-to-peer network automatically by simply discarding unreachable peers...  

Also the unstable and unreliable peers that could be up/down should not get listed perhaps either as it would be another source of unreliability ...

MrStahlfelge:
@pragmaxim Was this a question, a con or a pro for leaving it as it is? I agree on everything you've wrote but I am not sure what conclusion you take from it.

pragmaxim:
@MrStahlfelge Question is how many node operators is going to be there with a DNS record ... if at least 3 such nodes are going to be there, I'd incline to the current version as this settings basically says "hey, everyone can use this node" ... There could be node operators who want the restAPI to be exposed but just for their own use, they would not use this setting at all.

MrStahlfelge:
If they don't want the rest api to be exposed, they should not set it as exposed in the settings - that is exactly what my request was aiming for.

kushti (approved):

Issue #1642, opened 2022-02-18
Wallet - start the scan at a specific block


pragmaxim:
When rescanning all blocks after restoring wallet, there should be a parameter to start the scan at a specific block. 
For example if I only created the wallet at block 600k I should not need to scan starting from block 0

PR #1665, opened 2022-03-23
i1642 start the scan at specific height

pragmaxim:
See https://github.com/ergoplatform/ergo/issues/1642

I added also `/wallet/rescan` POST request as it is an execution command and GET request is quite unexpected.

It is well tested, both GET and POST request work fine, repository is deleted and wallet is rescanned from given height. If no parameter is given, rescanning starts from the beginning.


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -200,9 +200,9 @@ class ErgoWalletActor(settings: ErgoSettings,
       )
       context.become(loadedWallet(newState))
 
-    case ScanInThePast(blockHeight) =>
+    case ScanInThePast(blockHeight, forceScan) =>
       val nextBlockHeight = state.expectedNextBlockHeight(blockHeight, settings.nodeSettings.isFullBlocksPruned)
-      if (nextBlockHeight == blockHeight) {
+      if (nextBlockHeight == blockHeight || forceScan) {
```
pragmaxim, line 205: @kushti I had to make an exception here to allow rescaninng from arbitrary height

kushti (approved):

Issue #1668, opened 2022-03-24
Change scan id type to long


kushti:
Currently, scan id is Short, so max positive scan id is 32767. For some applications it seems this is not enough. for example, Auction House got into troubles with negative scan ids due to overflow. 

Make ScanId long, but also preserve compatibility for old scans already stored in the db.

pragmaxim:
@kushti isn't `Long` overkill? 2.15 billions with Integer could do, right?

pragmaxim:
@kushti the backward compatibility is very non-trivial ... I would suggest to have a Version check at boot time and if it is `4.0.25` then we would copy all the `short` scanIds into `int` scanIds and delete the shorts ...

cannonQ:
Is this bounty still in place?

PR #1670, opened 2022-03-29
I1668 scan id from short to int

pragmaxim:
See https://github.com/ergoplatform/ergo/issues/1668

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletState.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletState.scala
@@ -1,5 +1,7 @@
 package org.ergoplatform.nodeView.wallet
 
+import cats.Traverse
+import cats.implicits._
 import com.google.common.hash.BloomFilter
 import org.ergoplatform.ErgoBox.BoxId
 import org.ergoplatform._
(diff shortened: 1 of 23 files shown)


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletState.scala
```diff
@@ -135,24 +140,59 @@ object ErgoWalletState {
...
+          parameters,
+          maxInputsToUse
+        )
+
+      // compatibility bridge to version 4.0.26 where ScanId changed from Short to Int, let's migrate them
+      if (Version(ergoSettings.scorexSettings.network.appVersion).compare(Version(4, 0, 26)) >= 0) {
+        val legacyScans = ergoStorage.getLegacyScans
+        if (legacyScans.exists(_.scanId.intValue < 0)) {
+          log.info(s"Scans already migrated")
+          Success(buildState)
```
pragmaxim, line 169: @kushti This is the only problematic part, after migrating scanIds (deleting all scans with `Short` keys and creating new scan with `Int`  keys), `getLegacyScans` returns some weird negative `Short` key as if the `remove` operation has BUG in it.

kushti (changes requested):
It seems that old indexes (built with composeKey* functions) can be damaged. Please check and do new tests accordingly

pragmaxim:
@kushti how did you find it can be damaged? I could use some clue

pragmaxim:
@kushti I think you misunderstood the logic behind migration, I described it here https://github.com/ergoplatform/ergo/issues/1668#issuecomment-1079763327 ... we migrate the ids at the boot time and then we type-safely work only with Ints

kushti:
> @kushti I think you misunderstood the logic behind migration, I described it here [#1668 (comment)](https://github.com/ergoplatform/ergo/issues/1668#issuecomment-1079763327) ... we migrate the ids at the boot time and then we type-safely work only with Ints

How did you test it ?

pragmaxim:
@kushti the sequence of testing steps is in the PR description.

kushti:
@pragmaxim made custom testing, no results returned for old scans

pragmaxim:
@kushti I was working on this almost 14 months ago, it is too stale for me to continue ... the LevelDB layer is very hard to reason about

PR #1671, opened 2022-03-29
add testnet address to swagger

pragmaxim:


pragmaxim:
Not needed, see https://github.com/ergoplatform/ergo/pull/1673

PR #1700, opened 2022-05-10
EIP-27 implementation: edits

aslesarenko:
- changed contracts code a bit
- added test case which compares tree sizes and do pretty printing (compiled tree is smaller)

### DIFF
--- a/src/test/scala/org/ergoplatform/reemission/ReemissionRulesSpec.scala
+++ b/src/test/scala/org/ergoplatform/reemission/ReemissionRulesSpec.scala
@@ -1,18 +1,20 @@
 package org.ergoplatform.reemission
 
-import org.ergoplatform.{ErgoBox, ErgoBoxCandidate, ErgoLikeTransaction, ErgoScriptPredef, Input}
+import org.ergoplatform._
 import org.ergoplatform.settings.{MonetarySettings, ReemissionSettings}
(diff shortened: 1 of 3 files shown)


kushti (changes requested):
Refactoring is merged into fee-improvs

src/test/scala/org/ergoplatform/reemission/ReemissionRulesSpec.scala
```diff
@@ -210,4 +212,38 @@ class ReemissionRulesSpec extends ErgoPropertyTest with ErgoTestConstants {
...
+    val version: Byte = 1
+    val headerFlags = ErgoTree.headerWithVersion(version)
+    val compiledTree = ErgoTree.fromProposition(headerFlags, compiledProp)
+
+    println(s"====  compiled $name  ================================")
+    println(s"size of $name: ${compiledTree.bytes.length}")
+//    SigmaPPrint.pprintln(compiledTree, 150, 250)
+  }
+
+  property("compile reemission scripts") {
```
kushti, line 231: This is not a real test, rather, an utility printing something. so should be done as a *Printer utility in org.ergoplatform.tools. Also, I do not understand what is it doing. Comments are needed.

aslesarenko: yes, was just to understand the effect of compilation.
I think this PR can just be closed without merging.

Issue #1702, opened 2022-05-10
In case of recoverable failure, status in ErgoNodeViewSynchronizer not cleared


kushti:
```
    case RecoverableFailedModification(_, _) =>
      // we ignore this one as we should try to apply this modifier again
```

however, to apply it again, we should download it again first, and status is likely `received` at this point and will be so. Thus in case of headers coming out-of-order, a node will stuck anyway. 

Please clear the status, do not just ignore `RecoverableFailedModification` message.

Please do a test in `ErgoNodeViewSynchronizerSpecification` , showing that the node is clearing the status (and asking modifier again on getting its INV). 

Related issues and PRs: #1615, #1618, #1657, #1680

PR #1704, opened 2022-05-14
i1702 receiving headers out of order bug fix

pragmaxim:
Fixes https://github.com/ergoplatform/ergo/issues/1702


kushti (approved):

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -656,6 +656,9 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
         deliveryTracker.setInvalid(pmod.id, pmod.modifierTypeId)
         penalizeMisbehavingPeer(remote)
         false
+      case Failure(e) if e.isInstanceOf[RecoverableModifierError] =>
+        context.system.eventStream.publish(RecoverableFailedModification(pmod, e))
+        false
```
pragmaxim, line 661: We should recover ASAP here, there is no reason to pass this modifier further to be applied. Neither of the recoverable errors should imho lead to modifier application, it should be simply skipped and set as `Unknown` ASAP.

kushti (approved):

Issue #1714, opened 2022-05-20
Slow headers application since 4.0.28


kushti:
Since 4.0.28, headers being applying slowly. Probably related to quick invalidation introduced in #1704

kushti:
fixed in #1715

PR #1715, opened 2022-05-21
Fix for #1714: slow headers application

kushti:
In this PR, fail-fast condition in ErgoNodeViewSynchronizer which was added in #1704 : 

``
case Failure(e) if e.isInstanceOf[RecoverableModifierError] =>
        context.system.eventStream.publish(RecoverableFailedModification(pmod, e))
        false
```

now removed. This condition prevents headers from being placed into cache stored in a NodeViewHolder instance. 

`receiving out-of-order header should request it again` test fixed. See comments there on how the test works now.

### DIFF
--- a/src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
+++ b/src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
@@ -260,6 +260,8 @@ class ErgoNodeViewSynchronizerSpecification extends HistoryTestHelpers with Matc
 
       deliveryTracker.reset()
 
+      // we generate fork of two headers, starting from the parent of the best header
+      // so the depth of the rollback is 1, and the fork bypasses the best chain by 1 header
       val hhistory = ErgoHistory.readOrGenerate(settings, timeProvider)
       val parent = hhistory.lastHeaders(2).head
       val smallFork = genHeaderChain(_.size > 2, Some(parent), hhistory.difficultyCalculator, None, false)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala [hunk 1]
```diff
@@ -226,16 +247,32 @@ class ErgoNodeViewSynchronizerSpecification extends HistoryTestHelpers with Matc
...
+
       deliveryTracker.reset()
-      deliveryTracker.setRequested(Seq(chain.last.id), Header.modifierTypeId, Some(peer))(_ => Cancellable.alreadyCancelled)
-      val modData = ModifiersData(Header.modifierTypeId, Map(chain.last.id -> chain.last.bytes))
-      val modSpec = new ModifiersSpec(100)
-      node ! Message(modSpec, Left(modSpec.toBytes(modData)), Some(peer))
-      // desired state of submitting headers out of order is Unknown, they need to be downloaded again
+
+      // we generate fork of two headers, starting from the parent of the best header
+      // so the depth of the rollback is 1, and the fork bypasses the best chain by 1 header
```
pragmaxim, line 265: Why do we need a fork for reproducing this?

pragmaxim (approved):

src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
kushti: We need for a header @ height of best chain height + 1, otherwise the header will be applied to cache just and not validated inside `pmodModify` (where `RecoverableModifierError` is thrown). To test out-ot-order application then, we need to generate and a fork which first forked header is @ best chain height, and then there's the header we are trying to apply (@ height of best chain + 1) without presenting its parent first.

PR #1717, opened 2022-05-26
Removing EIP27Supported flag

kushti:
In this PR, since in both testnet and mainnet EIP 27 is locked in, eip27Supported flag removed for simplicity's sake. Then the node will has follow EIP-27 since activation height just ( while 4.0.29 and 4.0.30 need for blockchain scan to extract voting results). 

Since this PR, EIP-27 checks supported in "digest" mode.


knizhnik (approved):

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -217,105 +217,95 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
+      if (stateContext.currentHeight >= activationHeight) {
+        // reemission check logic below
+        var reemissionSpending = false
+        boxesToSpend.foreach { box =>
+          // checking EIP-27 rules for emission box
+          // for efficiency, skip boxes with less than 100K ERG
+          if (box.value > 100000 * EmissionRules.CoinsInOneErgo) {
+            // on activation height, emissionNft is not in emission box yet, but in injection box
+            if (box.tokens.contains(emissionNftId) ||
+              (height == activationHeight && boxesToSpend(1).tokens.contains(emissionNftId))) {
```
pragmaxim, line 245: `boxesToSpend(1).tokens.contains(emissionNftId)` would be probably better to have as `def` with self-explanatory name... why do we use tokens from the second box?

pragmaxim (approved):

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
kushti: Improved comment on injection box index. Not sure about good name for one-liner

PR #1719, opened 2022-05-26
Search for emission box when tracking is lost

kushti:
If emission box tracking is somehow lost (witnessed once, so still need to understand how is that possible), no new search is done, so the issue prevents from mining blocks with reward tx at all. 

In this PR additional search is done if tracking is lost.


arcadiancomp:
Thanks again for the fix here, RKDN pool is back up and working!

pragmaxim (approved):

Issue #1722, opened 2022-05-29
Do not wallet-rescan blockchain from height 0 after wallet initialization


kushti:
Currently, the node wallet is re-scanning the blockchain after wallet initialization, and that does not make much sense. Better to avoid rescan after init.

pragmaxim:
@kushti I don't understand, wallet is empty after initialization and we need to scan the blockchain to compute its state.

kushti:
@pragmaxim wallet initialized with a new secret key (generated from cryptographically strong randomness), thus blockchain does not contain anything related to the new wallet, unlike wallet restoration from (possibly, previously used) mnemonic

scalahub:
Yes, also one feature to add if possible:
- Manually add UTXOs to wallet instead of scanning

dimension-drifter:
@kushti please review my PR for this.

PR #1723, opened 2022-05-30
i1722-avoid-wallet-rescan-after-init [Testing]

pragmaxim:
Fixes https://github.com/ergoplatform/ergo/issues/1722

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -85,19 +85,19 @@ class ErgoWalletActor(settings: ErgoSettings,
       val ws = settings.walletSettings
       // Try to read wallet from json file or test mnemonic provided in a config file
       val newState = ergoWalletService.readWallet(state, ws.testMnemonic.map(SecretString.create(_)), ws.testKeysQty, ws.secretStorage)
-      context.become(loadedWallet(newState))
+      context.become(loadedWallet(newState, initialized = false))
       unstashAll()
     case _ => // stashing all messages until wallet is setup
       stash()
   }
 
-  private def loadedWallet(state: ErgoWalletState): Receive = {
+  private def loadedWallet(state: ErgoWalletState, initialized: Boolean): Receive = {
     // Init wallet (w. mnemonic generation) if secret is not set yet
     case InitWallet(pass, mnemonicPassOpt) if !state.secretIsSet(settings.walletSettings.testMnemonic) =>
       ergoWalletService.initWallet(state, settings, pass, mnemonicPassOpt) match {
         case Success((mnemonic, newState)) =>
           log.info("Wallet is initialized")
-          context.become(loadedWallet(newState))
+          context.become(loadedWallet(newState, initialized = true))
           self ! UnlockWallet(pass)
           sender() ! Success(mnemonic)
         case Failure(t) =>
@@ -111,7 +111,7 @@ class ErgoWalletActor(settings: ErgoSettings,
       ergoWalletService.restoreWallet(state, settings, mnemonic, mnemonicPassOpt, walletPass) match {
         case Success(newState) =>
           log.info("Wallet is restored")
-          context.become(loadedWallet(newState))
+          context.become(loadedWallet(newState, initialized = false))
           self ! UnlockWallet(walletPass)
           sender() ! Success(())
         case Failure(t) =>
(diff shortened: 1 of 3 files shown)


src/test/scala/org/ergoplatform/nodeView/wallet/ErgoWalletServiceSpec.scala
```diff
@@ -297,6 +299,8 @@ class ErgoWalletServiceSpec
         val pass = Random.nextString(10)
         val initializedState = walletService.initWallet(walletState, settings, SecretString.create(pass), Option.empty).get._2
 
+        initializedState.getWalletHeight shouldBe 0
```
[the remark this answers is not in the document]
pragmaxim: Why shouldn't it be true for wallet just initialized?

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -265,7 +265,8 @@ class ErgoWalletActor(settings: ErgoSettings,
     case ScanOnChain(newBlock) =>
       if (state.secretIsSet(settings.walletSettings.testMnemonic)) { // scan blocks only if wallet is initialized
         val nextBlockHeight = state.expectedNextBlockHeight(newBlock.height, settings.nodeSettings.isFullBlocksPruned)
-        if (nextBlockHeight == newBlock.height) {
+        // we want to scan a block either when it is its turn or when wallet is freshly created (no need to load the past)
+        if (nextBlockHeight == newBlock.height || (state.walletState == WalletPhase.Created && state.getWalletHeight == 0)) {
```
pragmaxim: @kushti I've been through this PR one more time, I once found this solution after many thoughts. We basically want the first block applied to chain to be the first  block added to freshly created wallet (out of order in this special case). We cannot manually change the digest of the wallet as that can be done only by applying modifiers to it. So this is the best place to do that.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -267,7 +267,9 @@ class ErgoWalletActor(settings: ErgoSettings,
     case ScanOnChain(newBlock) =>
       if (state.secretIsSet(settings.walletSettings.testMnemonic)) { // scan blocks only if wallet is initialized
         val nextBlockHeight = state.expectedNextBlockHeight(newBlock.height, settings.nodeSettings.isFullBlocksPruned)
-        if (nextBlockHeight == newBlock.height) {
+        // we want to scan a block either when it is its turn or when wallet is freshly created (no need to load the past)
+        val walletFreshlyCreated = state.walletPhase == WalletPhase.Created && state.getWalletHeight == 0
```
pragmaxim, line 271: If I remember correctly, only `state.getWalletHeight == 0` is not enough as restored wallet has also `state.getWalletHeight == 0`, that's why the `WalletPhase` introduction

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletService.scala
```diff
@@ -563,6 +564,7 @@ class ErgoWalletServiceImpl(override val ergoSettings: ErgoSettings) extends Erg
         block,
         state.outputsFilter,
         dustLimit,
+        state.walletPhase == WalletPhase.Created,
```
kushti, line 567: Why always Created here?

pragmaxim: The `scanBlockTransactions` takes `walletCreated: Boolean` parameter as there is check inside that requires information whether it was restored or created ...

src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala
```diff
@@ -252,7 +252,7 @@ class WalletRegistry(store: LDBVersionedStore)(ws: WalletSettings) extends Score
 
     // and update wallet digest
     updateDigest(bag3) { case WalletDigest(height, wBalance, wTokensSeq) =>
-      if (height + 1 != blockHeight) {
+      if (height + 1 != blockHeight && !walletCreated) {
         log.error(s"Blocks were skipped during wallet scanning, from $height until $blockHeight")
```
pragmaxim, line 256: This is the only reason why the walletPhase is being propagated down the `WalletRegistry` and `WalletScanLogic`

Issue #206, opened 2018-03-05
IT-tests


catena2w:
You can start it tests with `sbt it:test` command (it will start docker and require root access or additional rights)
- [ ] enable tests in jenkins (https://jenkins.ergoplatform.com/job/ergo-it/)
- [ ] limit nodes in resources
- [ ] ensure, they work correct
- [ ] write more tests

PR #1728, opened 2022-06-03
Comparing pay-to-reemission trees by proposition

### LINKED ISSUE #1726: Split in testnet between 4.0.29+ and prior versions
There's a consensus-level disagreement affecting testnet (only). As pay-to-reemission tree is switched to v1 in https://github.com/ergoplatform/ergo/pull/1373/commits/78a8bfc80be4901715bdb72f4f7c9ee0d48c981e (merged into 4.0.29), and comparison is done over the trees during transaction validation, 4.0.29+ and prior versions currently disagree on transactions burning re-emission tokens (while paying to pay-to-reemission outputs). 

As a particular solution, we can compare by propositions.

kushti:
This PR fixes #1726 : 

There's a consensus-level disagreement affecting testnet (only). As pay-to-reemission tree is switched to v1 in https://github.com/ergoplatform/ergo/commit/78a8bfc80be4901715bdb72f4f7c9ee0d48c981e (merged into 4.0.29), and comparison is done over the trees during transaction validation, 4.0.29+ and prior versions currently disagree on transactions burning re-emission tokens (while paying to pay-to-reemission outputs).

As a solution, we now compare trees by propositions. Done for the testnet only, in the mainnet, to avoid disagreement with older versions, we compare trees as before.

### DIFF
--- a/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
+++ b/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
@@ -300,7 +300,7 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
           log.debug(s"Reemission tokens to burn: $toBurn")
           val reemissionOutputs = outputCandidates.filter { out =>
             require(!out.tokens.contains(reemissionTokenId), "outputs contain reemission token")
-            out.ergoTree == payToReemissionContract
+            out.ergoTree.toProposition(true) == payToReemissionContract.toProposition(true)
           }
           val sentToReemission = reemissionOutputs.map(_.value).sum
           require(sentToReemission == toBurn, "Burning condition violated")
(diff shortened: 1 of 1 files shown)


greenhat (approved):

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -300,7 +300,11 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
           log.debug(s"Reemission tokens to burn: $toBurn")
           val reemissionOutputs = outputCandidates.filter { out =>
             require(!out.tokens.contains(reemissionTokenId), "outputs contain reemission token")
-            out.ergoTree == payToReemissionContract
+            if (chainSettings.isMainnet) {
+              out.ergoTree == payToReemissionContract
+            } else {
+              out.ergoTree.toProposition(true) == payToReemissionContract.toProposition(true)
+            }
```
aslesarenko, line 307: could you please describe, why the original comparison `out.ergoTree == payToReemissionContract` won't affect mainnet?
Also, shoudn't testnet also have protocol upgrades to be on par with mainnet?

kushti: The testnet was activated with old clients, issuing and accepting (the most important part) v0 transactions burning re-emission tokens while sending ERGs to pay-to-reemission outputs. Such transactions appeared in block #206,155. Mainnet has no such transactions in the mainnet (and they could be problematic after EIP-27 activation only).

To be on par, we need for testnet relaunch at this point.

Issue #1742, opened 2022-06-14
Extract new header from sync V2 when it is there 


kushti:
The node can get a new header from sync V2 message (a header which parent is our best header). 
In this case, instead of existing slow approach to get the header (which makes sense for v1), we can apply the header immediately.

kushti:
@pragmaxim I guess only last header (so one) header is enough, as others are not going in consecutive order, but shifted from chain tip known to the peer with offsets (16, 128, 512)

I think it makes sense to apply if the most recent header's parent ID is our best known header ID , otherwise use standard route (as you need to download more than header in this case).

Bypassing DeliveryTracker is okay. Header's status in DT will be "held" after application (so within milliseconds).

Issue #1743, opened 2022-06-14
Ask block sections directly from peer sending us a new header


kushti:
When the node is synced, ask block sections from the peer which delivered a new header (updating our best chain) immediately.

pragmaxim:
@kushti can you please clarify the reasoning behind this? Why would it be faster/better than the current way?

kushti:
@pragmaxim first of all, currently we're asking some random peers (if our peer is not connected to new block miner) for block sections. For new blocks, there's also broadcasting:

```
    case SemanticallySuccessfulModifier(mod) =>
      broadcastInvForNewModifier(mod)
```

which is making quick delivery.

However, as we now can extract a new header from sync (which will be sent immediately after block application), we can ask immediately corresponding block sections, so INV exchange could be avoided at all.

PR #1744, opened 2022-06-15
I1742 & I1743 apply continuation header from syncv2 and download block sections

pragmaxim:
Fixes https://github.com/ergoplatform/ergo/issues/1742 #1743

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
@@ -267,6 +267,26 @@ trait ErgoHistoryReader
   }
 
   /**
+    * Finding other peer's continuation header from a header that is common to our node's history
+    * @param syncInfo  other's node sync info
+    * @return maybe continuation header
+    */
+  def continuationHeaderV2(syncInfo: ErgoSyncInfoV2): Option[Header] = {
+    if (syncInfo.lastHeaders.isEmpty) {
+      Option.empty
+    } else {
+      // let's find most recent header that is common to our history
+      commonPoint(syncInfo.lastHeaders) match {
+        case Some(commonHeader) =>
+          // now let's find continuation header whose parent is the common header
+          syncInfo.lastHeaders.find(_.parentId == commonHeader.id)
+        case None =>
+          Option.empty
+      }
+    }
+  }
+
+  /**
     * Calculating continuation from common header which will be sent to another node
     * if comparison status is YOUNGER or FORK.
     *
--- a/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
@@ -275,13 +275,12 @@ trait ErgoHistoryReader
     if (syncInfo.lastHeaders.isEmpty) {
       Option.empty
     } else {
-      // let's find most recent header that is common to our history
-      commonPoint(syncInfo.lastHeaders) match {
-        case Some(commonHeader) =>
-          // now let's find continuation header whose parent is the common header
-          syncInfo.lastHeaders.find(_.parentId == commonHeader.id)
-        case None =>
-          Option.empty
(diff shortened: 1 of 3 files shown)


pragmaxim:
the biggest problem in the Synchronizer test suite is that this https://github.com/ergoplatform/ergo/blob/3fa0ed3d73bcc7d1e0130a347b182b8f5313f49e/src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala#L100
history is used in Synchronizer but not in ErgoNodeViewHolder, which reads it from disk empty and we cannot test effects of Synchronizer on history done in ErgoNodeViewHolder

kushti:
> @kushti the biggest problem in the Synchronizer test suite is that this
> 
> https://github.com/ergoplatform/ergo/blob/3fa0ed3d73bcc7d1e0130a347b182b8f5313f49e/src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala#L100
> 
> history is used in Synchronizer but not in ErgoNodeViewHolder, which reads it from disk empty and we cannot test effects of Synchronizer on history done in ErgoNodeViewHolder

Yes, history is different in Synchronizer and ENVH, that's why I made `Synchronizer2Fixture` in 
```
property("NodeViewSynchronizer: receiving out-of-order header should request it again") {
    withFixture2 { ctx =>
```

src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
```diff
@@ -236,37 +235,87 @@ class ErgoNodeViewSynchronizerSpecification extends HistoryTestHelpers with Matc
...
       }
 
       deliveryTracker.reset()
 
       // we generate fork of two headers, starting from the parent of the best header
       // so the depth of the rollback is 1, and the fork bypasses the best chain by 1 header
       val hhistory = ErgoHistory.readOrGenerate(settings, timeProvider)
-      val parent = hhistory.lastHeaders(2).head
+      val baseChain = genHeaderChain(_.size > 2, None, hhistory.difficultyCalculator, None, false)
+      baseChain.headers.foreach(sendHeader)
```
pragmaxim, line 306: @kushti I made this change in out-of-order test as it fails on CI with empty `.ergo_test/history` folder ... it only works if some preceding tests generates some headers apriory.

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -266,6 +266,26 @@ trait ErgoHistoryReader
...
+    */
+  def continuationHeaderV2(syncInfo: ErgoSyncInfoV2): Option[Header] = {
+    if (syncInfo.lastHeaders.isEmpty) {
+      Option.empty
+    } else {
+      // let's find most recent header that is common to our history
+      commonPoint(syncInfo.lastHeaders) match {
+        case Some(commonHeader) =>
+          // now let's find continuation header whose parent is the common header
+          syncInfo.lastHeaders.find(_.parentId == commonHeader.id)
```
kushti, line 282: In the real network (i.e. unlike tests) headers in syncv2 are not consecutive (they are located at offsets 0,16,128,512 from the chain tip), so this code will never succeed 

I think should be like

```
  if (syncInfo.lastHeaders.isEmpty) {
      Option.empty
    } else {
      val lastHeader = syncInfo.lastHeaders.head
      if (bestHeaderIdOpt.contains(lastHeader.parentId)) {
        Some(lastHeader)
      } else {
        Option.empty
      }
    }
```

kushti (changes requested):

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
pragmaxim: @kushti I was under the impression that the `syncV2` headers would rarely begin with our node's `bestHeaderIdOpt` so I was trying to find the `commonPoint` and apply header whose parent is the common point.

kushti (approved):

PR #1750, opened 2022-06-22
Prioritize blocks downloading

kushti:
In this PR, we introduce new rule for downloading transactions, so we do not downloading transactions if some peer around has a better chain tip. This prioritizes blocks downloading and so helping with getting synced again quicker.


pragmaxim (approved):
LGTM

greenhat (approved):

PR #1752, opened 2022-06-24
Find emission box by its NFT

### LINKED ISSUE #1718: Find emission box by its NFT (after EIP-27 activation height)
Since activation height, it would be more efficient to avoid comparing ErgoTree instances in `extractEmissionBox` :  

```
protected[state] def extractEmissionBox(fb: ErgoFullBlock): Option[ErgoBox] = {
    def fullSearch(fb: ErgoFullBlock): Option[ErgoBox] = {
      fb.transactions
        .find(_.outputs.head.ergoTree == constants.settings.chainSettings.monetary.emissionBoxProposition)
        .map(_.outputs.head)
    }

    emissionBoxIdOpt match {
      case Some(id) =>
        fb.blockTransactions.txs.view.reverse.find(_.inputs.exists(t => java.util.Arrays.equals(t.boxId, id))) match {
          case Some(tx) if tx.outputs.head.ergoTree == constants.settings.chainSettings.monetary.emissionBoxProposition =>
            tx.outputs.headOption
          case Some(_) =>
            log.info(s"Last possible emission box consumed")
            None
          case None =>
            log.warn(s"Emission box possibly not spent in block ${fb.encodedId}")
            boxById(id) match {
              case s: Some[ErgoBox] => s
              case None => fullSearch(fb)
            }

        }
      case None =>
        log.debug("No emission box: emission should be already finished before this block")
        fullSearch(fb)
    }
  }
```

by finding emission box via looking for a box having 2 tokens where first token is emission box NFT.

jellymlg:
closes #1718

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/UtxoStateReader.scala
@@ -80,11 +80,16 @@ trait UtxoStateReader extends ErgoStateReader with TransactionValidation {
     * @param fb - ergo full block
     * @return emission box from this block transactions
     */
-  // todo: search by emission box NFT after EIP-27 activation height, https://github.com/ergoplatform/ergo/issues/1718
   protected[state] def extractEmissionBox(fb: ErgoFullBlock): Option[ErgoBox] = {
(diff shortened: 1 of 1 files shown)


jellymlg:
@kushti

Issue #1756, opened 2022-06-27
Consider work done when comparing chains (using sync v2) 


kushti:
`ErgoHistory.compareV2` now blindly trust headers coming from peers, and checking only heights there. 

To improve things, we need:

* check PoW of headers we are getting
* compare chains using work (difficulty), not just chain length

PR #1757, opened 2022-06-28
I1756 pow check when comparing chains [WIP]

pragmaxim:
Fixes https://github.com/ergoplatform/ergo/issues/1756

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
@@ -105,7 +105,13 @@ trait ErgoHistoryReader
     */
   private def commonPoint(headers: Seq[Header]): Option[Header] = {
     headers.find { h =>
-      contains(h.id)
+      contains(h.id) &&
+        historyStorage.modifierById(h.id).exists {
+          case Header(_, _, _, _, _, _, _, _, _, pow, _, _) =>
+            h.powSolution == pow
+          case _ =>
+            false
+        }
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -105,7 +105,13 @@ trait ErgoHistoryReader
     */
   private def commonPoint(headers: Seq[Header]): Option[Header] = {
     headers.find { h =>
-      contains(h.id)
+      contains(h.id) &&
+        historyStorage.modifierById(h.id).exists {
+          case Header(_, _, _, _, _, _, _, _, _, pow, _, _) =>
+            h.powSolution == pow
```
kushti, line 111: No need for such check (pow already reflected in id, which collision-resistance is cryptographically strong).

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -140,14 +146,15 @@ trait ErgoHistoryReader
...
 
         val otherHeaders = info.lastHeaders
         val otherLastHeader = otherHeaders.head // always available
         val otherHeight = otherLastHeader.height
-        // todo: check PoW of otherLastHeader
+        val otherPow = otherLastHeader.powSolution
 
         if (otherHeight == myHeight) {
-          if (otherLastHeader.id == myLastHeader.id) {
+          if (otherLastHeader.id == myLastHeader.id && otherPow == myPow) {
```
kushti, line 157: No need to check here as well

kushti (changes requested):

PR #1763, opened 2022-07-04
Refactoring of History and Mempool class hierarchies 

kushti:
Another round of post-Scorex refactoring, a lot of abstract basic traits removed.


pragmaxim (approved):
Good idea :+1:

PR #1769, opened 2022-07-06
ErgoSyncTracker rework

### LINKED ISSUE #1774: Clear status in ErgoSyncTracker after some time

kushti:
In this PR:
* close #1774 : after enough time passed we clear status for potentially stalled peer. Helps when the peer is still recognized as connected (but in fact could be disconnected), but the node can query it still because peer status from the past looks good 
* update status functionality fully moved to StatusTracker


pragmaxim:
@kushti does this close https://github.com/ergoplatform/ergo/issues/1774 ?

kushti:
@pragmaxim now yes (wasnt so during your quection)

src/main/scala/org/ergoplatform/network/ErgoSyncTracker.scala
```diff
@@ -130,8 +160,13 @@ final case class ErgoSyncTracker(networkSettings: NetworkSettings, timeProvider:
         val unknowns = statuses.filter(_._2.status == Unknown).toVector
         val forks = statuses.filter(_._2.status == Fork).toVector
         val elders = statuses.filter(_._2.status == Older).toVector
-        val nonOutdated =
-          (if (elders.nonEmpty) elders(scala.util.Random.nextInt(elders.size)) +: unknowns else unknowns) ++ forks
+
+        val eldersAndUnknown = if (elders.nonEmpty) {
+          elders(scala.util.Random.nextInt(elders.size)) +: unknowns
```
pragmaxim, line 165: @kushti why do we pick only one older and all unkowns and forks :thinking: ?

pragmaxim (approved):

aslesarenko (approved):

PR #1771, opened 2022-07-11
i1756 consider work done when comparing chains

pragmaxim:
Closes https://github.com/ergoplatform/ergo/issues/1756

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
@@ -149,19 +149,19 @@ trait ErgoHistoryReader
         val otherHeight = otherLastHeader.height
         // todo: check PoW of otherLastHeader
 
-        if (otherHeight == myHeight) {
+        if (otherHeight == myHeight && otherLastHeader.nBits == myLastHeader.nBits) {
           if (otherLastHeader.id == myLastHeader.id) {
             // Last headers are the same => chains are equal
             Equal
-          } else {
-            if (otherLastHeader.nBits == myLastHeader.nBits &&
-              chainSettings.powScheme.validate(otherLastHeader).isSuccess &&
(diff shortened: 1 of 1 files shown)


src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -149,16 +149,19 @@ trait ErgoHistoryReader
         val otherHeight = otherLastHeader.height
         // todo: check PoW of otherLastHeader
 
-        if (otherHeight == myHeight) {
+        if (otherHeight == myHeight && otherLastHeader.nBits == myLastHeader.nBits) {
```
kushti, line 152: No need for nbits validation here

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -149,16 +149,19 @@ trait ErgoHistoryReader
...
           if (otherLastHeader.id == myLastHeader.id) {
             // Last headers are the same => chains are equal
             Equal
-          } else {
+          } else if (chainSettings.powScheme.validate(otherLastHeader).isSuccess) {
```
kushti, line 156: Later on we can compare nBits below

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -149,16 +149,19 @@ trait ErgoHistoryReader
...
             if (commonPoint(otherHeaders.tail).isDefined) {
               Fork
             } else {
               Unknown
             }
+          } else {
+            log.warn(s"Invalid PoW detected for header ${otherLastHeader.id}")
+            Unknown
```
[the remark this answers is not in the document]
pragmaxim: We should penalize the peer I guess in this case

Issue #1773, opened 2022-07-12
Simplify InvSpec, RequestModifierSpec, ModifiersSpec


kushti:
We have some message specs (inherited from Scorex), which are parametrized with config values, but in fact we can't change the values (could lead to the node being banned by peers). 

So hard-code `InvSpec`, `RequestModifierSpec`, `ModifiersSpec` constructor arguments, and remove them from network settings.

jozanek:
Fixed by #1776 @kushti ?

abebeos:
@kushti , closable (fixed, see #1776)

PR #1776, opened 2022-07-17
#1773 Simplify InvSpec, RequestModifierSpec, ModifiersSpec

ccellado:
We have some message specs (inherited from Scorex), which are parametrized with config values, but in fact we can't change the values (could lead to the node being banned by peers)


kushti:
@ccellado actually, the goal of the issue was to eliminate, e.g. `maxInvObjects` in  `class InvSpec(maxInvObjects: Int)`, replacing it with a hardcoded value corresponding to current value in config (and thus eliminating config setting as well), and so having `object` instead of `class` in result.

PR #1777, opened 2022-07-21
Candidate for 4.0.36 release

kushti:
* #1769 - after enough time passed we clear status for potentially stalled peer. Helps when the peer is still recognized as connected (but in fact could be disconnected), but the node can query it still because peer status from the past looks good
* #1779 - locally submitted transaction now checked synchronously 
* #1785 - "Stats collector got strange input: SemanticallySuccessfulModifier(ADProofs..." log messages eliminated

`maxDeliveryChecks` value by default set to 100


src/main/scala/org/ergoplatform/network/ErgoSyncTracker.scala
```diff
@@ -98,21 +114,43 @@ final case class ErgoSyncTracker(system: ActorSystem,
...
+  private[network] def clearOldStatuses(): Unit = {
+    val currentTime = timeProvider.time()
+    val peersToClear = statuses.filter { case (_, status) =>
+      status.lastSyncSentTime.exists(syncTime => (currentTime - syncTime).millis > ClearThreshold)
+    }.keys
+    if (peersToClear.nonEmpty) {
+      val keysToRemove = peersToClear.toVector
+      log.debug(s"Clearing stalled statuses for $keysToRemove")
+      // we set status to `Unknown` and reset peer's height
+      keysToRemove.foreach(p => updateStatus(p, Unknown, None))
```
pragmaxim, line 129: I guess no need for calling `peersToClear.toVector` ... `keysToRemove` is also confusing name

pragmaxim (approved):

src/main/scala/org/ergoplatform/network/ErgoSyncTracker.scala
kushti: Eliminated `keysToRemove`

PR #1786, opened 2022-07-29
1753 - Unconfirmed Transactions

ApexTheory:


src/main/scala/org/ergoplatform/modifiers/mempool/UnconfirmedTransaction.scala
```diff
@@ -0,0 +1,62 @@
...
+      tx,
+      None,
+      0,
+      0,
+      Some(ErgoTransactionSerializer.toBytes(tx))
+    )
+  }
+}
+
+object UnconfirmedTransactionSerializer extends ScorexSerializer[UnconfirmedTransaction] {
```
kushti, line 42: Why do you need for serializer ?

ApexTheory: Removed as per our discussion.  Added JSON codec instead.

PR #1805, opened 2022-08-11
Randomized pool sorting by transaction size or cost

kushti:
In this pool request, possible pool sorting by fee per cycle (execution cost unit) added. Currently pool sorting is chosen randomly during the node start. In future it makes sense to add a setting for that.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
@@ -147,7 +147,7 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool,
       val poolSizeLimit = nodeSettings.mempoolCapacity
       if (pool.size == poolSizeLimit &&
         weighted(tx, feeFactor).weight <= pool.orderedTransactions.lastKey.weight) {
-        this -> ProcessingOutcome.Declined(new Exception("Transaction is pool outsider"))
+        this -> ProcessingOutcome.Declined(new Exception("Transaction pays less than any other in the pool being full"))
       } else {
         new ErgoMemPool(pool.put(tx, feeFactor), stats, sortingOption) -> ProcessingOutcome.Accepted
       }
--- a/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
@@ -24,6 +24,7 @@ import scala.util.{Failure, Random, Success, Try}
   *                 used for implementing all transaction-related methods
   * @param stats    - Mempool statistics, that allows to track
   *                 information about mempool's state and transactions in it.
+  * @param sortingOption - this input sets how transactions are sorted in the pool, by fee-per-byte or fee-per-cycle
   */
 class ErgoMemPool private[mempool](pool: OrderedTxPool,
                                    private[mempool] val stats : MemPoolStatistics,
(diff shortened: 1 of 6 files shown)


src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -116,61 +126,75 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool,
...
         this -> ProcessingOutcome.DoubleSpendingLoser(doubleSpendingWtxs.map(_.id))
       }
     } else {
-      new ErgoMemPool(pool.put(tx), stats) -> ProcessingOutcome.Accepted
+      val poolSizeLimit = nodeSettings.mempoolCapacity
+      if (pool.size == poolSizeLimit &&
+        weighted(tx, feeFactor).weight <= pool.orderedTransactions.lastKey.weight) {
+        this -> ProcessingOutcome.Declined(new Exception("Transaction pays less than any other in the pool being full"))
+      } else {
+        new ErgoMemPool(pool.put(tx, feeFactor), stats, sortingOption) -> ProcessingOutcome.Accepted
```
knizhnik, line 154: In this case size of the pool will be greater than `poolSizeLimit`.
Is it ok or we should replace element with the smallest weight?

kushti: `pool.put` is removing outsider inside it:

```
if (newPool.orderedTransactions.size > mempoolCapacity) {
      val victim = newPool.orderedTransactions.last._2
      newPool.remove(victim)
    } else {
      newPool
    }
```

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
```diff
@@ -244,7 +269,38 @@ class ErgoMemPool private[mempool](pool: OrderedTxPool,
...
+
+    /**
+      * @return randomly chosen mempool sorting strategy
+      */
+    def random(): SortingOption = {
+      if (Random.nextBoolean()) {
+        log.info("Sorting mempool by fee-per-byte")
+        FeePerByte
+      } else {
+        log.info("Sorting mempool by fee-per-cycle")
```
aslesarenko, line 298: Maybe better free this function from logging side-effect and move the message to the ouside also giving more context info to the message.
Especially when this method is used as default argument value of `empty` method. (see below)

src/test/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolSpec.scala
```diff
@@ -38,6 +38,29 @@ class ErgoMemPoolSpec extends AnyFlatSpec
...
 
+  it should "respect given sorting order" in {
+    implicit val ms = settings.chainSettings.monetary
+    val (us, bh) = createUtxoState(parameters)
+    val genesis = validFullBlock(None, us, bh)
+    val wus = WrappedUtxoState(us, bh, stateConstants, parameters).applyModifier(genesis)(_ => ()).get
+    val inputBox = wus.takeBoxes(1).head
+    val feeOut = new ErgoBoxCandidate(inputBox.value, feeProp, creationHeight = 0)
+    val tx = ErgoTransaction(
+      IndexedSeq(new Input(inputBox.id, new ProverResult(Array.emptyByteArray, ContextExtension.empty))),
```
aslesarenko, line 49:
```suggestion
      IndexedSeq(new Input(inputBox.id, ProverResult.empty)),
```

src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
kushti: done

knizhnik (approved):

PR #1821, opened 2022-08-28
Show public API URL from other peers if own is off

kushti:
In this PR: 
* feature serializers moved in one place
* peer features related abstractions simplified
* now all the feature serializers used, dependless settings (in particular it fixes an issue when public API URL was not parsed if the node is not announcing own public API URL)


pragmaxim (approved):

MrStahlfelge (approved):
LGTM

greenhat (approved):

PR #1825, opened 2022-08-31
i1824 prevent oom from invalidating block

pragmaxim:
@kushti Something like this must be done on many places, especially all kinds of modifier validation.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
@@ -17,13 +17,14 @@ import scorex.core._
 import scorex.core.transaction.Transaction
 import scorex.core.transaction.state.TransactionValidation
 import scorex.core.utils.ScorexEncoding
-import scorex.core.validation.{ModifierValidator}
+import scorex.core.validation.ModifierValidator
 import scorex.crypto.authds.avltree.batch._
 import scorex.crypto.authds.{ADDigest, ADValue}
 import scorex.crypto.hash.Digest32
 import scorex.db.{ByteArrayWrapper, LDBVersionedStore}
 import scorex.util.ModifierId
 
+import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/scala/org/ergoplatform/nodeView/state/UtxoState.scala
```diff
@@ -195,12 +196,12 @@ class UtxoState(override val persistentProver: PersistentBatchAVLProver[Digest32
             new UtxoState(persistentProver, idToVersion(fb.id), store, constants)
           }
         }
-        stateTry.recoverWith[UtxoState] { case e =>
+        stateTry.recoverWith[UtxoState] { case NonFatal(ex) =>
```
pragmaxim, line 199: @kushti `Try` should not contain `Fatal` exceptions but this one actually is OOM error, I have to figure out how it gets there

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
```diff
@@ -24,17 +24,20 @@ class HistoryStorage private(indexStore: LDBKVStore, objectsStore: LDBKVStore, c
...
+    Caffeine.newBuilder()
+      .maximumSize(config.history.headersCacheSize)
+      .build[String, BlockSection]()
 
-  private val blockSectionsCache = CacheBuilder.newBuilder()
-    .maximumSize(config.history.blockSectionsCacheSize)
-    .build[String, BlockSection]
+  private val blockSectionsCache =
+    Caffeine.newBuilder()
+      .maximumSize(config.history.headersCacheSize)
```
kushti, line 34: why headersCacheSize here ? Looks like a mistake.

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
```diff
@@ -24,17 +24,20 @@ class HistoryStorage private(indexStore: LDBKVStore, objectsStore: LDBKVStore, c
...
+      .build[String, BlockSection]()
 
-  private val indexCache = CacheBuilder.newBuilder()
-    .maximumSize(config.history.indexesCacheSize)
-    .build[ByteArrayWrapper, Array[Byte]]
+  private val indexCache =
+    Caffeine.newBuilder()
+      .maximumSize(config.history.headersCacheSize)
```
kushti, line 39: same, why headersCacheSize ?

kushti (changes requested):

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
pragmaxim: Uff, I misread the lines

kushti (approved):

Issue #1827, opened 2022-09-03
Repair corrupted UtxoState & History states after OOM error


pragmaxim:
we should take care about this inconsistency https://github.com/ergoplatform/ergo/issues/1824 in https://github.com/ergoplatform/ergo/blob/66ae8ad214e0f56bd8ad39bda0c752fe5a3354a4/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala#L261
as the node gets stuck and could be unstuck this way.

Issue #1824, opened 2022-08-31
OOM error leads to node being stuck


pragmaxim:
The problem is that Guava cache `org.ergoplatform.nodeView.history.storage.HistoryStorage#headersCache` and friends wrap unchecked OOM Erorr into unchecked `ExecutionError` like this `com.google.common.util.concurrent.ExecutionError: java.lang.OutOfMemoryError: Java heap space` and it is then not propagated to the main thread but rather considered a Validation Error and modifiers get invalidated :facepalm:
```
20:24:10.244 WARN  [tor.default-dispatcher-18] o.e.n.state.UtxoState - Error while applying full block with header dcd6ea12ab0fa7dc7bf094cbde90180b35c82179f8b63772bba3f2a7ebdd3365 to UTXOState with root aeaa271187cf8a13d4503879df69bd83533b06cdf8707431d25c7cecc4e0c64619, reason: scorex.core.validation.MalformedModifierError: Scripts of all transaction inputs should pass verification. 3d524e8bef3698dc9bba5323f15a17a2bda33d86b25a655ce3b6428f3c7d126f: #1 => Failure(com.google.common.util.concurrent.ExecutionError: java.lang.OutOfMemoryError: Java heap space) 
scorex.core.validation.MalformedModifierError: Scripts of all transaction inputs should pass verification. 3d524e8bef3698dc9bba5323f15a17a2bda33d86b25a655ce3b6428f3c7d126f: #1 => Failure(com.google.common.util.concurrent.ExecutionError: java.lang.OutOfMemoryError: Java heap space)
        at org.ergoplatform.settings.ErgoValidationSettings.$anonfun$getError$1(ErgoValidationSettings.scala:33)
        at org.ergoplatform.settings.ErgoValidationSettings.getError(ErgoValidationSettings.scala:33)
20:24:10.245 WARN  [tor.default-dispatcher-18] o.e.n.UtxoNodeViewHolder - Invalidating transaction 3d524e8bef3698dc9bba5323f15a17a2bda33d86b25a655ce3b6428f3c7d126f in mempool due to Scripts of all transaction inputs should pass verification. 3d524e8bef3698dc9bba5323f15a17a2bda33d86b25a655ce3b6428f3c7d126f: #1 => Failure(com.google.common.util.concurrent.ExecutionError: java.lang.OutOfMemoryError: Java heap space)
[...]

PR #1828, opened 2022-09-03
i1827 improve node repairing when node is stuck

pragmaxim:
Closes https://github.com/ergoplatform/ergo/issues/1827

So far I just improved it so we try to repair history in case the node is stuck already after start.

But IDK how to improve the repair algorithm so we unstuck the case that happened in https://github.com/ergoplatform/ergo/issues/1824

@kushti Do you know how to do it?


pragmaxim:
This is incorrect as `repair` is called always on-start.

Issue #843, opened 2019-07-18
Move wallet/p2s and wallet/p2sh under /script


kushti:
A new /script prefix is intended for methods providing interface to the ErgoScript compiler and ErgoTree reducer. It is reasonable then to move wallet/p2s and wallet/p2sh under /compiler

kushti:
fixed in 3.0.7

PR #1845, opened 2022-09-26
EIP-37 implementation and activation

kushti:
This PR contains implementation and activation of EIP-37 described in https://github.com/ergoplatform/eips/pull/79. 

Activation: it is possible to activate EIP-37 after block #843,776 and before block #851,969 . For activation, 232 or more votes for activation required in last 256 blocks, with voting checked every 128 blocks (for blocks which height % 128 == 1), and immediate activation once threshold is met. 

Also, in this PR:
* `offlineGeneration` is now `true` by default for the mainnet (which is what is needed by the pools all the time)


src/main/scala/org/ergoplatform/mining/difficulty/DifficultyAdjustment.scala
```diff
@@ -33,8 +32,56 @@ class LinearDifficultyControl(val chainSettings: ChainSettings) extends ScorexLo
...
+    */
+  def eip37Calculate(previousHeaders: Seq[Header], epochLength: Int): Difficulty = {
+    require(previousHeaders.size >= 2, "at least two headers needed for diff recalc")
+    val lastDiff = previousHeaders.last.requiredDifficulty
+
+    val predictiveDiff = calculate(previousHeaders, epochLength)
+    val limitedPredictiveDiff = if (predictiveDiff > lastDiff) {
+      predictiveDiff.min(lastDiff * 3 / 2)
+    } else {
+      predictiveDiff.max(lastDiff * 2 / 3)
```
greenhat, line 65: `lastDiff * 2 / 3` is 66.66%. 50% should be `lastDiff / 2`. 
The same goes for `avg`.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -257,34 +258,77 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
...
     */
   def requiredDifficultyAfter(parent: Header): Difficulty = {
-    if (parent.height == settings.chainSettings.voting.version2ActivationHeight || parent.height + 1 == settings.chainSettings.voting.version2ActivationHeight) {
-      // Set difficulty for version 2 activation height (where specific difficulty is needed due to PoW change)
-      settings.chainSettings.initialDifficultyVersion2
+    val parentHeight = parent.height
+
+    // todo: this EIP-37 activation checking code could be removed after activation
+    if (settings.chainSettings.isMainnet &&
+        parentHeight > 843776 &&
```
greenhat, line 285: `843776` and `128` are good candidates for named constants.

greenhat (changes requested):
Looking good! Please check my comments. 
Also, I did not find any tests for the new code.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -257,34 +258,77 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
...
+      }
+    }
+
+    if (parentHeight > 843776 && parentHeight + 1 >= eip37VotedOn.getOrElse(Int.MaxValue)) {
+      // by eip37VotedOn definition could be on mainnet only
+      val epochLength = 128 // epoch length after EIP-37 activation
+      if (parentHeight % epochLength == 0) {
+        val heights = difficultyCalculator.previousHeadersRequiredForRecalculation(parentHeight + 1, epochLength)
+        // todo: if parent is on best chain, read headers directly, not via headerChainBack
+        val chain = headerChainBack(heights.max - heights.min + 1, parent, _ => false)
```
pragmaxim, line 302: What is the reasoning behind the `limit = heights.max - heights.min + 1` in `headerChainBack` ? Why to always increment it by `1` ?

src/main/scala/org/ergoplatform/mining/difficulty/DifficultyAdjustment.scala
aslesarenko: I think the intention here is to make these adjustments mutually inverse.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -257,34 +258,77 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
 
   protected def heightIdsKey(height: Int): ByteArrayWrapper = ByteArrayWrapper(Algos.hash(Ints.toByteArray(height)))
 
+
+  private val EIP37VotingParameter: Byte = 8 // todo: should be 6 after testing , and 6 = 2100 in config
```
aslesarenko, line 262: What `testing` means? Should it already be changed?

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
aslesarenko: Also 8192 and 512. The name can also document the meaning.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
```diff
@@ -257,34 +258,77 @@ trait HeadersProcessor extends ToDownloadProcessor with ScorexLogging with Score
...
-          val chain = headerChainBack(heights.max - heights.min + 1, parent, _ => false)
-          val headers = chain.headers.filter(p => heights.contains(p.height))
-          difficultyCalculator.calculate(headers)
+          parent.requiredDifficulty
         }
-      } else {
-        parent.requiredDifficulty
       }
     }
   }
```
aslesarenko, line 334: I really worry about complexity of this method. Can we make old version unchanged and be invoked only before activation? At the same the new logic to be written separately and to be invoked AFTER activation. We need to have some guarantee that before activation the old code works unchanged. And the new logic will be easier to analyze in isolation.
Also complex conditions can be encapsulated into method with descriptive names and ScalaDoc.

src/main/scala/org/ergoplatform/settings/Constants.scala
```diff
@@ -53,7 +53,7 @@ object Constants {
 
   val SoftForkEpochs = 32 //about 45.5 days
 
-  val TrueLeaf: ErgoTree = Values.TrueLeaf.toSigmaProp
+  def TrueLeaf: ErgoTree = Values.TrueLeaf.toSigmaProp
   val FalseLeaf: ErgoTree = Values.FalseLeaf.toSigmaProp
```
aslesarenko, line 57:
```suggestion
  def FalseLeaf: ErgoTree = Values.FalseLeaf.toSigmaProp
```

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/mining/difficulty/DifficultyAdjustment.scala
kushti: Yes, that was the intention, changed to / 2 though.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
kushti: done

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
kushti: See below on testing

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
kushti: I made manual testing, syncing from height 600K+ till current moment. Thus I checked how old code works before the activation. To test activation, I used parameter #8, as ~100% blocks are voting for increasing it. Thus I got activation done on my client. 

Please note, both versions will be called after activation, condition `parentHeight > 843776 && parentHeight + 1 >= eip37ActivationHeight.getOrElse(Int.MaxValue)` will be changed in the next version if activation will happen (so we can hardcode activation height).

kushti:
> Looking good! Please check my comments. Also, I did not find any tests for the new code.

See [above](https://github.com/ergoplatform/ergo/pull/1845#discussion_r981177510) on testing. Unfortunately, automated tests would be tricky, will take a look how to do them.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala
kushti: As header heights.max  counted, you need + 1 to catch heights.min. For example, if heights.max = 843776 , heights.max = 843668, you will end up  with 843669 without + 1

aslesarenko (approved):

src/main/scala/org/ergoplatform/mining/difficulty/DifficultyAdjustment.scala
```diff
@@ -33,8 +32,57 @@ class LinearDifficultyControl(val chainSettings: ChainSettings) extends ScorexLo
...
+    */
+  def eip37Calculate(previousHeaders: Seq[Header], epochLength: Int): Difficulty = {
+    require(previousHeaders.size >= 2, "at least two headers needed for diff recalc")
+    val lastDiff = previousHeaders.last.requiredDifficulty
+
+    val predictiveDiff = calculate(previousHeaders, epochLength)
+    val limitedPredictiveDiff = if (predictiveDiff > lastDiff) {
+      predictiveDiff.min(lastDiff * 3 / 2)
+    } else {
+      predictiveDiff.max(lastDiff * 2)
```
greenhat, line 66: Should it be `lastDiff / 2`? The same goes for `uncompressedDiff`.

greenhat (changes requested):
One question for diff lower bound. 
Regarding the tests for new code, maybe just test `eip37Calculate`?

src/main/scala/org/ergoplatform/mining/difficulty/DifficultyAdjustment.scala
kushti: Sorry, my bad, fixed.

For tests, thinking to add some test vectors. Anything better we can do?

pragmaxim (approved):

src/main/scala/org/ergoplatform/mining/difficulty/DifficultyAdjustment.scala
greenhat: PBT might be a good fit here.

greenhat (approved):

PR #1848, opened 2022-09-28
SIGTERM should be handled as remote shutdown

pragmaxim:
Some time ago I implemented proper `CoordinatedShutdown`  that is invoked from `/shutdown` endpoint and gracefully shuts down all services, JVM `shutdownHook` that is received on `SIGTERM` or `SIGKILL` signals should do the same thing.

### DIFF
--- a/src/main/scala/scorex/core/app/Application.scala
+++ b/src/main/scala/scorex/core/app/Application.scala
@@ -3,6 +3,7 @@ package scorex.core.app
 import akka.actor.{ActorRef, ActorSystem}
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.server.{ExceptionHandler, RejectionHandler, Route}
+import org.ergoplatform.ErgoApp
 import org.ergoplatform.settings.ErgoSettings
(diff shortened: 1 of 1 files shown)


src/main/scala/scorex/core/app/Application.scala
```diff
@@ -100,20 +98,8 @@ trait Application extends ScorexLogging {
...
         log.error("Unexpected shutdown")
-        stopAll()
+        ErgoApp.shutdownSystem()
       }
     })
   }
-
-  def stopAll(): Unit = synchronized {
-    log.info("Stopping network services")
-    upnpGateway.foreach(_.deletePort(scorexSettings.network.bindAddress.getPort))
```
kushti, line 110: this line is lost, why not needed?

kushti (changes requested):

src/main/scala/scorex/core/app/Application.scala
pragmaxim: https://github.com/ergoplatform/ergo/blob/master/src/main/scala/org/ergoplatform/ErgoApp.scala#L158

kushti (approved):

PR #1856, opened 2022-10-01
Remove invalid headers on chain repair

kushti:
Unlike previous attempts to repair db of node being stuck, this one is removing invalid blocks in the past (in 128 blocks depth from the tip) and in future (not just the next height).

Could be useful to unstuck with pre 4,0,100 databases stuck after the HF.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
@@ -260,7 +260,7 @@ object ErgoHistory extends ScorexLogging {
 
   // check if there is possible database corruption when there is header after
   // recognized blockchain tip marked as invalid
-  protected[nodeView] def repairIfNeeded(history: ErgoHistory): Boolean = history.historyStorage.synchronized {
+  protected[nodeView] def repairIfNeeded(history: ErgoHistory): Unit = history.historyStorage.synchronized {
     val RepairDepth = 128
 
     val bestHeaderHeight = history.headersHeight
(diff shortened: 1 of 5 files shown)


src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -258,21 +260,40 @@ object ErgoHistory extends ScorexLogging {
...
+          } else {
+            true
+          }
+        }
+        val updatedHeightIdsValue: Array[Byte] = notInvalidHeaders.foldLeft(Array.empty[Byte]) { case (acc, id) =>
+          acc ++ idToBytes(id)
+        }
+        if(updatedHeightIdsValue.isEmpty) {
+          //could be the case after bestHeaderHeight
+          history.historyStorage.remove(Array(history.heightIdsKey(h)), Nil)
```
pragmaxim, line 286: Shouldn't we also remove the Modifier at this height? Not just index? Or is there an idempotency in place that the modifer gets overriden if index is empty?

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -258,21 +260,40 @@ object ErgoHistory extends ScorexLogging {
...
+        } else {
+          history.historyStorage.insert(Array(history.heightIdsKey(h) -> updatedHeightIdsValue), Nil)
```
pragmaxim, line 288: I don't understand what we are updating the valid header index with and why, that would need at least a comment

aslesarenko (approved):

PR #1865, opened 2022-10-09
EIP-39: monotonic creation height rule

kushti:
In this PR, EIP-39 (https://github.com/ergoplatform/eips/pull/80) is implemented as follows:

* starting from block version #3 (5.0 soft-fork and after), the client is checking that for every output of a transaction, its creation height is not less than max creation height in inputs
* note, only inputs are considered, data inputs are not
* this rule is about soft-fork, as old nodes will continue to validate new blocks. So it will be activated in 5.0 soft-fork (can be merged with existing code even now). Also, the new rule is soft-forkable, so can be switched off (and so, replaced with another rule) via a soft-fork.

### DIFF
--- a/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
+++ b/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
@@ -376,6 +376,9 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
 
     val blockVersion = stateContext.blockVersion
 
+    // Before 5.0 soft-fork (v3 block version), we are checking only that
+    // creation height in outputs is not non-negative
+    // After, we are checking that it is not less that max creation height in inputs
     val maxCreationHeightInInputs = if (blockVersion <= Header.HardeningVersion) {
       0
     } else {
--- a/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
+++ b/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
@@ -6,6 +6,7 @@ import org.ergoplatform._
 import org.ergoplatform.http.api.ApiCodecs
 import org.ergoplatform.mining.emission.EmissionRules
 import org.ergoplatform.modifiers.ErgoNodeViewModifier
+import org.ergoplatform.modifiers.history.header.Header
 import org.ergoplatform.nodeView.ErgoContext
 import org.ergoplatform.nodeView.state.ErgoStateContext
 import org.ergoplatform.utils.ArithUtils._
@@ -157,7 +158,8 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
 
   private def verifyOutput(validationBefore: ValidationState[Long],
                            out: ErgoBox,
-                           stateContext: ErgoStateContext): ValidationResult[Long] = {
+                           stateContext: ErgoStateContext,
+                           maxCreationHeightInInputs: Int): ValidationResult[Long] = {
 
     val blockVersion = stateContext.blockVersion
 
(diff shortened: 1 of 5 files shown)


greenhat (approved):

MrStahlfelge (approved):

src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -371,6 +374,17 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
     // Cost limit per block
     val maxCost = stateContext.currentParameters.maxBlockCost.toLong
 
+    val blockVersion = stateContext.blockVersion
+
+    // Before 5.0 soft-fork (v3 block version), we are checking only that
+    // creation height in outputs is not non-negative
```
aslesarenko, line 380:
```suggestion
    // creation height in outputs is not negative
```

src/test/scala/org/ergoplatform/modifiers/mempool/ErgoTransactionSpec.scala
```diff
@@ -36,6 +36,52 @@ class ErgoTransactionSpec extends ErgoPropertyTest with ErgoTestConstants {
...
+  }
+
+  private def modifyAsset(boxCandidate: ErgoBoxCandidate,
+                          deltaFn: Long => Long,
+                          idToskip: TokenId): ErgoBoxCandidate = {
+    val assetId = boxCandidate.additionalTokens.find(t => !java.util.Arrays.equals(t._1, idToskip)).get._1
+
+    val tokens = boxCandidate.additionalTokens.map { case (id, amount) =>
+      if (java.util.Arrays.equals(id, assetId)) assetId -> deltaFn(amount) else assetId -> amount
+    }
```
aslesarenko, line 71: The search over boxCandidate.additionalTokens happens 3 times. Passing an index in additionalTokens will be both faster and cleaner.

src/test/scala/org/ergoplatform/modifiers/mempool/ErgoTransactionSpec.scala
```diff
@@ -36,6 +36,52 @@ class ErgoTransactionSpec extends ErgoPropertyTest with ErgoTestConstants {
...
+    new ErgoBoxCandidate(
+      boxCandidate.value,
+      boxCandidate.ergoTree,
+      boxCandidate.creationHeight,
+      tokens,
+      boxCandidate.additionalRegisters)
+  }
+
+  private def checkTx(from: IndexedSeq[ErgoBox], wrongTx: ErgoTransaction): Try[Int] = {
+    wrongTx.statelessValidity().flatMap(_ => wrongTx.statefulValidity(from, emptyDataBoxes, emptyStateContext))
```
aslesarenko, line 82: please add comment on why it is called `wrongTx`?

pragmaxim (approved):

src/test/scala/org/ergoplatform/modifiers/mempool/ErgoTransactionSpec.scala
kushti: This is testing code

src/test/scala/org/ergoplatform/modifiers/mempool/ErgoTransactionSpec.scala
kushti: Out of scope of this PR, renamed to tx anyway

scalahub:
> * its creation height is not less than max creation height in inputs

Do you mean "its creation height is not less *or equal to* than max creation height in inputs" ?
If not then we should be able to preserve the same (max) creation height from inputs (cc @greenhat)

greenhat:
> > * its creation height is not less than max creation height in inputs
> 
> Do you mean "its creation height is not less _or equal to_ than max creation height in inputs" ? If not then we should be able to preserve the same (max) creation height from inputs (cc @greenhat)

There are other inputs (with much higher heights) besides the pool box in the update tx.

kushti:
> > * its creation height is not less than max creation height in inputs
> 
> Do you mean "its creation height is not less _or equal to_ than max creation height in inputs" ? If not then we should be able to preserve the same (max) creation height from inputs (cc @greenhat)

not less means equal or greater

PR #1867, opened 2022-10-11
i1861 - filtering connected peers

### LINKED ISSUE #1861: Send only peers which were active recently
In response to `getPeers`, the node should send peers which were active recently, e.g. within last 24 hrs.

jellymlg:
Only send peers, which sent messages in the last 3 hours.

closes #1861

### DIFF
--- a/src/main/scala/scorex/core/network/SendingStrategy.scala
+++ b/src/main/scala/scorex/core/network/SendingStrategy.scala
@@ -7,11 +7,16 @@ trait SendingStrategy {
 }
 
 object SendToRandom extends SendingStrategy {
+
+  private val dayInMs: Long = 24 * 60 * 60 * 1000
+
   override def choose(peers: Seq[ConnectedPeer]): Seq[ConnectedPeer] = {
-    if (peers.nonEmpty) {
-      Seq(peers(Random.nextInt(peers.length)))
+    val activeRecently: Seq[ConnectedPeer] = // choose peers which were active in the last 24 hours
+      peers.filter(p => p.peerInfo.isDefined && System.currentTimeMillis - p.peerInfo.get.lastHandshake  < dayInMs)
+    if (activeRecently.nonEmpty) {
+      Seq(activeRecently(Random.nextInt(activeRecently.length)))
     } else {
-      Seq.empty
+      Nil
     }
   }
 }
--- a/src/main/scala/scorex/core/network/SendingStrategy.scala
+++ b/src/main/scala/scorex/core/network/SendingStrategy.scala
@@ -8,11 +8,11 @@ trait SendingStrategy {
 
 object SendToRandom extends SendingStrategy {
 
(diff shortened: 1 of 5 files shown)


satsen:
I recommend to do 3 hours, since this is what Bitcoin does and it's the largest network. If it works fine for it then it should be ok for us too. (https://en.bitcoin.it/wiki/Protocol_documentation#getaddr)

satsen:
Also, I don't know how the node handles P2P, but in my P2P implementation I only send handshake when the connection is established. Does this mean that connections would get filtered if stayed connected for more than x hours? I think you should either check when the TCP socket closed or the last message transfer time (sent/received).

jellymlg:
@satsen addressed comments

src/main/scala/scorex/core/network/SendingStrategy.scala
```diff
@@ -7,11 +7,16 @@ trait SendingStrategy {
...
 
 object SendToRandom extends SendingStrategy {
+
+  private val limit: Long = 3 * 60 * 60 * 1000 // 3h
+
   override def choose(peers: Seq[ConnectedPeer]): Seq[ConnectedPeer] = {
-    if (peers.nonEmpty) {
-      Seq(peers(Random.nextInt(peers.length)))
+    val activeRecently: Seq[ConnectedPeer] = // choose peers which were active in the last limit hours
+      peers.filter(System.currentTimeMillis - _.lastMessage  < limit)
```
kushti, line 15: this filter is saying that the node is not sending messages to peers not active in last 3 hours. However, they will be disconnected before, so the code is not needed. 

The issue was about filtering peers sent in response to GetPeer network message. They are chosen is SeenPeers:

```
    case class SeenPeers(howMany: Int) extends GetPeers[Seq[PeerInfo]] {

      override def choose(knownPeers: Map[InetSocketAddress, PeerInfo],
                          blacklistedPeers: Seq[InetAddress],
                          sc: ScorexContext): Seq[PeerInfo] = {
        val recentlySeenNonBlacklisted = knownPeers.values.toSeq
          .filter { p =>
            (p.connectionType.isDefined || p.lastHandshake > 0) &&
              !blacklistedPeers.exists(ip => p.peerSpec.declaredAddress.exists(_.getAddress == ip))
          }
        Random.shuffle(recentlySeenNonBlacklisted).take(howMany)
      }
    }
```

so you need to add a filter to this code (and write a test).

kushti (changes requested):

src/main/scala/scorex/core/network/SendingStrategy.scala
jellymlg: Should I use lastHandshake or lastMessage to filter peers?

jellymlg: done

kushti (approved):

Issue #1854, opened 2022-09-30
Additional mempool API endpoints


anon-br:
Keeping the mempool synced with an external service (like an explorer, for example) is not very effective and such an overhead for many reasons. So as mempool is a relativelly small in-memory data set, I think that it would be really good to be able to query the mempool directly in the node. Also, it can potentially make possible for explorers' APIs to reflect mempool in realtime  –which hopefully will be the case for [ergo-uexplorer`s](https://github.com/pragmaxim/ergo-uexplorer) GraphQL server.

Here I'm listing a set of endpoints that should be enough for a wide variety of use cases.

  * ### Transactions
    * `HEAD`: `mempool/transactions/{transactionId}` - Check if a transaction if present in mempool
    * `GET`: `mempool/transactions/byTransactionId/{transactionId}` - Get a transaction by `transactionId`
    * `GET`: `mempool/transactions/byErgoTree/{ergoTree}` - Get a list of transactions filtered by `ergoTree`, considering inputs and outputs
    
   * ### Inputs
     * `POST`: `mempool/transactions/inputs/?check={true|false}` - Batch get a list of inputs filtered by multiple `boxIds`. If `check=true` it must check if `boxIds` list elements are present on mempool as inputs and only return a list of `boxId` of included ones.
     * `GET`: `mempool/transactions/inputs/byBoxId/{boxId}` - Get an input by `boxId`;

  * ### Outputs
     * `POST`: `mempool/transactions/outputs/?check={true|false}` - Batch get a list of outputs filtered by multiple `boxIds`. If `check=true` it must check if `boxIds` list elements are present on mempool as outputs and only return a list of `boxId` of included ones.
     * `GET`: `mempool/transactions/outputs/byBoxId/{boxId}` - Get a box by `boxId`
     * `GET`: `mempool/transactions/outputs/byErgoTree/{ergoTree}` - Get a list of boxes filtered by `ergoTree`
     * `GET`: `mempool/transactions/outputs/byTokenId/{tokenId}` - Get a list of boxes filtered by `tokenId`
     * `POST`: `mempool/transactions/outputs/byRegisters` - Get a list of boxes filtered by a set of hex registers. It must filter only by defined registers and not consider `undefined` ones.
        ```ts
        // proposed registers filter param 
        {
          R4: string | undefined,
          R5: string | undefined,
          R6: string | undefined,
          R7: string | undefined,
          R8: string | undefined,
          R9: string | undefined
        }
       ```

[...]

PR #1869, opened 2022-10-12
i1854 additional mempool endpoints

pragmaxim:
Closes #https://github.com/ergoplatform/ergo/issues/1854

I added the new endpoints under existing `/transactions/unconfirmed/` endpoint as it is all searching through unconfirmed transactions ... probably better than involving the new word `mempool`.

@MrStahlfelge I implemented it all besides :
```
POST: mempool/transactions/inputs/?check={true|false}
POST: mempool/transactions/outputs/?check={true|false}
```
As they seem hard to reason about, the description : 
>> Batch get a list of inputs filtered by multiple boxIds. If check=true it must check if boxIds list elements are present on mempool as inputs and only return a list of boxId of included ones.

says : "Either return Boxes Or only BoxIds if `check=true` and they are present in mempool. I'd be happy to implement something like this but imho it should be easier to reason about.

### DIFF
--- a/src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
+++ b/src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
@@ -1,20 +1,29 @@
 package org.ergoplatform.http.api
 
 import akka.actor.{ActorRef, ActorRefFactory}
-import akka.http.scaladsl.server.{Directive, Route}
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.{Directive, Directive1, Route, ValidationRejection}
 import akka.pattern.ask
 import io.circe.Json
 import io.circe.syntax._
+import org.ergoplatform.ErgoBox.{BoxId, NonMandatoryRegisterId, TokenId}
 import org.ergoplatform.modifiers.mempool.{ErgoTransaction, UnconfirmedTransaction}
 import org.ergoplatform.nodeView.ErgoReadersHolder.{GetReaders, Readers}
 import org.ergoplatform.nodeView.mempool.ErgoMemPoolReader
 import org.ergoplatform.nodeView.mempool.HistogramStats.getFeeHistogram
-import org.ergoplatform.settings.ErgoSettings
+import org.ergoplatform.nodeView.state.{ErgoStateReader, UtxoStateReader}
+import org.ergoplatform.settings.{Algos, ErgoSettings}
 import scorex.core.api.http.ApiError.BadRequest
-import scorex.core.api.http.ApiResponse
+import scorex.core.api.http.{ApiError, ApiResponse}
 import scorex.core.settings.RESTApiSettings
+import scorex.crypto.authds.ADKey
+import scorex.crypto.hash.Digest32
+import scorex.util.encode.Base16
+import sigmastate.SType
+import sigmastate.Values.EvaluatedValue
 
 import scala.concurrent.Future
+import scala.util.Success
 
 case class TransactionsApiRoute(readersHolder: ActorRef,
                                 nodeViewActorRef: ActorRef,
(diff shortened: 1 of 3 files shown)


src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -87,4 +124,120 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
...
+
+  /** Get unconfirmed transaction by its id */
+  def getUnconfirmedTxByIdR: Route =
+    (pathPrefix("unconfirmed" / "byTransactionId") & get & modifierId) { modifierId =>
+      ApiResponse(getMemPool.map(_.modifierById(modifierId)))
+    }
+
+  /** Collect all transactions which inputs or outputs contain given ErgoTree hex */
+  def getUnconfirmedTxByErgoTreeR: Route =
+    (pathPrefix("unconfirmed" / "byErgoTree") & post & entity(as[Json])) { body =>
```
kushti, line 147: I think paging should be implemented for this query, for some trees there could be thousands of transactions

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -87,4 +124,120 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
...
+  def getUnconfirmedOutputByBoxIdR: Route =
+    (pathPrefix("unconfirmed" / "outputs" / "byBoxId") & get & boxId) { boxId =>
+      ApiResponse(
+        getMemPool.map(_.getAll.flatMap(_.transaction.outputs.filter(_.id.sameElements(boxId))).headOption)
+      )
+    }
+
+  /** Collect all tx outputs which contain given ErgoTree hex */
+  def getUnconfirmedOutputByErgoTreeR: Route =
+    (pathPrefix("unconfirmed" / "outputs" / "byErgoTree") & post & entity(as[Json])) { body =>
```
kushti, line 205: I think paging should be implemented here as well

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -87,4 +124,120 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
...
+  def getUnconfirmedOutputByTokenIdR: Route =
+    (pathPrefix("unconfirmed" / "outputs" / "byTokenId") & get & tokenId) { tokenId =>
+      ApiResponse(
+        getMemPool.map(_.getAll.flatMap(_.transaction.outputs.filter(_.additionalTokens.exists(_._1.sameElements(tokenId)))))
+      )
+    }
+
+  /** Collect all tx outputs which contain all given Registers */
+  def getUnconfirmedOutputByRegistersR: Route =
+    (pathPrefix("unconfirmed" / "outputs" / "byRegisters") & post & entity(as[Json])) { body =>
```
kushti, line 226: I think paging should be implemented here as well

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -87,4 +124,121 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
...
+  /** Collect all transactions which inputs or outputs contain given ErgoTree hex */
+  def getUnconfirmedTxsByErgoTreeR: Route =
+    (pathPrefix("unconfirmed" / "byErgoTree") & post & entity(as[Json]) & txPaging) { case (body, offset, limit) =>
+      body.as[String] match {
+        case Left(ex) =>
+          ApiError(StatusCodes.BadRequest, ex.getMessage())
+        case Right(ergoTree) =>
+          ApiResponse(
+            getMemPool.flatMap { pool =>
+              val allTxs = pool.getAll.slice(offset, offset + limit)
```
kushti, line 154: So if I am asking for 50 transactions, your code may return 5, as slicing happens before filtering. The same is true for all other new methods with paging

kushti (changes requested):

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
pragmaxim: True, slicing result should happen after filtering

kushti (approved):

Issue #1868, opened 2022-10-11
Rest endpoints `/utils/*` accept GET parameters with oversized ErgoTree/Address


pragmaxim:
`ErgoTree/Address` can be quite huge and violate HTTP limit for 2000 characters long URI ... They should be part of `POST` body.

jellymlg:
> `ErgoTree/Address` can be quite huge and violate HTTP limit for 2000 characters long URI ... They should be part of `POST` body.

good point, should I change the enpoints in #1859 ?

aslesarenko:
> is there a limit for ErgoTree size?
Yes, size of a Box is 4K bytes, which includes the ErgoTree and other registers.

abebeos:
@aslesarenko , it looks #1871 "Closes" did not function. Issue should be closable.

abebeos:
@kushti , closable?

Ergologica:
Checking before anyone spends time on this: it looks mostly done already.

`ErgoUtilsApiRoute` on master has POST variants for both routes that take an ErgoTree or an address:

- `POST /utils/address` — `validateAddressPostR`
- `POST /utils/ergoTreeToAddress` — `ergoTreeToAddressPostR`

Both are in `openapi.yaml` too (`/utils/address`, `/utils/ergoTreeToAddress`), alongside the GET forms. `/utils/hash/blake2b` was POST-only from the start.

What is left is `GET /utils/addressToRaw/{address}`, which has no POST form. A P2S address encodes a whole ErgoTree, so that one can indeed go past the 2000 character limit. The other remaining GET, `rawToAddress/{pubKeyHex}`, takes a 33-byte group element — 66 hex characters — so it is never affected.

So the question is whether the `addressToRaw` variant alone is still wanted, and whether the bounty still applies to that remainder. I am happy to send a small PR (POST route, openapi entry, a case in `UtilsApiRouteSpec`) if it is; otherwise this could probably be closed.

PR #1871, opened 2022-10-13
I1868 utils endpoint with post requests

pragmaxim:
Closes https://github.com/ergoplatform/ergo/issues/1868

The rest of the endpoints accepts only `P2PK` addresses which is fine for `GET` request.

### DIFF
--- a/src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
+++ b/src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
@@ -1,6 +1,7 @@
 package org.ergoplatform.http.api
 
 import akka.actor.ActorRefFactory
+import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Route
 import io.circe.Json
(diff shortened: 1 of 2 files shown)


src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -60,9 +60,9 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
 
   def hashBlake2b: Route = {
     (post & path("hash" / "blake2b") & entity(as[Json])) { json =>
-      json.asString match {
-        case Some(message) => ApiResponse(encoder.encode(Blake2b256(message)))
-        case None          => ApiError.BadRequest
+      json.as[String] match {
+        case Right(message) => ApiResponse(encoder.encode(Blake2b256(message)))
+        case Left(ex)       => ApiError(StatusCodes.BadRequest, ex.getMessage())
```
pragmaxim, line 65: It is better to propagate the error message to user

kushti (approved):

Issue #1870, opened 2022-10-12
Consider slicing by height in /boxes/unspent


kushti:
Add heights range in `unspentBoxes` arguments in `WalletRegistry`, and so extract only boxes in given range from DB, like done for `/wallet/transactions`.

pragmaxim:
@kushti by `inclusionHeight` or `spendingHeight` ?

dimension-drifter:
@kushti i have raised a PR for this issue. Please Review it

Ergologica:
I have a patch for this.

@pragmaxim asked above whether to slice by `inclusionHeight` or `spendingHeight` — by `inclusionHeight`: the route parameters are already named `minInclusionHeight`/`maxInclusionHeight`, the in-memory filter these endpoints apply today reads `trackedBox.inclusionHeightOpt`, the index that already exists in `WalletRegistry` is keyed by inclusion height, and for *unspent* boxes a spending height does not exist by definition.

One thing worth flagging while looking at this: it is not only a performance question. `getWalletBoxes` reads `walletUnspentBoxes(maxInputsToUse * ScanDepthFactor)`, and that index (`UnspentIndexPrefix`) is keyed by **box id, not by height**, so the 30000 limit takes an arbitrary subset. For a wallet holding more unspent boxes than that, a query for a height window can come back with an arbitrary part of it, or empty, with nothing indicating anything was dropped — and `offset`/`limit` paging on top is unstable for the same reason.

Meanwhile `WalletRegistry.unspentBoxesByInclusionHeight` already exists and already backs `/scan/unspentBoxes`; it is just not wired to the wallet path. The patch wires it, and only when a height range is actually requested — without one it keeps the current limited read, so no query becomes more expensive than it is today.

Opening a PR shortly. #2284 is open against this issue; happy to close mine if @dimension-drifter is still on it.

PR #1872, opened 2022-10-14
i1870 filter /scan/unspent boxes by inclusion height efficiently

pragmaxim:
closes https://github.com/ergoplatform/ergo/issues/1870

@kushti I need more time for this one, the KV store where the keys are composed of multiple "navigation" byte sequences are very hard to reason about.


kushti (changes requested):
It seems spent boxes are included into results.

Submitted failing test case.

I think you need to pass height into

```
  private def firstScanBoxSpaceKey(scanId: ScanId): Array[Byte] =
    composeKey(UnspentIndexPrefix, scanId, 0)

  private def lastScanBoxSpaceKey(scanId: ScanId): Array[Byte] =
    composeKey(UnspentIndexPrefix, scanId, -1)
```

pragmaxim:
@kushti The problem I encountered was backward compatibility : 
https://github.com/ergoplatform/ergo/blob/39fab7039f02278cffdfbb6b2632982254bb0c80/src/main/scala/org/ergoplatform/nodeView/wallet/persistence/WalletRegistry.scala#L519-L520

I'd have to put `height` here like this `composeKeyWithId(prefix, scanId, height, trackedBox.box.id)` but it would no longer be backward compatible, right?

kushti:
@pragmaxim oh yes, indeed my proposed solution is breaking backwards compatibility. With your implementation, reading db is not efficient for big ranges. Maybe we need to use new method for small ranges, and old one otherwise ?

pragmaxim:
@kushti on the contrary, I believe that this will be faster as we called `sortByHeight` on unordered boxes and now we load the boxes already sorted by height.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletService.scala
```diff
@@ -405,16 +407,21 @@ class ErgoWalletServiceImpl(override val ergoSettings: ErgoSettings) extends Erg
...
   }
 
-  override def getScanUnspentBoxes(state: ErgoWalletState, scanId: ScanId, considerUnconfirmed: Boolean): Seq[WalletBox] = {
+  override def getScanUnspentBoxes(state: ErgoWalletState, scanId: ScanId, considerUnconfirmed: Boolean, minHeight: Int, maxHeight: Int): Seq[WalletBox] = {
     val unconfirmed = if (considerUnconfirmed) {
-      state.offChainRegistry.offChainBoxes.filter(_.scans.contains(scanId))
+      state.offChainRegistry.offChainBoxes
+        .filter { box =>
+          box.scans.contains(scanId) &&
+          box.inclusionHeightOpt.getOrElse(0) >= minHeight  &&
```
kushti, line 415: for unconfirmed boxes, inclusion height is always defined

I think you need to return unconfirmed with old filter, and now only if maxHeight is not defined by a user

kushti (changes requested):

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletService.scala
```diff
@@ -405,16 +407,21 @@ class ErgoWalletServiceImpl(override val ergoSettings: ErgoSettings) extends Erg
     boxes.map(tb => WalletBox(tb, currentHeight)).sortBy(_.trackedBox.inclusionHeightOpt)
   }
 
-  override def getScanUnspentBoxes(state: ErgoWalletState, scanId: ScanId, considerUnconfirmed: Boolean): Seq[WalletBox] = {
+  override def getScanUnspentBoxes(state: ErgoWalletState, scanId: ScanId, considerUnconfirmed: Boolean, minHeight: Int, maxHeight: Int): Seq[WalletBox] = {
```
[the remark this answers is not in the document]
pragmaxim: It is `override `, all the `ErgoWalletService` implementations are documented in the interface methods.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletService.scala
pragmaxim: I don't understand this comment, old filter was `_.scans.contains(scanId)` and now we also want to filter out boxes unconfirmed boxes that are out of height range :thinking:

kushti: Unconfirmed boxes do not have inclusion height, they are not in the chain yet

pragmaxim: I see, we could think about it also in a way that `considerUnconfirmed` overrules the height range ...

kushti: I think we need to allow considerUnconfirmed to be set only if there's no limit for maxHeight.

kushti (approved):

Issue #1778, opened 2022-07-22
Hard-code message sets and handlers


kushti:
Currently message processing has a lot of flexibility we do not actually need for Ergo protocol client

In `NetworkController`
* eliminate `RegisterMessageSpecs`
* hard-code handlers in messageHandlers

In ScorexContext:
* Remove `messageHandlers`

Hardcode massage specs we have, and simplify processing logic based on that,

PR #1880, opened 2022-10-23
#1778 Remove RegisterMessageSpecs

ccellado:
Currently message processing has a lot of flexibility we do not actually need for Ergo protocol client. Eliminate RegisterMessageSpecs and hard-code handlers in messageHandlers.

### DIFF
--- a/src/main/scala/org/ergoplatform/ErgoApp.scala
+++ b/src/main/scala/org/ergoplatform/ErgoApp.scala
@@ -20,6 +20,7 @@ import scorex.core.api.http._
 import scorex.core.app.ScorexContext
 import scorex.core.network.NetworkController.ReceivableMessages.ShutdownNetwork
 import scorex.core.network._
+import scorex.core.network.message.Message.MessageCode
 import scorex.core.network.message._
 import scorex.core.network.peer.PeerManagerRef
 import scorex.core.settings.ScorexSettings
@@ -36,7 +37,10 @@ class ErgoApp(args: Args) extends ScorexLogging {
 
   private val ergoSettings: ErgoSettings = ErgoSettings.read(args)
 
-  require(ergoSettings.scorexSettings.restApi.apiKeyHash.isDefined, "API key hash must be set")
+  require(
+    ergoSettings.scorexSettings.restApi.apiKeyHash.isDefined,
+    "API key hash must be set"
+  )
 
   log.info(s"Working directory: ${ergoSettings.directory}")
   log.info(s"Secret directory: ${ergoSettings.walletSettings.secretStorage.secretDir}")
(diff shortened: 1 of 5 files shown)


src/main/scala/org/ergoplatform/ErgoApp.scala
```diff
@@ -96,14 +102,12 @@ class ErgoApp(args: Args) extends ScorexLogging {
       None
     }
 
-
   private val syncTracker = ErgoSyncTracker(scorexSettings.network, timeProvider)
-  private val statsCollectorRef: ActorRef = ErgoStatsCollectorRef(readersHolderRef, networkControllerRef, syncTracker, ergoSettings, timeProvider)
```
kushti, line 101: Like with ErgoNodeViewSynchronizer, we need to touch ErgoStatsCollectorRef to run it

src/main/scala/org/ergoplatform/ErgoApp.scala
```diff
@@ -112,29 +116,69 @@ class ErgoApp(args: Args) extends ScorexLogging {
...
-  if (ergoSettings.scorexSettings.network.peerDiscovery) {
-    // Launching PeerSynchronizer actor which is then registering itself at network controller
-    PeerSynchronizerRef("PeerSynchronizer", networkControllerRef, peerManagerRef, scorexSettings.network)
-  }
+  private val messageHandlers: ActorRef => Map[MessageCode, ActorRef] =
+    networkControllerRef => {
+      val ergoNodeViewSynchronizerRef = ergoNodeViewSynchronizerRefPartial(
+        networkControllerRef
+      )
+      var map: Map[MessageCode, ActorRef] = Map(
```
kushti, line 124: GetPeers missed

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -29,7 +29,8 @@ import scala.util.{Random, Try}
 class NetworkController(ergoSettings: ErgoSettings,
                         peerManagerRef: ActorRef,
                         scorexContext: ScorexContext,
-                        tcpManager: ActorRef
+                        tcpManager: ActorRef,
+                        messageHandlersPartial: ActorRef => Map[MessageCode, ActorRef]
```
kushti, line 33: There's no need in such flexibility probably, handlers can be just hardcoded in NetworkController

kushti (changes requested):

src/main/scala/org/ergoplatform/ErgoApp.scala
ccellado: This variable is untouched by PR

src/main/scala/scorex/core/network/NetworkController.scala
ccellado: The problem is that code becomes very mess, if I try to create these actors inside NetworkController.
So in my opinion partial construction works better for readability.

kushti (approved):

Issue #1822, opened 2022-08-28
Introduce per-peer rules for pulling transactions in 


kushti:
Currently incoming unconfirmed transactions flow has global limits (stored in interblockCost). As a next step, lets have per-peer limits, like
 
```
private val txInfo = mutable.Map[ConnectedPeer, IncomingTxInfo]()
```

and so do not accept a transaction from a peer if per-block limit is exceeded

PR #1891, opened 2022-10-30
1822 - Per-Peer Transaction Rules

ApexTheory:
https://github.com/ergoplatform/ergo/issues/1822


kushti (changes requested):
I think we need to have both per-block and per-peer-per-block limits

ApexTheory:
Hey @kushti , I reverted the interblock cost changes.  It now considers both Interblock and Per-Peer costs.  Not sure what to set the per-peer block limit to though.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -179,17 +189,35 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
       // should not be here, and so ReserveCostValue should not be used
       log.warn("Cost is empty in processMempoolResult")
     }
+
+    //TODO peer should always be set?  Throw exception?
+    val peer = processingResult.transaction.source.get
+    val peerTxInfo = txInfo(peer)
```
kushti, line 195: Exception can be thrown here if txInfo has no a record for `peer`

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -179,17 +189,35 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
       // should not be here, and so ReserveCostValue should not be used
       log.warn("Cost is empty in processMempoolResult")
     }
+
+    //TODO peer should always be set?  Throw exception?
+    val peer = processingResult.transaction.source.get
```
kushti, line 194: source can be empty if transaction is submitted via local API. In this case we need to skip txInfo update.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -179,17 +189,35 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+    val peerTxInfo = txInfo(peer)
+
     val cost = costOpt.getOrElse(FallbackCostValue)
     val ng = processingResult match {
-      case _: FailedTransaction => interblockCost.copy(invalidatedCost = interblockCost.invalidatedCost + cost)
-      case _: SuccessfulTransaction => interblockCost.copy(acceptedCost = interblockCost.acceptedCost + cost)
-      case _: DeclinedTransaction => interblockCost.copy(declinedCost = interblockCost.declinedCost + cost)
+      case _: FailedTransaction => {
+        interblockCost.copy(invalidatedCost = interblockCost.invalidatedCost + cost)
+        peerTxInfo.copy(invalidatedCost = peerTxInfo.invalidatedCost + cost)
```
kushti, line 201: ng would be peer-related in this branch, and others also

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -179,17 +189,35 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+        peerTxInfo.copy(acceptedCost = peerTxInfo.acceptedCost + cost)
+      }
+      case _: DeclinedTransaction => {
+        interblockCost.copy(declinedCost = interblockCost.declinedCost + cost)
+        peerTxInfo.copy(declinedCost = peerTxInfo.declinedCost + cost)
+      }
     }
 
     log.debug(s"Old global cost info: $interblockCost, new $ng, tx processing cache size: ${txProcessingCache.size}")
     interblockCost = ng
```
kushti, line 214: but then you put ng into interblock (all peers) data

kushti (changes requested):

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -109,6 +108,11 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     */
   private val MempoolCostPerBlock = 12000000
 
+  /**
+    * Max cost of per-peer transactions we are going to process between blocks
+    */
+  private val MempoolPeerCostPerBlock = 12000000
```
ApexTheory, line 114: This should be lowered I assume.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -179,17 +189,37 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+        log.debug(s"Old peer ${peer.connectionId} cost info: ${peerTxInfo.totalCost}, " +
+          s"new: $newPeerCost, tx processing cache size: ${txProcessingCache.size}")
+        perPeerCost.put(peer, newPeerCost)
+      case _ => log.debug("No peer set, perPeerCost not updated.")
+    }
 
-    if (interblockCost.totalCost < MempoolCostPerBlock) {
+    if ((peerOpt.isDefined &&
+      perPeerCost.getOrElse(peerOpt.get, IncomingTxInfo.empty()).totalCost < MempoolPeerCostPerBlock) ||
+      interblockCost.totalCost < MempoolCostPerBlock) {
```
ApexTheory, line 222: Maybe a bit overkill here as the perPeerCost should always be set when the peer is defined.  But I figured it's better to continue gracefully in case of any edge cases.

Issue #1893, opened 2022-11-01
Implicit unlock after wallet initialization fails


pragmaxim:
When wallet is newly created then we call automatically `Unlock` with given wallet password 
https://github.com/ergoplatform/ergo/blob/978a170d45c08f8e0aaf051b93c965c40f47259d/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala#L101

But now it throws : 
```
10:40:53.351 INFO  [ergoref-api-dispatcher-10] o.e.n.w.ErgoWalletActor - Wallet is initialized                
10:40:53.352 INFO  [ergoref-api-dispatcher-10] o.e.n.w.ErgoWalletActor - Unlocking wallet                                            
10:40:53.363 WARN  [ergoref-api-dispatcher-10] o.e.n.w.ErgoWalletActor - Wallet unlock failed with:                                                
java.lang.RuntimeException: SecretString already erased                                                                                            
        at org.ergoplatform.wallet.interface4j.SecretString.checkErased(SecretString.java:29)                  
        at org.ergoplatform.wallet.interface4j.SecretString.getData(SecretString.java:55)                                                                                                                                                                                                             
        at org.ergoplatform.wallet.secrets.JsonSecretStorage.$anonfun$unlock$6(JsonSecretStorage.scala:68)                                                                                                                                                                                            
        at scala.util.Success.flatMap(Try.scala:251)                                                                                               
        at org.ergoplatform.wallet.secrets.JsonSecretStorage.$anonfun$unlock$1(JsonSecretStorage.scala:67)                                                                                                                                                                                            
        at scala.util.Either$RightProjection.map(Either.scala:713)                                                                                 
        at org.ergoplatform.wallet.secrets.JsonSecretStorage.unlock(JsonSecretStorage.scala:58)                                                    
        at org.ergoplatform.nodeView.wallet.ErgoWalletServiceImpl.unlockWallet(ErgoWalletService.scala:344)                                        
        at org.ergoplatform.nodeView.wallet.ErgoWalletActor$$anonfun$org$ergoplatform$nodeView$wallet$ErgoWalletActor$$loadedWallet$1.applyOrElse(ErgoWalletActor.scala:322)                                                                                                                          
        at akka.actor.Actor.aroundReceive(Actor.scala:537)                                                                                                                                                                                                                                            
        at akka.actor.Actor.aroundReceive$(Actor.scala:535)              
[...]

PR #1894, opened 2022-11-01
i1893 do not erase password before Unlock

pragmaxim:
Closes https://github.com/ergoplatform/ergo/issues/1893

Wallet flow is as follows : 
 - wallet `Created/Restored` -> automatically `Unlocked` with password given during `Creation/Restoration`

Problem : 
 - we automatically Erase password in the `Create/Restore` method and it is already Erased for `Unlock`

Solution : 
 - avoid Erasing password during `Create/Restore` as `Unlocking` ALWAYS follows implicitly without user having to do so
 - Erase password during `Unlock` OR if `Create/Restore` fails

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -314,15 +314,16 @@ class ErgoWalletActor(settings: ErgoSettings,
           sender() ! Failure(new Exception("Wallet not initialized"))
       }
 
-    case UnlockWallet(encPass) =>
+    case UnlockWallet(walletPass) =>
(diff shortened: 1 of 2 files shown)


src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
```diff
@@ -320,6 +322,7 @@ class ErgoWalletActor(settings: ErgoSettings,
           context.become(loadedWallet(newState))
           sender() ! Success(())
```
kushti, line 323: I guess we need to erase pass in successful path also ?

kushti (changes requested):

kushti (approved):

PR #1901, opened 2022-11-08
API methods to send and check transaction bytes 

kushti:


src/main/resources/api/openapi.yaml
```diff
@@ -2817,6 +2817,33 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiError'
 
+  /transactions/bytes:
+      post:
+        summary: Submit an Ergo transaction given as hex-encoded transaction bytes to unconfirmed pool to send it over the network
+        operationId: sendTransactionBytes
```
pragmaxim, line 2823: I would call it `sendTransactionAsBytes` as this will be visible in swagger and it is self-explanatory

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -68,6 +69,7 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
       getUnconfirmedTransactionsR ~
       unconfirmedContainsR ~
       sendTransactionR ~
+      sendTransactionBytesR ~
```
pragmaxim, line 72: Possibly `sendTransactionAsBytes` ... `sendTransactionBytes` looks as if we are sending bytes of a transaction

src/main/resources/api/openapi.yaml
kushti: Makes sense, renamed

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
kushti: renamed

kushti:
@pragmaxim comments addressed, please make another pass!

pragmaxim (approved):

PR #1911, opened 2022-11-26
i1909 - attempting debug log rest api requests

pragmaxim:
WIP


build.sbt
```diff
@@ -60,6 +60,7 @@ libraryDependencies ++= Seq(
   "com.typesafe.akka" %% "akka-http" % akkaHttpVersion,
   "com.typesafe.akka" %% "akka-parsing" % akkaHttpVersion,
   "com.typesafe.akka" %% "akka-stream" % akkaVersion,
+  "com.typesafe.akka" %% "akka-slf4j" % akkaVersion,
```
kushti, line 63: Is the new dependency needed for purposes of this PR ?

pragmaxim: And the configuration section I added should be always present when SLF4J

PR #1934, opened 2023-01-14
Fix #1913

### LINKED ISSUE #1913: Lower log levels than INFO are working for log file but not for stdout

ccellado:
Fixes loglevel change, removed MDC\turbofilter based solution


kushti:
Tested, works for both console and file outputs

PR #1943, opened 2023-01-28
NTP removed

kushti:
In this protocol, NTP removed from the reference client. Now it is job of OS and user to make clocks synced. This simplifies the code and makes syncing more secure (as modern OS could do this job better than simple implementation in the Ergo client)


ccellado (approved):

greenhat (approved):

pragmaxim (approved):
It might be worth introducing a `TimeSupport` trait with `System.currentTimeMillis()` method and all time related helper method that are used throughout the codebase

PR #1947, opened 2023-02-03
Utxo set bootstrapping code part 3

kushti:
In this PR:

* SnapshotsInfo entity which is a container for AVL trees the node holds, and tests for it

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
@@ -10,7 +10,16 @@ import scorex.util.ScorexLogging
 
 trait ErgoStateReader extends NodeViewComponent with ScorexLogging {
 
+  /**
+   * Root hash and height of AVL+ tree authenticating UTXO set
+   */
   def rootHash: ADDigest
+
+  /**
+    * Current version of the state
+    * Must be ID of last block applied
+    */
+  def version: VersionTag
 
   val store: LDBVersionedStore
   val constants: StateConstants
@@ -18,6 +27,13 @@ trait ErgoStateReader extends NodeViewComponent with ScorexLogging {
   private lazy val chainSettings = constants.settings.chainSettings
 
   protected lazy val votingSettings: VotingSettings = chainSettings.voting
+
+  /**
+    * If the state is in its genesis version (before genesis block)
+    */
+  def isGenesis: Boolean = {
+    rootHash.sameElements(constants.settings.chainSettings.genesisStateDigest)
+  }
 
   def stateContext: ErgoStateContext = ErgoStateReader.storageStateContext(store, constants)
 
@@ -27,9 +43,6 @@ trait ErgoStateReader extends NodeViewComponent with ScorexLogging {
   def parameters: Parameters = stateContext.currentParameters
 
   def genesisBoxes: Seq[ErgoBox] = ErgoState.genesisBoxes(chainSettings)
-
-  //must be ID of last applied modifier
-  def version: VersionTag
 
   def closeStorage(): Unit = {
     log.warn("Closing state's store.")
--- a/src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
@@ -8,12 +8,16 @@ import scorex.crypto.hash.Digest32
 import scorex.db.LDBVersionedStore
 import scorex.util.ScorexLogging
 
+/**
+  * State-related data and functions related to any state implementation ("utxo" or "digest") which are
+  * not modifying the state (so only reading it)
+  */
 trait ErgoStateReader extends NodeViewComponent with ScorexLogging {
 
   /**
    * Root hash and height of AVL+ tree authenticating UTXO set
    */
-  def rootHash: ADDigest
+  def rootDigest: ADDigest
 
   /**
     * Current version of the state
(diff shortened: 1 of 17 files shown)


src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
```diff
@@ -10,15 +10,31 @@ import scorex.util.ScorexLogging
 
 trait ErgoStateReader extends NodeViewComponent with ScorexLogging {
 
+  /**
+   * Root hash and height of AVL+ tree authenticating UTXO set
+   */
   def rootHash: ADDigest
```
aslesarenko, line 16: Please what the `height` in the description mean. This field represents only hash.

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
```diff
@@ -0,0 +1,36 @@
+package org.ergoplatform.nodeView.state
+
+import org.ergoplatform.ErgoLikeContext.Height
+import org.ergoplatform.nodeView.state.UtxoState.ManifestId
+
+/**
+  * Container for UTXO set snapshots the node holds
+  * @param availableManifests - available UTXO set snapshot manifests and corresponding heights
+  */
```
aslesarenko, line 9: What is the key in the map? It doesn't have ScalaDoc

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
```diff
@@ -0,0 +1,36 @@
...
+    */
+  def nonEmpty: Boolean = availableManifests.nonEmpty
+}
+
+object SnapshotsInfo {
+
+  /**
+    * @return create container with no snapshots there
+    */
+  def makeEmpty(): SnapshotsInfo = SnapshotsInfo(Map.empty)
```
aslesarenko, line 30: Can be made val as it is immutable.

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
```diff
@@ -0,0 +1,36 @@
+package org.ergoplatform.nodeView.state
+
+import org.ergoplatform.ErgoLikeContext.Height
+import org.ergoplatform.nodeView.state.UtxoState.ManifestId
+
+/**
+  * Container for UTXO set snapshots the node holds
+  * @param availableManifests - available UTXO set snapshot manifests and corresponding heights
+  */
+case class SnapshotsInfo(availableManifests: Map[Height, ManifestId]) {
```
aslesarenko, line 10: I suggest to avoid using arrays (which is runtime representation of ManifestId) directly in case classes. Arrays don't have equality, as result equality of case class becomes broken.
As an alternative it is better to use Seq[Byte] (or TaggedType over it) or better yet implement ManifestId as a wrapper over Digest32.

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
kushti: It is not used as case class field here though

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
kushti: height, it is covered in the scaladoc

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
kushti: Height of the tree as ScalaDoc says

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
aslesarenko: It is used in the Map, and hence comparing two maps is broken.

greenhat (approved):

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
aslesarenko: It is not clear what height has to do with this property, which is only a hash.

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
```diff
@@ -10,15 +10,31 @@ import scorex.util.ScorexLogging
 
 trait ErgoStateReader extends NodeViewComponent with ScorexLogging {
```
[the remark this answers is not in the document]
kushti: Added ScalaDoc

src/main/scala/org/ergoplatform/nodeView/state/ErgoStateReader.scala
kushti: renamed to rootDigest

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
kushti: done

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsInfo.scala
kushti: reworked into class , as equality is not needed

aslesarenko (approved):

PR #1950, opened 2023-02-09
I1050 indexing - comments addressed

jellymlg:
fixed a few mistakes

### DIFF
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -281,7 +281,7 @@ components:
           type: boolean
           example: false
         globalIndex:
-          description: The global index of the box
+          description: Global index of the output in the blockchain
           type: integer
           format: int64
           minimum: 0
@@ -484,7 +484,7 @@ components:
           format: int32
           example: 3
         globalIndex:
-          description: index of the box globally
+          description: Global index of the transaction in the blockchain
           type: integer
           format: int64
           example: 3565445
@@ -5780,7 +5780,7 @@ paths:
                       $ref: '#/components/schemas/IndexedErgoTransaction'
                   total:
                     type: integer
-                    description: Total count of transactions
+                    description: Total count of retreived transactions
         '404':
           description: No transactions found for wanted address
           content:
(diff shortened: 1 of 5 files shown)


src/main/resources/api/openapi.yaml
```diff
@@ -281,7 +281,7 @@ components:
           type: boolean
           example: false
         globalIndex:
-          description: The global index of the box
+          description: Global index of the output in the blockchain
```
aslesarenko, line 284: Still not clear what this index means. Has it anything to do with block height?

src/main/scala/org/ergoplatform/http/api/BlockchainApiRoute.scala
```diff
@@ -30,6 +30,9 @@ case class BlockchainApiRoute(readersHolder: ActorRef, ergoSettings: ErgoSetting
 
   private val paging: Directive[(Int, Int)] = parameters("offset".as[Int] ? 0, "limit".as[Int] ? 5)
 
+  /**
+    * Total number of boxes/transactions that can be requested at once
+    */
   private val MaxItems = 16384
```
aslesarenko, line 36: ok, but why this exact number is maximum. What is you motivation for choosing it?

aslesarenko (changes requested):

src/main/scala/org/ergoplatform/http/api/BlockchainApiRoute.scala
jellymlg: MaxHeaders is set to this number in BlocksApiRoute to "avoid too heavy requests", so I thought it is reasonable to use as a limit

src/main/resources/api/openapi.yaml
jellymlg: The indexer runs from block 1 to the current height.
In every block from transaction 1 to the total number of transactions in that block.
In every transaction from output 0 to the total number of outputs in that transaction.
Every output processed increases globalIndex by 1.
So the globalIndex could be thought of as a kind of 'serial number' for boxes. And also the total number of boxes ever created.

aslesarenko: Perfect, this text should be somewhere in the code and referred in descriptions.

src/main/scala/org/ergoplatform/http/api/BlockchainApiRoute.scala
aslesarenko: Again, this explanation better be in ScalaDoc.

aslesarenko (approved):

PR #2028, opened 2023-08-22
Update README with info on running the node in different modes

kushti:


glasgowm148 (approved):
Pushed some changes - LGTM now!

PR #2042, opened 2023-10-02
Add bulk fetching for mempool boxes

### LINKED ISSUE #2037: Bulk endpoint for /utxo/withPool/byId
Currently when fetching the entire mempool the inputs of those transactions only contain the boxId.

When trying to parse the entire mempool (in my case for SigmaExplorer) we need to request all input boxes one by one currently.
Considering transactions can contain many hundreds of input boxes this makes it impossible to retrieve them all in a timely way.

A new endpoint for fetching unspent boxes by POST'ing an array of boxIds would speed up things enormously.
Considering this only does a UTXO and mempool scan this should be fairly light load on the node side.

We could also consider this for other such endpoints if there is any demand.

ccellado:
Resolve #2037

### DIFF
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -5106,10 +5106,10 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiError'
 
-  /utxo/withPool/byIdsList:
+  /utxo/withPool/byIds:
     post:
-      summary: Get box contents for a list of boxes by a unique identifier, from UTXO set and also the mempool.
-      operationId: getBoxWithPoolByIdsList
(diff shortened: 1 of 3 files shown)


src/main/resources/api/openapi.yaml
```diff
@@ -5106,6 +5106,42 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiError'
 
+  /utxo/withPool/byIdsList:
+    post:
+      summary: Get box contents for a list of boxes by a unique identifier, from UTXO set and also the mempool.
```
kushti, line 5111: Get boxes JSONs for ids in array provided, from UTXO or the mempool.

kushti (approved):

PR #2043, opened 2023-10-02
Add bulk fetching for blocks by headerIds

### LINKED ISSUE #2039: Bulk endpoint for /blocks/{headerId}
We can use /blocks/chainSlice to fetch multiple block headers in a single call but there is no way to fetch multiple full blocks in a single call.

Same idea as I proposed in #2037 a new endpoint where we can POST an array of headerIds and receive an array of full blocks would be helpfull!

ccellado:
Resolve #2039


src/main/resources/api/openapi.yaml
```diff
@@ -2721,6 +2721,42 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiError'
 
+  /blocks/headerIdsList:
```
kushti, line 2724: /blocks/byHeaderIds is better

src/main/resources/api/openapi.yaml
```diff
@@ -2721,6 +2721,42 @@ paths:
...
+          required: true
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+      responses:
+        '200':
+          description: List of block objects representing the full block data
```
kushti, line 2740: Full blocks corresponding to ids provided

kushti (approved):

Issue #2047, opened 2023-10-05
Limit logs disk usage 


kushti:
Store only one week  of logs by default

PR #2054, opened 2023-10-27
Change logging maxHistory to 1 week

ccellado:
Related #2047


kushti:
> Do we want to change
> 
> ```
> <maxFileSize>500MB</maxFileSize>
> <totalSizeCap>1GB</totalSizeCap>
> ```
> 
> too?

I think those settings are ok

PR #2055, opened 2023-10-30
Input/Ordering Block Implementation

kushti:
Tests coverage todo:

* test that transactions in locally as well as remotely generated input blocks are cleared from the mempool 
* test that transactions in locally as well as remotely generated input blocks are accounted for in the wallet (spent in input blocks boxes cant be spent again, boxes created in input blocks can be spent ) 
* test mempool and wallet processing on inputs chain fork (when support implemented)


papers/subblocks.md
```diff
@@ -0,0 +1,108 @@
...
+Sub-Blocks And Transactions Propagation
+---------------------------------------
+
+Let's consider that new block is just generated. Miners A and B (among others) are working on a new block. Users are
+submitting new unconfirmed transactions at the same time to the p2p network, and eventually they are reaching miners 
+(including A and B, but at a given time a transaction could be in one of the mempools just, not necessarily both, it 
+could also be somewhere else and not known to both A and B).
+
+Then, for example, miner A is generating a sub-block committing to new transactions after last block. It sends sub-block
+header as well as weak transaction ids (6 bytes hashes) of transactions included into this sub-block but not previous
```
ross-weir, line 57: Could be worth detailing weak transaction ids a little bit, why 6 bytes of hashes specifically?

flysky9981:
when the sub-block time is 1 second, high probability to produce many orphan blocks, is it suitable to still use the longest chain rule? or consider using GHOST algo, greedy heaviest observed sub-tree, to choose the chain, but the security of this chain is also secured by ergo main-chain, so just a suggestion.

Issue #1965, opened 2023-03-02
Extract NiPoPoW verifier into a separate library for SPV workflow in appkit


greenhat:
Move `NipopowProof`, `NipopowVerifier`, and all dependent types from the node code into a new library; The goal is to be able to use them for SPV workflow in appkit https://github.com/ergoplatform/ergo-appkit/issues/187

The above NiPoPoW-related types would better be in a dedicated ergo-nipopow library than in ergo-wallet since they are not directly related to wallet features. However, the bigger challenge will be the dependent more low-level types like `Header`, etc. They might have a better fit to their separate lib (ergo-core/ergo-chain-types?).

PR #2060, opened 2023-11-11
Extract NiPoPow

ccellado:
See #1965 

Classes moved from src -> ergo-core/src

Some refactoring decoupling has to be made:

```
Extraction:
    - From mining/CandidateGenerator
      - deriveUnprovenHeader into mining/CandidateUtils
    - From reemission/ReimissionRules
      - Extract object into ReemissionRulesUtils
    - From ErgoSettings
      - Extract object into ErgoSettingsReader (stay in ergo-node)
    - From ErgoMempool
      - Extract object types into ErgoMempoolTypes
    - From nodeView/history/extra/ExtraIndexer
      - Extract trait ExtraIndex
    - From ErgoNodeViewSynchronizer
      - Extract ReseiveableMessages into ErgoNodeViewSynchronizerMessages
    - From BlockchainApiRoute
      - Extract SortDirection
    - From ErgoApp
      - Extract CriticalSystemException
    - From ErgoNodeViewHolder
      - Extract LocallyGeneratedModifier
  ```


kushti:
@ccellado so if the main motivation for the `core` module I guess is to extract to there functionality needed for building alternative clients just. Then I guess node-specific parts should not be there, only consensus-related and p2p-related objects and functions ,so following parts should be moved away from it:

* /history/extra , /history/storage and all the classes in /history but ErgoSyncInfo (which is used in p2p) 
* nodeview/mempool (aside of TransactionMembershipProof)
* /http/api/requests, /scorex/api/http/ApiError 
* /scorex/core/network/peer, ConnectedPeer, ConnectionDirection 

From org.ergoplatform.settings, makes sense to have ChainSettings, ValidationRules, Parameters, PeerFeatureDescriptors, removing NodeSettings etc

kushti (approved):

Issue #2034, opened 2023-09-10
Implement UTXO set scan


kushti:
Implement a scan process for UTXO set , in addition to full blocks scan. This will allow to have scans and wallet having unspent outputs after bootstrapping with UTXO set.

shubham5080:
deos this issue still open open to work ?

###

kushti:
ah, there is some implementation already, need to check what is there (last time it wasnt complete)

PR #2072, opened 2023-12-05
I2034 utxo set scan

jellymlg:
Added Utxo set scanner, which scans unspent boxes for the wallet when bootstrapping with Utxo set snapshot.

Closes https://github.com/ergoplatform/ergo/issues/2034


kushti (changes requested):
1. Please add ScalaDoc to all the non-private functions (such as iterateAVLTree, it is complex by both signature and implementation) and classes (including signals like StartUtxoSetScanWithHistory and StartUtxoSetScan)
2. What is difference between StartUtxoSetScanWithHistory and StartUtxoSetScan ?
3. How is it tested? 
4. Please do tests for iterateAVLTree at least, like done for `dumpSnapshot` (see `property("dumping snapshot")`)

Issue #2051, opened 2023-10-13
Add limit and offset parameter to /wallet/boxes/unspent


mgpai22:
This is required to speed up the process of getting wallet boxes. Currently this endpoint returns all boxes a wallet contains. This is an expensive call.

PR #2074, opened 2023-12-07
#2051 Add limit and offset parameter to /wallet/boxes/unspent

anon-yum:
This PR addresses:
https://github.com/ergoplatform/ergo/issues/2051

and adds limit and offset to wallet/boxes/unspent, using default value of 50 for limit and max value of 500 for limit. Includes basic parameter checks and uses slice to apply offset and limit


kushti (approved):
LGTM but scan/ methods descriptions not updated in openapi.yaml

PR #2085, opened 2023-12-21
Add do-release bash script

### LINKED ISSUE #2083: Release script
Make a shell script `do-release.sh`, which is:

* checking Java version (`java -version` , should be 1.8.x)
* removing /target folder
* calling `sbt clean assembly`
* extracting version from .jar file in `target/scala-2.12/ergo-xxx.jar'
* checking that the version in `src/main/resources/application.conf` (e.g. `appVersion = 5.0.16`) is corresponding to the jar version, if not, remove the jar file and exit with error
* checking that version (`info/version`) in `src/main/resources/api/openapi.yaml` and `src/main/resources/api/openapi-ai.yaml` is  is corresponding to the jar version, if not, remove the jar file and exit with error

ross-weir:
closes #2083

### DIFF
--- a/do-release.sh
+++ b/do-release.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-required_java_version="17"
+required_java_version="1.8"
 java_version=$(java -version 2>&1 | awk -F '"' '/version/ {print $2}')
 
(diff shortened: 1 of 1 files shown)


kushti:
I think printing one error just is ok in this scenario

Issue #2106, opened 2024-01-31
/blocks/chainSlice testnet bug


mgpai22:
This mainnet request loads all 1000 elements containing header info:

http://213.239.193.208:9053/blocks/chainSlice?fromHeight=0&toHeight=1000

This testnet request only loads 256 elements at most:

http://213.239.193.208:9052/blocks/chainSlice?fromHeight=0&toHeight=1000
https://tn-ergonode-api.ergohost.io/blocks/chainSlice?fromHeight=0&toHeight=1000

This has been tested with latest node version and previous old ones too.

PR #2110, opened 2024-02-05
Fix for #2106 : MaxHeaders constant is used instead of epochLength related value in chainSlice limit

kushti:


mgpai22:
looks good!

Issue #2127, opened 2024-02-27
[BUILD] Full Build within IntelliJ Fails (NoSuchMethodError)


stenolog:
### Reproduction

* load the ergo project into IntellJ IDEA
* Build or Rebuild

```sh
scala: ## Exception when compiling 194 sources to P:\work\ergo\target\scala-2.12\it2-classes
java.lang.BootstrapMethodError: java.lang.NoSuchMethodError: scala.collection.immutable.List.$anonfun$flatMap$1$adapted(Lscala/runtime/BooleanRef;Lscala/runtime/ObjectRef;Lscala/runtime/ObjectRef;Ljava/lang/Object;)Ljava/lang/Object;
scala.async.internal.AnfTransform$_anf$2$.$anonfun$_transformToList$4(AnfTransform.scala:228)
scala.async.internal.AnfTransform$trace$2$.apply(AnfTransform.scala:43)
scala.async.internal.AnfTransform$_anf$2$._transformToList(AnfTransform.scala:174)
scala.async.internal.AnfTransform.anfLinearize$1(AnfTransform.scala:390)
scala.async.internal.AnfTransform.$anonfun$anfTransform$1(AnfTransform.scala:402)
scala.reflect.macros.contexts.Internals$$anon$1$HofTypingTransformer.transform(Internals.scala:49)
scala.reflect.macros.contexts.Internals$$anon$1.$anonfun$typingTransform$1(Internals.scala:56)
scala.tools.nsc.transform.TypingTransformers$TypingTransformer.atOwner(TypingTransformers.scala:43)
scala.reflect.macros.contexts.Internals$$anon$1.typingTransform(Internals.scala:56)
scala.async.internal.AnfTransform.anfTransform(AnfTransform.scala:53)
scala.async.internal.AnfTransform.anfTransform$(AnfTransform.scala:19)
scala.async.internal.AsyncMacro$$anon$1.anfTransform(AsyncMacro.scala:6)
scala.async.internal.AsyncTransform.asyncTransform(AsyncTransform.scala:25)
scala.async.internal.AsyncTransform.asyncTransform$(AsyncTransform.scala:12)
scala.async.internal.AsyncMacro$$anon$1.asyncTransform(AsyncMacro.scala:6)
scala.async.internal.AsyncBase.asyncImpl(AsyncBase.scala:48)
[...]
```

### Environment

Windows, ubuntu, different JDKs

kushti:
[...]

PR #2128, opened 2024-02-27
Fix for #2127 : fix for Scala 2.12 compilation failing w. NoSuchMethodError

kushti:
Scala 2.12 dependency downgraded to 2.12.11 due to regression bug in 2.12.12 https://github.com/scala/bug/issues/12100

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,7 @@ logLevel := Level.Debug
 
 // this values should be in sync with ergo-wallet/build.sbt
 val scala211 = "2.11.12"
-val scala212 = "2.12.18"
+val scala212 = "2.12.11"
 val scala213 = "2.13.12"
 
 lazy val commonSettings = Seq(
--- a/ergo-core/src/main/scala/org/ergoplatform/http/api/ApiCodecs.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/http/api/ApiCodecs.scala
@@ -34,7 +34,6 @@ import sigmastate.eval.Extensions.ArrayOps
 import sigmastate.utils.Helpers._
 
 import java.math.BigInteger
-import scala.annotation.nowarn
 import scala.util.{Failure, Success, Try}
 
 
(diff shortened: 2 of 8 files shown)


build.sbt
```diff
@@ -5,7 +5,7 @@ logLevel := Level.Debug
 
 // this values should be in sync with ergo-wallet/build.sbt
```
stenolog, line 6: what about this?

stenolog:
> Scala 2.12 dependency downgraded to 2.12.11

Would it be preferable to fix this within the `it2` code? (e.g. keep 2.12.18 and adjust code) ?

### Testing

It seems the `it2` tests need at minimum an `it2:compile` within the CI, to catch bugs/regressions.

stenolog:
The big problem is that this very basic documentation:

- checkout code
- open within IntelliJ IDEA
- use the IDE build

...is broken.

It needs an immediate fix, independent of any ongoing work.

stenolog:
I guess it could be enough for now to exclude `it2` from the `global build` that IDEA executes.

build.sbt
```diff
@@ -5,7 +5,7 @@ logLevel := Level.Debug
...
 val scala211 = "2.11.12"
-val scala212 = "2.12.18"
+val scala212 = "2.12.11"
```
stenolog, line 8: prior to 2.12.18 it was 2.12.10, possibly revert to this.

Changing commit: https://github.com/ergoplatform/ergo/commit/7865dd4ce99ece9c06b8beebe555b05086e1af4a#diff-57d3c344906eae6433d7decac6004ce19dcd27a73c6cda7f59ef95939ffe63b7

stenolog: > thing is addressed already

Could you send a PR with the relevant changes to avoid the reversion?

Or at least clarify what the concrete fix is?

stenolog: > updating `scala-async` does the trick

Great!

stenolog: @ccellado , tried scala-async update + `-Xasync`, but can't get it to work.

Fetched your `docker-fix` branch: a similar error (within IntelliJ)

```sh
P:\work\ergo\src\it\scala\org\ergoplatform\it\DeepRollBackSpec.scala:52:44
The async requires the compiler option -Xasync (supported only by Scala 2.12.12+ / 2.13.3+)
    val result: Future[Unit] = Async.async {
```

(cli complile via `sbt it:compile it2:compile` works though)

Issue #2093, opened 2024-01-05
[API Request] /blockchain/box/unspent/byAddress Unconfirmed Query Parameter


mgpai22:
Currently, this endpoint returns all confirmed unspent boxes residing in an address. A parameter is needed to also include unconfirmed boxes. I think it should include boxes with ergotrees belonging to the respective address queried.

The use case for this would be to generate chained transactions where offchain service does not know prior box(es) used. For example when ergo mobile connects with a dapp to submit a transaction, the dapp uses unspent boxes by address to retrieve input boxes. If a ergo mobile user wants to interact several times with a dapp in a short period of time it will not be possible due to double spend errors. Having the unconfirmed addition will allow for dapps to have quick successive transactions for mobile users.

ApexTheory:
I can take this.

mgpai22:
@ApexTheory It is already implemented by @jellymlg . However, I also need a flag such that it returns boxes which are not currently spent in the mempool. 

For example, let's say Address A has only one box, box A. I send box A to Address B. The outputs are two boxes, one that goes to B, and one that is change, which goes to A. This transaction is now unconfirmed.

Right now, I want to send to Address C. If I use the api to get boxes at Address A, it will return two boxes. The original box which is currently being spent and the change box which is available for use.

The flag required should only return boxes which are not in the process of being spent. In other words, none of the boxes should be in the input of transactions in the mempool.

@kushti not sure if the bounty applies for this.

kushti:
@mgpai22 @ApexTheory yet, it applies. Where and what exactly @jellymlg has implemented ?

mgpai22:
@kushti the parameter to include unconfirmed boxes `includeUnconfirmed` has already been implemented by @jellymlg 

Please see [here](http://213.239.193.208:9053/swagger#/blockchain/getBoxesByAddressUnspent)

PR #2131, opened 2024-03-01
Add excludeMempoolSpent Query Parameter to /blockchain/box/unspent/byAddress

anon-yum2:
PR addressing issue: https://github.com/ergoplatform/ergo/issues/2093
by default excludeMempoolSpent set to false to not impact any current uses of the endpoint.


kushti (approved):

anon-yum2:
Since filtering out UTXOs after receiving them from mempool may result in less than limit amount of UTXOs being returned (when there are more available), I have added recursive fetching for more UTXOs until the limit is hit or no more UTXOs are available. Will only fetch 200 a limit of 200 UTXOs at a time.

kushti:
@anon-yum2 compilation is failing now

kushti (approved):

kushti:
import scorex.util.{ModifierId, bytesToId} => import scorex.util.{ModifierId, bytesToId, idToBytes}

PR #2136, opened 2024-03-09
[DEMO] V5.0.21 rebased

stenolog:
re https://github.com/ergoplatform/ergo/issues/2137

Demonstration of how simple it is to create PRs with a clean history.

This would replace:

- https://github.com/ergoplatform/ergo/pull/2111


kushti (changes requested):
I do not understand why you did it for this repo only, and not for 5.0.14 upcoming release in sigma. Makes sense to start with upstream dependencies

stenolog:
> I do not understand why you did it for this repo only,

This is just a demo, to showcase the difference (how would it look if everyone would rebase).

PR #2167, opened 2024-06-21
Fix for duplicates in /blockchain/boxByAdress results

### LINKED ISSUE #2166: Duplicates in /blockchain/box/byErgoTree results
/blockchain/box/byErgoTree results can contain duplicates , as well as /blockchain/box/byErgoTree , /blockchain/box/unspent etc

Can be checked by following Python script:

```
import requests
import json


all_items = []

def fetch_all_items(url, headers, data, total_items=None):
  """
  Fetches all items from the paginated API and returns a combined list.

  Args:
      url (str): The base URL of the API endpoint.
      headers (dict): The headers to include in the request.
      data (str): The data to send in the request body.
      total_items (int, optional): The total number of items returned by the API (if known). Defaults to None.

  Returns:
      list: A list containing all items retrieved from the paginated API.
  """
  offset = 0
  limit = 100

  if total_items is None:
    # Fetch the first page to get the total count
    response = requests.post(url, headers=headers, data=json.dumps(data))
    response.raise_for_status() # Raise an error for non-200 status codes
    data1 = response.json()
    total_items = data1.get("total", 0)

  while offset < total_items:
    # Construct the query parameters with current offset and limit
    params = {"offset": offset, "limit": limit}
    response = requests.post(url, headers=headers, data=data, params=params)
    print(response.status_code)

    data2 = response.json()
    items = data2.get("items", [])
    all_items.extend(items)
    offset += limit

  return all_items

def find_duplicates(items):
  """
  Identifies and returns duplicate items in the provided list.

  Args:
      items (list): A list of items to check for duplicates.

  Returns:
      list: A list containing duplicate items found in the original list.
  """
  seen = set()
  duplicates = []
  for item in items:
    if json.dumps(item) in seen:
      duplicates.append(json.dumps(item))
    else:
      seen.add(json.dumps(item))
  return duplicates

# Replace with your actual values
url = "http://213.239.193.208:9053/blockchain/box/byAddress"
headers = {"accept": "application/json", "Content-Type": "application/json"}
data = "9enWZ4fCvhw7pukG94oMMVvNXW3ZMWHNJABe9hRciKPfGTSA2z1"

all_items = fetch_all_items(url, headers, data)
duplicates = find_duplicates(all_items)

print("Duplicate items:", duplicates)
print("\n\n\nTotal items fetched:", len(all_items))
print("Duplicate number:", len(duplicates))
```

kushti:
close #2166


kushti:
@jellymlg can you fix /unspent accordingly, with a test also ?

kushti:
@jellymlg added test for limit = 200 , it is failing

jellymlg:
> @jellymlg added test for limit = 200 , it is failing

There is only 284 (1706 / 6) UTXOs in the test, so the expected values are not correct

kushti:
> > @jellymlg added test for limit = 200 , it is failing
> 
> @kushti There is only 284 (1706 / 6) UTXOs in the test, so the expected values are not correct

Ok, will fix the test! 

Can you also do a test and fix transactions output? Seems ordering is not good thee, while they are stored in ascending order , they sorted in descending during API results forming, I guess, they should be fetched in descending as well then

kushti:
@jellymlg fixed the test, assigned a bounty for completion, any ETA on tx ordering?

jellymlg:
@kushti Is the problem maybe with `Segment.getTxs` re-sorting txs?

kushti:
> @kushti Is the problem maybe with `Segment.getTxs` re-sorting txs?

yes, this also, there was some motivation behind that I guess ?

jellymlg:
> > @kushti Is the problem maybe with `Segment.getTxs` re-sorting txs?
> 
> yes, this also, there was some motivation behind that I guess ?

It was resorted to match explorer sorting scheme

kushti:
> > > @kushti Is the problem maybe with `Segment.getTxs` re-sorting txs?
> > 
> > 
> > yes, this also, there was some motivation behind that I guess ?
> 
> It was resorted to match explorer sorting scheme

Do we need it ? If so, then the only way is to fix ordering inside

jellymlg:
@kushti i do not think its that important, so not needed

kushti:
@jellymlg fixed I believe

PR #1028, opened 2020-01-14
Fixing styling issues and minor issues found by an automatic inspection 

kushti:
This PR contains trivial improvements proposed by Intellij Idea's automatic code inspection tool

PR #2260, opened 2025-12-13
Fix (#1028): Add Automatic-Module-Name to JAR manifests for JPMS compatibility 

dimension-drifter:
## Description

This PR addresses issue #1028 by adding the `Automatic-Module-Name` manifest attribute to all published JAR files in the Ergo project. This ensures proper Java Platform Module System (JPMS) compatibility when these JARs are used as automatic modules in Java 9+ applications.

## Changes

Added `Automatic-Module-Name` manifest attribute to the following modules:

| Module | Automatic-Module-Name |
|--------|----------------------|
| ergo (main node) | `org.ergoplatform.node` |
| ergo-core | `org.ergoplatform.core` |
| ergo-wallet | `org.ergoplatform.wallet` |
| avldb | `org.ergoplatform.avldb` |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,7 @@ local.conf
 project/metals.sbt
 null/
 out/
-
+ergoplatform-ergo-8a5edab282632443.txt
+.github
 # scala worksheets
 *.worksheet.sc
--- a/build.sbt
+++ b/build.sbt
@@ -212,6 +212,10 @@ lazy val avldb = (project in file("avldb"))
     crossScalaVersions := Seq(scala213, scalaVersion.value, scala211),
     commonSettings,
     name := "avldb",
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -64,6 +64,7 @@ local.conf
 project/metals.sbt
 null/
 out/
-
+ergoplatform-ergo-8a5edab282632443.txt
```
kushti, line 67: better add to .gitignore some folder with LLM reports

build.sbt
```diff
@@ -331,30 +346,3 @@ lazy val ergo = (project in file("."))
   .dependsOn(avldb % "test->test;compile->compile")
   .configs(It2Test)
 
-
```
kushti, line 334: why removed?

kushti (changes requested):

PR #2261, opened 2025-12-13
Fixed Storage rent with current parameters causing overflow error for…

### LINKED ISSUE #2251: Storage rent with current parameters cause overflow error for big boxes
Found by @Alesfatalis  : 

```
There an integer overflow on this line: https://github.com/ergoplatform/ergo/blob/21ca898f21b01345913fbdea68c183bcd8be0134/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoInterpreter.scala#L43


For example the affected box:
e7d20f8d1c063ef1d614ac75f26c16f468f438722f915f740b3caabb1a528a09


Has the size: 2239
And the storage fee factor is: 1250000

So 2239 * 1250000 = 2798750000 > 2147483647
```

I think we can do following about that : 

* ask miners to lower `storageFeeFactor`  value to have every box (max 4,096 bytes) collectable 
* prepare a fix , need to check how to do it in a soft-fork way.

Garry400:
## Integer Overflow Fix
This fixes #2251 integer overflow issue in `ErgoInterpreter.scala`

## Changes
1. `ErgoInterpreter.scala`
I modified the storage fee calculation to use Long arithmetic, preventing overflow when box.bytes.length and storageFeeFactor are large.

```protected def checkExpiredBox(box: ErgoBox, output: ErgoBoxCandidate, currentHeight: Height): Boolean = {
-    val storageFee = params.storageFeeFactor * box.bytes.length
+    val storageFee = params.storageFeeFactor.toLong * box.bytes.length
 
     val storageFeeNotCovered = box.value - storageFee <= 0
```
### Logic Check
- Before: Int * Int resulted in an Int. If the product exceeded 2,147,483,647, it wrapped around to a negative number.
- After: Int.toLong * Int results in a Long. The maximum possible value is Long.MaxValue, which is far larger than any possible
storage fee (approx 9e18), easily accommodating the example case 2798750000.
## Reproduction Test
I added a new test file 
`ergo-wallet/src/test/scala/org/ergoplatform/wallet/interpreter/OverflowSpec.scala`
 that reproduces the overflow with Int and verifies the fix with Long.

```it should "not overflow with Long cast" in {
    val storageFeeFactor: Int = 1250000
    val boxBytesLength: Int = 2239
    val result: Long = storageFeeFactor.toLong * boxBytesLength
    result should be (2798750000L)
  }
```
This test confirms that the fix correctly handles the values provided in the issue report.

### DIFF
--- a/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoInterpreter.scala
+++ b/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoInterpreter.scala
@@ -40,7 +40,7 @@ class ErgoInterpreter(params: BlockchainParameters)
     * @return whether the box is spent properly according to the storage fee rule
     */
   protected def checkExpiredBox(box: ErgoBox, output: ErgoBoxCandidate, currentHeight: Height): Boolean = {
-    val storageFee = params.storageFeeFactor * box.bytes.length
+    val storageFee = params.storageFeeFactor.toLong * box.bytes.length
 
     val storageFeeNotCovered = box.value - storageFee <= 0
     lazy val correctCreationHeight = output.creationHeight == currentHeight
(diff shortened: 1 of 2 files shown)


kushti:
proposed change is hardfork-ish , would cause network split between old and new  versions. A proper implementation would handle versioning , activation etc

PR #2272, opened 2025-12-13
Fix integer overflow in storage fee calculation (issue #2251)

### LINKED ISSUE #2251: Storage rent with current parameters cause overflow error for big boxes

EHyashu:
This commit fixes an integer overflow issue when calculating storage fees for large boxes. The problem occurred when multiplying storageFeeFactor (Int) by box.bytes.length (Int), which could exceed Int.MaxValue for boxes larger than ~1700 bytes with the current storageFeeFactor of 1250000.

Example: For a box of 2239 bytes with storageFeeFactor=1250000:
- Before: 2239 * 1250000 = 2798750000 > Int.MaxValue (2147483647) -> overflow
- After: 2239L * 1250000 = 2798750000L -> correct result

Changes:
- Fixed ErgoInterpreter.checkExpiredBox to use .toLong conversion
- Fixed all test cases in ExpirationSpecification to use .toLong conversion
- Added comprehensive test case to verify overflow fix

This fix ensures that all boxes up to the maximum size (4096 bytes) can be properly handled without integer overflow, making the storage fee calculation safe and correct for all valid box sizes.

Fixes #2251

### DIFF
--- a/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoInterpreter.scala
+++ b/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoInterpreter.scala
@@ -40,7 +40,7 @@ class ErgoInterpreter(params: BlockchainParameters)
     * @return whether the box is spent properly according to the storage fee rule
     */
   protected def checkExpiredBox(box: ErgoBox, output: ErgoBoxCandidate, currentHeight: Height): Boolean = {
-    val storageFee = params.storageFeeFactor * box.bytes.length
+    val storageFee = params.storageFeeFactor.toLong * box.bytes.length
 
     val storageFeeNotCovered = box.value - storageFee <= 0
     lazy val correctCreationHeight = output.creationHeight == currentHeight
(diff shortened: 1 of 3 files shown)


kushti:
Proposed changes are hardfork-ish, causing network split between old and new versions . A proper implementation would need for versioning etc

PR #2290, opened 2026-01-16
/utils/schnorrSign API metod

kushti:
# Schnorr Signature API Specification

## Overview
The new `/utils/schnorrSign` endpoint allows users to generate Schnorr signatures for arbitrary messages using the private key associated with a specified P2PK address in the node's wallet. It would be useful for offchain applications working with Schnorr signature verifying contracts, see https://www.ergoforum.org/t/verifying-schnorr-signatures-in-ergoscript/3407

## API Endpoint
```
POST /utils/schnorrSign
```

## Request Format
```json
{
  "signerAddress": "3WwbzW6u8hKWBcL1W7kNVMr25s2UHfSBnYtwSHvrRQt7DdPuoXrt",
  "message": "68656c6c6f20776f726c64"
}
```

### Parameters
- `signerAddress` (string, required): The P2PK address whose associated private key will be used for signing
- `message` (string, required): The hex-encoded message to be signed

## Response Format
On success (HTTP 200):
```json
{
  "signedMessage": "68656c6c6f20776f726c64",
  "signature": "01abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
  "publicKey": "02a7955281885bf0f0ca4a48678848cad8dc5b328ce8bc1d4481d041c98e891ff3"
}
```

### Response Fields
- `signedMessage`: The hex-encoded signed message
- `signature`: The hex-encoded Schnorr signature (concatenated R and s values)
- `publicKey`: The hex-encoded public key used for signing

## Error Responses
- HTTP 400: Invalid address, message, or node does not have the secret key for the specified address
- HTTP 500: Internal server error (e.g., wallet not initialized)

## Signing Process
1. The API validates that the provided `signerAddress` is a valid P2PK address
2. It looks up the corresponding extended public key in the wallet
3. It retrieves the private key for the derivation path associated with that public key
4. It generates a Schnorr signature using the following algorithm:
   - Generate a random nonce `k`
   - Calculate point `a = g^k` (where `g` is the generator)
   - Calculate challenge `e = H(a || message || publicKey)`
   - Calculate response `z = k + e * s (mod n)` where `s` is the private key
   - Return signature as `a` concatenated with `z`

## Usage Example
```bash
curl -X POST http://localhost:9053/utils/schnorrSign \
  -H "Content-Type: application/json" \
  -d '{
    "signerAddress": "3WwbzW6u8hKWBcL1W7kNVMr25s2UHfSBnYtwSHvrRQt7DdPuoXrt",
    "message": "68656c6c6f20776f726c64"
  }'
```

## Prerequisites
- The node must have the private key for the specified address in its wallet
- The address must be a valid P2PK address
- The message must be properly hex-encoded


odiseusme:
Took a pass (with Claude) through this; findings ordered by severity, happy to be corrected on any:

1. /utils/schnorrSign sits in the unauthenticated `utils` group but signs with wallet keys.
   schnorrSignR is mounted under pathPrefix("utils") next to seed, hash/blake2b and the
   address helpers, with no auth directive — and the OpenAPI block has no `security:` field
   (unlike /wallet/init right below it). The /utils group is served without api-key auth, so
   this route is callable unauthenticated, yet it returns a signature made with the wallet
   private key for any address the node controls. On a node with apiKeyHash set, /wallet/status
   is rejected without the key while /utils/seed still answers — so as written this is an
   unauthenticated signing oracle over the node's keys whenever the wallet is unlocked.
   Suggest moving it behind api-key protection (or into the wallet route group).

2. Hand-rolled Schnorr that diverges from ErgoSignature.
   ergo-wallet already provides ErgoSignature.sign/verify, whose challenge is
   H(genCommitment(pk, w) ++ msg) truncated to 24 bytes with the SigmaTree-equivalent framing,
   serialized as c(24) ++ z(32) — output that matches the ProveDlog proof form and has a
   matching verifier. This route instead computes e = Blake2b256(a ++ message ++ pubKey)
   (full 32 bytes, no framing) and serializes a(33) ++ z(32). The result is not verifiable by
   ErgoSignature.verify and is not a valid sigma/ProveDlog proof, so nothing in the repo can
   check it. Recommend reusing ErgoSignature.sign, or documenting which verifier this targets.

3. Debug logging flagged for removal.
   log.info(s"e: $eBI , challenge input: …") carries `// todo: remove before release`. It logs
   the challenge and a‖msg‖pk per call (no secret key or nonce), but should come out before this ships.

4. No happy-path test.
   UtilsApiRouteSpec covers four rejection paths (uninitialised wallet, non-P2PK address, bad
   hex, bad address) but never signs with a real key and verifies the output — which is why (2)
   isn't caught. A sign→verify round-trip test would pin the intended format.

5. Minor:
   - catch { case _: Throwable => InternalError("WalletError") } — prefer NonFatal, and
     surfacing the underlying message would aid debugging.
   - Three sequential blocking Await.result(...) calls in the route can starve the akka-http
     dispatcher under load; the future/async style used elsewhere would be safer.
   - Error responses mix sentences and tokens (InvalidAddress, InvalidMessage), and a JSON-parse
     failure returns 500 InternalError where 400 BadRequest seems more apt.

(Disclosure: reviewed with AI assistance, findings verified against the source.)

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
       case Left(ex)          => ApiError(StatusCodes.BadRequest, ex.getMessage())
     }
   }
+
+  def schnorrSignR: Route = (post & path("schnorrSign") & entity(as[Json])) { json =>
```
jozanek, line 143: **[BLOCKER]** Unauthenticated signing oracle over wallet private keys. `/utils/*` isn't wrapped in `withAuth`, but this endpoint reads `getPrivateKeyFromPath` and produces a signature under the wallet's P2PK key. Every other private-key operation (`getPrivateKeyR`, `signTransactionR`) sits behind `withAuth` in `WalletApiRoute` for exactly this reason — any client that can reach the REST API can now mint signatures for arbitrary bytes without the API key. Please gate this behind `withAuth`, or move it into `WalletApiRoute`.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
...
+                          val kBI = BigInt(BigIntegers.fromUnsignedByteArray(kBytes))
+
+                          // Calculate a = g^k (random point)
+                          val aPoint = CryptoConstants.dlogGroup.exponentiate(CryptoConstants.dlogGroup.generator, kBI.bigInteger)
+
+                          // Calculate challenge e = H(a || message || public_key)
+                          val aBytes = GroupElementSerializer.toBytes(aPoint)
+                          val challengeInput = aBytes ++ messageBytes ++ publicKeyBytes
+                          val eFull = Blake2b256(challengeInput)
+                          val eBI = BigInt(BigIntegers.fromUnsignedByteArray(eFull)).mod(CryptoConstants.groupOrder)
```
jozanek, line 200: **[MAJOR]** No domain separation on the challenge. `e = H(a ‖ message ‖ pubkey)` signs raw caller-supplied bytes with the same key that controls funds, with no protocol tag. A `message` could be chosen to collide with another signing context (cross-protocol key reuse), which is especially dangerous given the endpoint is currently unauthenticated. Add a fixed domain-separation prefix to the hash input and document it in the spec.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
...
+                  // Find the private key for the given address by looking up the public key
+                  val extKeysFuture = walletReader.allExtendedPublicKeys()
+                  val extKeys = Await.result(extKeysFuture, 5.seconds)
+
+                  extKeys.find(_.key.value.equals(p2pkAddress.pubkey.value)) match {
+                    case Some(extKey) =>
+                      val path = extKey.path
+                      // Get the private key for the derivation path
+                      val privateKeyFuture = walletReader.getPrivateKeyFromPath(path)
+                      val privateKeyTry = Await.result(privateKeyFuture, 5.seconds)
```
jozanek, line 173: **[MAJOR]** `Await.result` blocks the Akka HTTP dispatcher for up to 15s per request (3 × 5s, sequential). `ApiResponse` already accepts `Future[Json]` (see `ApiResponse.scala:19-20`), so chain the three reads with `flatMap`/a for-comprehension and return the `Future` instead of awaiting. That also lets failures land in `onComplete` and removes the need for the surrounding `try/catch`.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
...
+                          val aPoint = CryptoConstants.dlogGroup.exponentiate(CryptoConstants.dlogGroup.generator, kBI.bigInteger)
+
+                          // Calculate challenge e = H(a || message || public_key)
+                          val aBytes = GroupElementSerializer.toBytes(aPoint)
+                          val challengeInput = aBytes ++ messageBytes ++ publicKeyBytes
+                          val eFull = Blake2b256(challengeInput)
+                          val eBI = BigInt(BigIntegers.fromUnsignedByteArray(eFull)).mod(CryptoConstants.groupOrder)
+
+                          // todo: remove before release
+                          log.info(s"e: $eBI , challenge input: ${Base16.encode(challengeInput)}")
```
jozanek, line 203: **[MAJOR]** Leftover debug log, flagged by your own `// todo: remove before release`. It logs the challenge and full challenge-input at `INFO` on every call — please remove before merge.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
...
+
+                        case scala.util.Failure(exception) =>
+                          BadRequest(s"Node does not have the secret key for the specified address - ${exception.getMessage}")
+                      }
+                    case None =>
+                      BadRequest("Node does not have the secret key for the specified address")
+                  }
+                } catch {
+                  case _: Throwable =>
+                    InternalError("WalletError")
```
jozanek, line 231: **[MAJOR]** `catch { case _: Throwable => InternalError("WalletError") }` swallows the real cause and catches fatal errors / `InterruptedException`. Use `scala.util.control.NonFatal` and log the cause — or, if the awaits become a future chain (see the concurrency comment), drop the `try/catch` entirely and handle failure in `onComplete`.

src/test/scala/org/ergoplatform/http/routes/UtilsApiRouteSpec.scala
```diff
@@ -153,5 +154,58 @@ class UtilsApiRouteSpec extends AnyFlatSpec
...
+      "signerAddress" -> "invalid_address".asJson,
+      "message" -> "02415748f8eef16c5ea6896cec3a8defccc8a0dace245248be66ffd6ff2159da32000000000003d09000000000694fa26d".asJson
+    )
+
+    Post(s"$prefix/schnorrSign", requestJson) ~> route ~> check {
+      status shouldBe StatusCodes.BadRequest
+      val response = responseAs[Json]
+      response.hcursor.downField("detail").as[String] shouldEqual Right("InvalidAddress")
+    }
+  }
```
jozanek, line 208: **[MAJOR]** All four new tests assert error paths only — nothing signs a message and verifies the signature. For a crypto endpoint, signature correctness is the whole point and is currently unverified. Please add a sign-then-verify test against an initialized wallet (reconstruct `a' = g^z · pk^(−e)`, check `H(a' ‖ m ‖ pk) ≡ e`) so a future change can't silently break the scheme.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
...
+                BadRequest("InvalidAddressType")
+
+              case scala.util.Failure(_) =>
+                BadRequest("InvalidAddress")
+            }
+          case scala.util.Failure(_) =>
+            BadRequest("InvalidMessage")
+        }
+      case Left(ex) =>
+        InternalError(ex.getMessage())
```
jozanek, line 244: **[MINOR]** A malformed request body returns `InternalError` (HTTP 500). A bad JSON body should be `BadRequest` (400), matching the sibling routes (`:64`, `:109`, `:132`). The OpenAPI for this endpoint only documents 200/400/default anyway.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
       case Left(ex)          => ApiError(StatusCodes.BadRequest, ex.getMessage())
     }
   }
+
+  def schnorrSignR: Route = (post & path("schnorrSign") & entity(as[Json])) { json =>
+
+    import io.circe.generic.auto._
```
jozanek, line 145: **[MINOR]** Method-local imports including a wildcard `import io.circe.generic.auto._`. The project convention is no wildcard imports, and auto-derived request codecs should live in `ApiCodecs`/`ApiRequestsCodecs` rather than inline. `import java.security.SecureRandom` (line 184) also duplicates the file-level import at line 17. Hoist these to the top and define a `SchnorrSignRequest` decoder alongside the existing ones.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,14 +139,121 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
...
+
+                          // todo: remove before release
+                          log.info(s"e: $eBI , challenge input: ${Base16.encode(challengeInput)}")
+
+                          // Calculate response z = k + e * s (mod n) where s is the private key
+                          val privateKeyBI = privateKeyInput.w
+                          val zBI = kBI.bigInteger.add(eBI.bigInteger.multiply(privateKeyBI)).mod(CryptoConstants.groupOrder)
+
+                          // Get the compressed form of R for the signature format
+                          val aComponent = GroupElementSerializer.toBytes(aPoint)
```
jozanek, line 210: **[MINOR]** `aComponent` here and `aBytes` at line 197 recompute `GroupElementSerializer.toBytes(aPoint)` identically — collapse to a single val.

src/test/scala/org/ergoplatform/http/routes/UtilsApiRouteSpec.scala
```diff
@@ -153,5 +154,58 @@ class UtilsApiRouteSpec extends AnyFlatSpec
...
+  it should "return error for schnorrSign when wallet is not initialized" in {
+    val requestJson = Json.obj(
+      "signerAddress" -> p2pkaddress.toString.asJson,
+      "message" -> "02415748f8eef16c5ea6896cec3a8defccc8a0dace245248be66ffd6ff2159da32000000000003d09000000000694fa26d".asJson
+    )
+
+    Post(s"$prefix/schnorrSign", requestJson) ~> route ~> check {
+      status shouldBe StatusCodes.InternalServerError
+      val response = responseAs[Json]
+      println(s"SchnorrSign response: $response")
```
jozanek, line 166: **[MINOR]** `println` debug output left in the specs (here, and in the other new test cases at lines 178) — please drop it.

src/main/resources/api/openapi.yaml
```diff
@@ -4183,6 +4183,63 @@ paths:
...
+                    type: string
+                    example: '68656c6c6f20776f726c64'
+                  signature:
+                    description: The hex-encoded Schnorr signature
+                    type: string
+                    example: '01abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890'
+                  publicKey:
+                    description: The hex-encoded public key used for signing
+                    type: string
+                    example: '02a7955281885bf0f0ca4a48678848cad8dc5b328ce8bc1d4481d041c98e891ff3'
```
jozanek, line 4229: **[NIT]** Response properties (`signedMessage`, `signature`, `publicKey`) aren't marked `required`. Also `signedMessage` just echoes the request input, so it's redundant in the response.

jozanek (commented):
Review of the `/utils/schnorrSign` endpoint. Inline comments below. Verdict: **Request changes** — 1 BLOCKER, 5 MAJOR, 4 MINOR, 1 NIT. The headline issue is that this exposes a wallet-key signing oracle on the unauthenticated `/utils` surface; please also address the blocking `Await`s, the leftover debug logging, the `Throwable` catch, and the missing sign-then-verify test.

Non-blocking note on the scheme: it's a self-consistent Schnorr variant, but it is **not** Ergo's internal Sigma protocol (which truncates the challenge to `soundnessBits`, not full-256-bit `mod n`). That's fine for a bespoke ErgoScript verifier, but the verifier must decode the Blake hash as an *unsigned* big-endian integer `mod n`; if the contract uses signed `byteArrayToBigInt`, verification will mismatch. Please pin the exact verifier algorithm in the spec and cover it with the round-trip test.

PR #2291, opened 2026-01-22
Candidate for 6.0.3 release

### LINKED ISSUE #1448: Check spending of tokens created in offchain transaction
Reportedly, spending of a token created in UTXO which is offchain results in:

invalidated. Cause: For every token, its amount in outputs should not exceed its amount in inputs. 157f887f6698d1ba643990b93ea7bb1474918f56a24fcc899a4b0630fb85d685: Amount in = -1, out = 1. Allowed new asset = ByteArrayWrapper[5165884FF23345F374019C86016AE409C9E6E444357DEA5CFF3ED3BF19AC7D97], out = ByteArrayWrapper[870BECF7B15D6345DC746078306CAFA5D05E5B769580DE35BFA9035E6A300179]

kushti:
PRs merged:
- [Put tree version into Ergo tree in /script/p2s #2247](https://github.com/ergoplatform/ergo/pull/2247)
- [New public testnet parameters #2252](https://github.com/ergoplatform/ergo/pull/2252)
- [Custom public key mining api route in testnet #2287](https://github.com/ergoplatform/ergo/pull/2287)
- [Fix for sbt it:test #2294](https://github.com/ergoplatform/ergo/pull/2294)
- [Do not try to mine when the blockchain is not synced #2295](https://github.com/ergoplatform/ergo/pull/2295)
- [Mempool removal improvements #2299](https://github.com/ergoplatform/ergo/pull/2299)
- [Fix for "bounds must be positive" and faster missing parent header download #2305](https://github.com/ergoplatform/ergo/pull/2305)
- [Tests: Add regression test for chained off-chain token minting (Closes #1448) #2302](https://github.com/ergoplatform/ergo/pull/2302)
- [P2P logic improvements #2306](https://github.com/ergoplatform/ergo/pull/2306)
- [cover exSize/exIlEncoding/exKeyLength/exEmpty rules #2310](https://github.com/ergoplatform/ergo/pull/2310)
- [docs(openapi): expand wallet init/restore/derive endpoint descriptions #2318](https://github.com/ergoplatform/ergo/pull/2318)
- [Fix mining candidate cache for custom public keys #2373](https://github.com/ergoplatform/ergo/pull/2373)

Contributors: [@jozanek](https://github.com/jozanek) , [@K-Singh](https://github.com/K-Singh) , [@kushti](https://github.com/kushti) , [@a-shannon](https://github.com/a-shannon)

---

# Ergo Node v6.0.3 Release Notes

## Overview

This release brings mining API enhancements, a new public testnet (testnet60), P2P networking improvements, mempool handling fixes, and expanded test coverage. It also includes infrastructure updates and better API documentation.

### Mining API Enhancements

- **Custom public key mining route** ([#2287](https://github.com/ergoplatform/ergo/pull/2287)) — New `/mining/candidateWithTxsAndPk` endpoint allows miners to generate candidate blocks containing only transactions filtered by a given public key. Useful for pools and specialized mining setups.
  - New `MiningRequest` data structure for request parameters.
  - Fixes parameter handling bugs in `GenerateCandidate` ([#2287](https://github.com/ergoplatform/ergo/pull/2287)).
  - **Mining candidate cache fix** ([#2373](https://github.com/ergoplatform/ergo/pull/2373)) — Candidate cache is now properly isolated by public key, preventing cache collisions when switching between different mining keys.

### Testnet60 Launch

- **New public testnet parameters** ([#2252](https://github.com/ergoplatform/ergo/pull/2252)) — Launches testnet60 with updated genesis parameters:
  - Block version set to `Interpreter60Version` at genesis.
  - Rules 215 and 409 disabled at genesis.
  - New magic bytes `[2,3,2,3]`, port `9023`.
  - Old testnet data directories are incompatible; a fresh sync is required.

### API Improvements

- **`/script/p2s` tree version** ([#2247](https://github.com/ergoplatform/ergo/pull/2247)) — The `/script/p2s` endpoint now includes the Ergo tree version in the response, improving contract debugging and tooling.
- **OpenAPI documentation** ([#2318](https://github.com/ergoplatform/ergo/pull/2318)) — Expanded descriptions for wallet init, restore, and key derivation endpoints.

### Mining & Node Behavior

- **Do not mine when unsynced** ([#2295](https://github.com/ergoplatform/ergo/pull/2295)) — The node now checks whether the blockchain is nearly synced (within 6 blocks) before starting internal mining. Prevents wasted work and stale block submission during initial sync.
- **Better error logging** — Transaction parsing failures now log the actual failure reason for easier debugging.

### P2P Networking

- **P2P logic improvements** ([#2306](https://github.com/ergoplatform/ergo/pull/2306)):
  - Message size checks derived from `maxModifierSize` for consistency.
  - `NetworkUtils.checkLocalOnly` renamed to `allowLocal` for clarity.
  - `FallbackToEqualPeerThreshold` constant replaces magic number for peer fallback logic.
  - Local peers are removed from the peer database to avoid stale connections.
  - New blocks are announced immediately before full validation, improving propagation speed.
  - `setRequestedDirectly` flag for better modifier request tracking.

### Mempool

- **Mempool removal improvements** ([#2299](https://github.com/ergoplatform/ergo/pull/2299)):
  - Fixed edge cases in transaction replacement with higher-fee variants.
  - Improved handling of stale entries in `OrderedTxPool`.

### Infrastructure

- **Docker JDK update** — Base image updated from `openjdk:11-jre-slim` to `eclipse-temurin:11-jre-jammy` for better security and long-term support.
- **Integration test fixes** ([#2294](https://github.com/ergoplatform/ergo/pull/2294)) — Fixes for `sbt it:test` stability.

## 🐛 Bug Fixes

- **"bounds must be positive" fix** ([#2305](https://github.com/ergoplatform/ergo/pull/2305)) — Fixes a rare crash during modifier processing. Also improves automatic redownload of missing parent headers.
- **Duplicate IDs fix** — Resolved duplicate transaction ID issues in candidate generation.
- **Wallet cache fix** ([#2244](https://github.com/ergoplatform/ergo/pull/2244)) — Fix for unversioned tree template in wallet cache.

## 🧪 Testing

- **Chained off-chain token minting regression test** ([#2302](https://github.com/ergoplatform/ergo/pull/2302)) — Adds a regression test for issue #1448 covering chained off-chain token minting scenarios.
- **Synthetic network type tests** — Added `NetworkTypeSpec`, `MiningRequestSpec`, `ScriptApiRouteSpec`, `LaunchParametersSpec`, and comprehensive `CandidateGeneratorSpec` tests.
- **Malformed extension rules coverage** ([#2310](https://github.com/ergoplatform/ergo/pull/2310)) — Tests covering `exSize`, `exIlEncoding`, `exKeyLength`, and `exEmpty` validation rules.
- **Integration test stabilization** — Multiple fixes for flaky IT tests including `RollbackSpec`, `NetworkControllerSpec`, `OffchainRegistrySpec`, and Autolykos PoW property tests.

## ⚠️ Breaking Changes

- **Testnet operators**: The old public testnet is discontinued. Testnet60 requires a fresh data directory and uses port `9023` by default. Update your `testnet.conf` accordingly.

## 📝 Migration Notes

- **Docker users**: The base image has changed to `eclipse-temurin:11-jre-jammy`.
- **Testnet operators**: Use the new `testnet.conf` with testnet60 parameters. Old testnet databases are incompatible.
- **Miners**: New `/mining/candidateWithTxsAndPk` endpoint is available for public-key-filtered candidate generation.

## 🙏 Contributors

- **Alexander Chepurnoy** — Core development, testnet60 launch, P2P improvements, mempool fixes, testing.
- **K-Singh** — Mining API enhancements and candidate cache fixes.
- **jozanek** — OpenAPI docs, malformed extension tests, review and feedback.
- **a-shannon** — Regression test for chained off-chain token minting.

---

**Full commit diff**: [`v6.0.2...v6.0.3`](https://github.com/ergoplatform/ergo/compare/v6.0.2...v6.0.3)


src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -230,11 +232,17 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
...
+          updateInfo.state.applyModifier(modToApply, chainTipOpt)(lm => pmodModify(lm.pmod, local)) match {
             case Success(stateAfterApply) =>
               history.reportModifierIsValid(modToApply).map { newHis =>
                 if (modToApply.modifierTypeId == ErgoFullBlock.modifierTypeId) {
-                  context.system.eventStream.publish(FullBlockApplied(modToApply.asInstanceOf[ErgoFullBlock].header))
+                  val header = modToApply.asInstanceOf[ErgoFullBlock].header
+                  val event = if (local) {
+                    LocalBlockApplied(header)
+                  } else {
+                    RemoteBlockApplied(header)
```
jozanek: [MAJOR] Blocks submitted via the API reach the holder as `LocallyGeneratedModifier` (`BlocksApiRoute.scala:130`), so they apply with `local = true` → `LocalBlockApplied` → the synchronizer skips the inv broadcast (`ErgoNodeViewSynchronizer.scala:1429`). But `NewBlockMined` is only published by `CandidateGenerator`, so API-submitted blocks are never announced to peers and propagate only via slow sync.

Suggestion: publish `NewBlockMined` in `BlocksApiRoute` too, or derive Local/Remote from "was already announced" rather than from the `local` flag.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -66,6 +66,15 @@ class CandidateGenerator(
...
       s"New block ${newBlock.id} w. nonce ${Longs.fromByteArray(newBlock.header.powSolution.n)}"
     )
+
+    // Immediately announce newly mined block to network BEFORE local application.
+    // This reduces propagation latency by avoiding the wait for NodeViewHolder
+    // to fully validate and apply the block. LocalBlockApplied arrives later
+    // and skips broadcast since the block was already announced.
+    // TODO: Consider switching to direct actor message for lower latency
+    //       instead of event bus publish.
+    context.system.eventStream.publish(NewBlockMined(newBlock.header))
```
jozanek, line 76: [MAJOR] `NewBlockMined` invs are broadcast before the block is applied, but `GetModifiers` is served from history — peers' requests race local application. Slow or failed application earns this node non-delivery penalties, and peers that exhaust `maxDeliveryChecks` mark the announced header Invalid permanently. For UTXO miners the announced ADProofs don't even exist until application completes.

Suggestion: announce after header application, or make the modifier-serving path able to serve the solved block from `CandidateGenerator`'s `solvedBlock` cache.

src/main/scala/org/ergoplatform/network/message/MessageSerializer.scala
```diff
@@ -48,6 +50,9 @@ class MessageSerializer(specs: Seq[MessageSpec[_]], magicBytes: Array[Byte]) {
       if (length < 0) {
         throw MaliciousBehaviorException("Data length is negative!")
       }
+      if (length > MaxMessageSize) {
+        throw MaliciousBehaviorException("Data length is above limit!")
+      }
```
jozanek, line 56: [MINOR] The new 16.4 MB frame cap rejects input the old deserializer accepted. It looks wire-compatible — every legitimate message is bounded far below it (ModifiersSpec reserve 8.4 MB, manifest/chunk 4 MB, peers/sync messages are tiny) — but neither the PR body nor a comment states this argument.

Suggestion: document the bound rationale next to `MaxMessageSize` (`MessageConstants.scala:19`) and in the release notes.

ergo-core/src/main/scala/org/ergoplatform/settings/LaunchParameters.scala
```diff
@@ -13,8 +13,8 @@ object MainnetLaunchParameters extends Parameters(height = 0,
   * Parameters corresponding to the genesis block in the public testnet
   */
 object TestnetLaunchParameters extends Parameters(height = 0,
-  parametersTable = Parameters.DefaultParameters,
-  proposedUpdate = ErgoValidationSettingsUpdate.empty)
+  parametersTable = Parameters.DefaultParameters.updated(Parameters.BlockVersion, Header.Interpreter60Version),
+  proposedUpdate = ErgoValidationSettingsUpdate(Seq(215, 409), Seq.empty))
```
jozanek, line 17: [MINOR] Testnet genesis parameters changed (BlockVersion → `Interpreter60Version`, rules 215/409 disabled at genesis) together with `testnet.conf` magic bytes `[2,0,2,3]` → `[2,3,2,3]`, port 9022 → 9023, checkpoint and re-emission ids removed — a deliberate testnet60 relaunch, properly isolated from the old testnet by the magic-byte change; mainnet unaffected.

Suggestion: call the testnet reset out explicitly in the release notes so operators know old testnet data dirs are incompatible.

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -117,25 +117,41 @@ class ErgoMiner(
...
+                ergoSettings,
+                minerState.candidateGeneratorRef,
+                minerState.secretKeyOpt.get.w
+              )(context)
+            }
+          }
+          context.system.eventStream
+            .unsubscribe(self, classOf[FullBlockApplied])
+          context.become(started(minerState))
+        } else {
```
jozanek: [MAJOR] The sync-check closure passed to `GetDataFromCurrentView` executes on ErgoNodeViewHolder's thread (`sender() ! f(...)`, `ErgoNodeViewHolder.scala:679`), yet calls `context.become(started(...))` and spawns `ErgoMiningThread`s via the miner's `context` — illegal cross-thread `ActorContext` use. It is also racy: several `StartMining`/`FullBlockApplied` messages can fire more closures before `become` lands, spawning duplicate miner thread sets. (The pre-existing closure at line 74 only `tell`s, which is safe.)

Suggestion: have the closure return `(headersHeight, fullBlockHeight)` as a reply message and do the `become`/spawn/unsubscribe inside the miner's own `receive`.

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -117,25 +117,41 @@ class ErgoMiner(
...
-          s"Starting ${ergoSettings.nodeSettings.internalMinersCount} native miner(s)"
-        )
-        (1 to ergoSettings.nodeSettings.internalMinersCount) foreach { _ =>
-          ErgoMiningThread(
-            ergoSettings,
-            minerState.candidateGeneratorRef,
-            minerState.secretKeyOpt.get.w
-          )(context)
+      // Check if blockchain is synced before starting mining
+      viewHolderRef ! GetDataFromCurrentView[DigestState, Unit] { v =>
```
jozanek: [NIT] `GetDataFromCurrentView[DigestState, Unit]` hard-codes `DigestState` for a query that runs on UTXO nodes (works only via erasure), and the `Unit` reply lands in the miner's mailbox as an unhandled message.

Suggestion: use a properly typed request/response message.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -152,7 +161,7 @@ class CandidateGenerator(
         context.become(initialized(state))
       }
 
-    case gen @ GenerateCandidate(txsToInclude, reply, forced) =>
+    case gen @ GenerateCandidate(txsToInclude, reply, forced, optPk) =>
       val senderOpt = if (reply) Some(sender()) else None
       if (!forced && cachedFor(state.cachedCandidate, txsToInclude)) {
```
jozanek: [MAJOR] The candidate cache ignores `optPk`: `cachedFor` matches on transactions only (and any cache satisfies an empty-`txs` request), and a freshly generated custom-pk candidate is stored as `cachedCandidate`. Both directions break:

1. `/mining/candidateWithTxsAndPk` can return a candidate paying the default key;
2. after a custom-pk generation, the internal miner / `/mining/candidate` mines a block paying the API caller's key until the cache rolls.

Suggestion: include the effective miner pk in the cache key (compare the requested pk in `cachedFor`), or never cache custom-pk candidates.

src/main/scala/org/ergoplatform/http/api/MiningApiRoute.scala
```diff
@@ -53,6 +59,20 @@ case class MiningApiRoute(miner: ActorRef,
     ApiResponse(candidateF)
   }
 
+  def candidateWithTxsAndPkR: Route = (path("candidateWithTxsAndPk")
+    & post & entity(as[MiningRequest]) & withAuth) { txsAndPk =>
+    val tryPk = Try(GroupElementSerializer.fromBytes(Hex.decode(txsAndPk.pk)))
+    val result = tryPk match {
+      case Failure(_) =>
+        Future.failed(new Exception("Could not decode hexadecimal string for given public key"))
```
jozanek: [MINOR] Malformed pk hex yields `Future.failed(new Exception(...))` → HTTP 500, while sibling routes return `BadRequest` for bad input.

Suggestion: reply with `BadRequest` on the `Failure` branch.

src/main/scala/org/ergoplatform/http/api/MiningApiRoute.scala
```diff
@@ -53,6 +59,20 @@ case class MiningApiRoute(miner: ActorRef,
     ApiResponse(candidateF)
   }
 
+  def candidateWithTxsAndPkR: Route = (path("candidateWithTxsAndPk")
```
jozanek, line 63: [MINOR] The new `/mining/candidateWithTxsAndPk` endpoint and its `MiningRequest` schema are absent from `openapi.yaml` (only `candidateWithTxs` is documented); the spec is authoritative for client generation.

Suggestion: add the path + request schema to `src/main/resources/api/openapi.yaml`.

src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
```diff
@@ -39,6 +39,8 @@ case class ErgoSettings(directory: String,
       Devnet60LaunchParameters
     } else if (networkType == NetworkType.TestNet) {
       TestnetLaunchParameters
+    } else if (networkType == NetworkType.Tests) {
+      MainnetLaunchParameters
```
jozanek, line 44: [NIT] The `NetworkType.Tests` branch returns `MainnetLaunchParameters`, identical to the `else` fallback.

Suggestion: drop the redundant branch or add a comment stating it's intentional.

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -105,15 +105,26 @@ class OrderedTxPool(val orderedTransactions: TreeMap[WeightedTxId, UnconfirmedTr
...
           outputs -- tx.outputs.map(_.id),
           inputs -- tx.inputs.map(_.boxId)
         ).updateFamily(tx, -wtx.weight, System.currentTimeMillis(), depth = 0)
-      case None => this
+      case _ =>
+        if (orderedTransactions.valuesIterator.exists(_.id == tx.id)) {
+          new OrderedTxPool(
+            orderedTransactions.filter(_._2.id != tx.id),
+            transactionsRegistry - tx.id,
+            invalidatedTxIds,
```
jozanek, line 121: [MAJOR] `remove(tx)`'s fallback (`case _`) now runs `orderedTransactions.valuesIterator.exists(...)` — an O(n) scan (plus a full TreeMap rebuild on hit) — for every removed tx whose id misses the registry. `remove` is called per block transaction on block application, and txs not in the pool are the common case, turning the old O(log n) `case None => this` into O(n·m) per block on the NodeViewHolder hot path.

Suggestion: keep three cases — registry hit + TreeMap hit → fast path; registry hit + TreeMap miss (the actual stale-entry bug) → fallback scan; registry miss → `this`.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -1262,15 +1275,31 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+               // randomly choose a peer for another block sections download attempt
               val newPeerCandidates: Seq[ConnectedPeer] = if (modifierTypeId == Header.modifierTypeId) {
                 getPeersForDownloadingHeaders(peer).toSeq
               } else {
                 getPeersForDownloadingBlocks.map(_.toSeq).getOrElse(Seq(peer))
               }
-              val newPeerIndex = scala.util.Random.nextInt(newPeerCandidates.size)
-              val newPeer = newPeerCandidates(newPeerIndex)
+              val newPeer = if (newPeerCandidates.isEmpty) {
+                if (checksDone > 5) {
```
jozanek: [NIT] Magic `checksDone > 5` threshold for the Equal/Older-peer fallback sits alongside the configurable `maxDeliveryChecks`.

Suggestion: derive it from config or name the constant.

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
```diff
@@ -369,64 +369,4 @@ class ErgoMinerSpec extends AnyFlatSpec with ErgoTestHelpers with Eventually {
     system.terminate()
   }
 
-  it should "mine after HF" in new TestKit(ActorSystem()) {
```
jozanek, line 372: [MAJOR] "mine after HF" (60 lines) was deleted — the only test of mining across the v1→v2 activation boundary. The stated justification (activation heights are mainnet-only) doesn't hold: the test built its own `forkSettings` with `version2ActivationHeight = 3` and didn't depend on network config; it was deleted rather than adapted to the new `networkType = Tests` constants.

Suggestion: restore it with settings compatible with the synthetic network type, or document why the scenario is untestable now.

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -588,46 +588,493 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
...
+    candidate.candidateBlock should not be null
+    
+    system.terminate()
+  }
+
+  it should "use default minerPk when optPk is None" in new TestKit(ActorSystem()) {
+    val testProbe = new TestProbe(system)
+    system.eventStream.subscribe(testProbe.ref, newBlockSignal)
+
+    val viewHolderRef: ActorRef = ErgoNodeViewRef(defaultSettings)
```
jozanek, line 641: [MINOR] The new `optPk` tests assert only `candidate should not be null`; "generate different candidates for different optPk values" never compares the candidates or checks which pk the candidate pays — which is exactly why the `cachedFor` pk issue (see comment in `CandidateGenerator.scala`) passes CI.

Suggestion: assert the candidate's miner pk equals the requested pk, including the cached-then-custom-pk sequence.

src/test/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolSpec.scala
```diff
@@ -526,4 +595,299 @@ class ErgoMemPoolSpec extends AnyFlatSpec
...
+    poolAfter.size shouldBe 1
+    poolAfter.contains(tx3.transaction.id) shouldBe true
+    poolAfter.contains(tx1.transaction.id) shouldBe false
+    poolAfter.contains(tx2.transaction.id) shouldBe false
+
+    // (Loser case skipped: after TX3 replaces TX1+TX2 via acceptIfNoDoubleSpend,
+    //  pool.inputs loses the spent box entries due to put-before-remove ordering
+    //  in acceptIfNoDoubleSpend, so a subsequent lower-fee double-spender is not
+    //  detected. This is a known production bug.)
+  }
```
jozanek, line 763: [MINOR] This trailing comment skips the loser case as "a known production bug", but this PR fixes that bug (remove-before-put in `ErgoMemPool.scala:206`) and the adjacent "detect double spend after replace-by-fee replacement" test proves the loser case now works.

Suggestion: un-skip the loser assertion and delete the stale comment.

src/test/scala/scorex/core/network/NetworkControllerSpec.scala
```diff
@@ -0,0 +1,275 @@
...
+        f.establishIncomingConnection(controller, peerManagerProbe, addr)
+      }
+
+      val testAddress = new InetSocketAddress("192.168.1.100", 9999)
+      val connectionProbe = TestProbe("TestConnection")
+      val localAddress = settings.scorexSettings.network.bindAddress
+
+      connectionProbe.send(controller, Tcp.Connected(testAddress, localAddress))
+      connectionProbe.expectMsg(Tcp.Close)
+      peerManagerProbe.expectNoMessage(500.millis)
```
jozanek, line 190: [MINOR] Tautological tests: the `incomingLimit` property computes the same formula twice and compares it to itself, and "outgoing connection scheduler should respect maxConnections" (line 239) is just `succeed`.

Suggestion: test against `NetworkController`'s actual behavior or remove the placeholders.

build.sbt
```diff
@@ -174,18 +174,34 @@ inConfig(Linux)(
...
   test := (test dependsOn docker).value,
   scalacOptions ++= Seq("-Xasync")
 ))
 
+// Cap how many forked test JVMs run concurrently. `+=` can only *reduce* concurrency, never
+// raise it, so this is safe against RAM blowup: worst case the integration suites stay
+// sequential. Raise the limit once parallel runs prove stable on the host (see CLAUDE.md notes
+// on fork/tip flakiness). Unit tests run in a single forked group by default, so they are
+// unaffected by this limit.
+Global / concurrentRestrictions += Tags.limit(Tags.ForkedTestGroup, 2)
```
jozanek: [MINOR] sbt's default `Global / concurrentRestrictions` already contains `Tags.limit(Tags.ForkedTestGroup, 1)`; `+=` adds a second, looser rule and the strictest wins, so the advertised 2-way it-suite parallelism is likely a no-op (the comment half-admits this).

Suggestion: replace the default rule (`:=` with the rest of the defaults, or filter out the old `ForkedTestGroup` limit) and verify with `show Global/concurrentRestrictions`.

src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
```diff
@@ -450,4 +523,621 @@ class ErgoNodeViewSynchronizerSpecification extends AnyPropSpec
...
+      val (us, bh) = createUtxoState(settings)
+      val bestBlockOpt = hhistory.bestFullBlockOpt
+      val newBlock = validFullBlock(bestBlockOpt, us, bh)
+
+      // First: NewBlockMined triggers immediate broadcast
+      synchronizerMockRef ! NewBlockMined(newBlock.header)
+
+      // Consume all inv messages from NewBlockMined
+      val deadline1 = System.currentTimeMillis() + 2000
+      while (System.currentTimeMillis() < deadline1) {
```
jozanek: [NIT] New tests drain probes with `System.currentTimeMillis()` deadline loops — wall-clock dependence in tests is flaky in CI.

Suggestion: use `probe.receiveWhile(2.seconds)`.

ergo-core/src/main/scala/org/ergoplatform/settings/Settings.scala
```diff
@@ -23,7 +23,7 @@ case class RESTApiSettings(
 case class NetworkSettings(
   nodeName: String,
   addedMaxDelay: Option[FiniteDuration],
-  localOnly: Boolean,
+  allowLocal: Boolean,
```
jozanek, line 26: [MAJOR] `scorex.network.localOnly` renamed to `allowLocal` with no back-compat: operators with `localOnly` in their conf silently fall back to `allowLocal = false`, which now also filters loopback addresses and actively purges local peers from the peer DB (`PeerManager.scala:53,67`, `NetworkController.scala:339`) — private-LAN deployments break silently on upgrade.

Suggestion: add a release-note migration entry and ideally read the old key as a fallback for one release.

src/main/resources/application.conf
```diff
@@ -488,7 +489,7 @@ scorex {
     upnpEnabled = no
 
     # Accept only local connections
-    localOnly = false
+    allowLocal = false
```
jozanek: [MINOR] The comment above `allowLocal` still reads "Accept only local connections" — the old `localOnly` semantics, now wrong.

Suggestion: reword to "Allow connections to/from local (site-local, link-local, loopback) addresses".

ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
```diff
@@ -45,6 +45,12 @@ class RecoverableModifierError(val message: String, val modifierId: ModifierId,
   def toThrowable: Throwable = this
 }
 
+/** Special case of recoverable error when parent header is not found in history.
+  * This allows the node to specifically request the missing parent header from peers.
+  */
+@SuppressWarnings(Array("org.wartremover.warts.Null"))
```
jozanek: [NIT] `@SuppressWarnings(...Null)` on `ParentHeaderNotFoundError` is vestigial — the class passes `None`, no nulls involved.

Suggestion: drop the annotation.

jozanek:
[MINOR] Release notes in the PR description are stale: stats claim 26 files/+1,133 while the diff is 75 files/+3,357, and the notes omit the `localOnly` → `allowLocal` rename, the incoming-connection cap (`NetworkController.scala`, hardcoded `OutgoingConnections = 8`), the 16 MB frame limit, fast block propagation, the miner sync gate, and the mempool replace-by-fee fix — operators read these notes.

Suggestion: regenerate the notes from the actual commit range.

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
jozanek: this was solved in the meantime

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
jozanek: solved

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -295,12 +304,14 @@ object CandidateGenerator extends ScorexLogging {
...
     candidateOpt: Option[Candidate],
-    txs: Seq[ErgoTransaction]
+    txs: Seq[ErgoTransaction],
+    minerPk: ProveDlog
   ): Boolean = {
     candidateOpt.isDefined && candidateOpt.exists { c =>
-      txs.isEmpty || (txs.size == c.txsToInclude.size && txs.forall(
-        c.txsToInclude.contains
-      ))
+      c.externalVersion.pk == minerPk &&
```
jozanek, line 319: [MINOR] The cache is a single slot now keyed by pk, so alternating requests with different pks (e.g. the node's own miner polling `/mining/candidate` while an external party polls `/mining/candidateWithTxsAndPk`) miss the cache every time and trigger full candidate generation (mempool packing + state proofs) per request. The internal `ChangedMempool` forced regeneration also always uses the default pk, evicting any custom-pk candidate.

Suggestion: consider a small bounded per-pk cache map, or document that the endpoint assumes a single active miner pk.

ergo-core/src/test/scala/org/ergoplatform/mining/AutolykosPowSchemeSpec.scala
```diff
@@ -35,9 +35,14 @@ class AutolykosPowSchemeSpec extends ErgoCorePropertyTest {
...
-        val newHeader2 = pow.checkNonces(ver, hbs, msg2, sk, x, b, N, 0, 1000)
-          .map(s => h.copy(powSolution = s)).get
-        pow.validate(newHeader2) shouldBe 'failure
+        val invalidHeader2 = Iterator.iterate(0L)(_ + 1000L).take(10)
+          .flatMap { startNonce =>
+            pow.checkNonces(ver, hbs, msg2, sk, x, b, N, startNonce, startNonce + 1000)
+              .map(s => h.copy(powSolution = s))
+          }
+          .find(pow.validate(_).isFailure)
+
```
jozanek: [MINOR] The stabilization weakened the property: it now searches for a solution that fails validation (`.find(pow.validate(_).isFailure)`) instead of asserting that the found solution fails. If `checkNonces` ever produced a tampered-msg solution that incorrectly passes validation (a soundness bug), this test would skip past it and still go green; the old assertion would have caught it.

Suggestion: take the first solution found across the ranges (`.headOption` after the `flatMap`) and assert `pow.validate(_) shouldBe 'failure` on it — same flakiness fix, original assertion strength.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -295,12 +304,14 @@ object CandidateGenerator extends ScorexLogging {
   /** checks that current candidate block is cached with given `txs` */
```
jozanek: [NIT] `cachedFor`'s scaladoc still says "cached with given `txs`" — the new pk condition isn't mentioned.

Suggestion: update the doc line.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
odiseusme: Complementary pass over the v6.0.3 branch at f1a62b7a3, deliberately avoiding the areas @upwindstrategy listed and, after reading @jozanek’s review, dropping everything he already covered (several of my findings — the ErgoMiner sync-check threading issue, the NewBlockMined announce race, the OrderedTxPool scan cost — independently matched his; no need to repeat them). One finding on this branch appears not yet raised, plus one small question.
Disclosure: AI-assisted review (Claude doing the code reading under my direction); static analysis only, I did not build or run the test suite — so this is phrased as a question for the authors.
[MAJOR/MINOR — is this intended?] The local change at ErgoNodeViewHolder.scala:235 also disables the mainnet ADProofs fast path for remote blocks.
This is a second consequence of the same line @jozanek flagged for API-submitted blocks. On master the state-application callback was hardcoded pmodModify(lm.pmod, local = true); the branch threads the block’s origin flag instead. But this callback only ever receives state-regenerated ADProofs — UtxoState.applyModifier calls generate(LocallyGeneratedModifier(adProofs)) (UtxoState.scala:199) when applying a near-tip block that arrived without proofs, which is the normal case for blocks received from peers. The regenerated proofs are locally generated regardless of where the block came from.
Net effect: on mainnet UTXO nodes, regenerated ADProofs for remote blocks no longer hit the dumpToDb fast path (ErgoNodeViewHolder.scala:472, gated on local && MainNet) and instead go through the full history().append() path for essentially every applied block at the tip. Testnet has always run the slow path so nothing crashes, but this is an unflagged behavior/performance change on the block-application hot path of every mainnet UTXO node — it looks like a mechanical side effect of threading the new flag rather than a deliberate choice. If unintended, the fix is to keep the callback’s argument at local = true (the modifier’s origin) and carry the block’s origin separately for the LocalBlockApplied/RemoteBlockApplied split — which would also dovetail with @jozanek’s suggestion to derive Local/Remote from “was already announced” rather than from this flag.
[NIT — question] #2287’s title says the candidateWithTxsAndPk route is “in testnet”, but it has no network gate and is live on mainnet (behind withAuth, consistent with candidateWithTxsR). Just confirming mainnet exposure is intended.
I did not find other issues in targeted passes over the fast-broadcast path, the event split, the miner sync gate, the new mining route, and the mempool removal changes that were not already covered by @jozanek’s review.

kushti: Added NewBlockMined signal to BlocksApiRoute

ergo-core/src/main/scala/org/ergoplatform/settings/LaunchParameters.scala
kushti: Testnet with new parameters already alive while the old one is dead

src/main/scala/org/ergoplatform/http/api/MiningApiRoute.scala
kushti: description added to openapi.yaml

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
kushti: made FallbackToEqualPeerThreshold constant

src/test/scala/org/ergoplatform/nodeView/mempool/ErgoMemPoolSpec.scala
kushti: fixed

src/test/scala/scorex/core/network/NetworkControllerSpec.scala
kushti: fixed

build.sbt
kushti: done

src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
kushti: done

ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
kushti: done

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
kushti: updated

ergo-core/src/test/scala/org/ergoplatform/mining/AutolykosPowSchemeSpec.scala
kushti: the biggest issue is that the test is assuming a wrong thing after all, that for modified bytes validation of PoW must fail, which does not hold for low difficulty with noticeable probability. The question is how to rework the condition to a more meaningful one, and for now to avoid a flaky test we have this

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
kushti: @K-Singh what do you think on this? for now updated documentation just

src/main/resources/application.conf
kushti: updated

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
kushti: since v2 activation the test is not needed I guess, mining in >= v2 is proven by every block in the mainnet , testnet etc

src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
kushti: they can be different in principle, but for now not

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
kushti: changed to ErgoState[_]

src/main/scala/org/ergoplatform/http/api/MiningApiRoute.scala
kushti: done

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
kushti: fixed

ergo-core/src/main/scala/org/ergoplatform/settings/Settings.scala
kushti: added fallback to localOnly

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
kushti: there should be no inconsistency between orderedTransactions & transactionsRegistry in principle, but careful investigation and fix left for a next version. Inconsistency does happen rarely

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
odiseusme: Second pass over the fix commits (head d3e93a13e): the changes look solid. Two items still appear open:


The ADProofs point above looks unaddressed: ErgoNodeViewHolder.scala:235 is still pmodModify(lm.pmod, local). The BlocksApiRoute change handles the broadcast side, but the fast-path question is separate, UtxoState.applyModifier regenerates ADProofs as LocallyGeneratedModifier for peer-received near-tip blocks, so they now pass local = false and miss the mainnet dumpToDb branch at L472. Resolve this thread if it's intended.
Neutral note, not my finding: the "mine after HF" test @jozanek flagged as deleted still appears absent on the current head, leaving that call to him.

jozanek: [MAJOR] The NewBlockMined publish resolves the original broadcast gap — API blocks are announced now, thanks. Leaving open for the second consequence on the same line: L235 now threads `local` into the state-application callback (master passed `local = true`). UtxoState regenerates missing ADProofs for near-tip blocks via `generate(LocallyGeneratedModifier(...))` (UtxoState.scala:199) — the normal peer-block path — so they reach pmodModify with `local = false` and miss the mainnet `dumpToDb` fast path (L472), hitting full `history().append()` for every block at the tip.

Suggestion: keep the callback at `local = true` and carry block origin separately for the Local/Remote split.

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
jozanek: The scan cost is independent of the inconsistency you're deferring: the O(n) `valuesIterator.exists(...)` fires on every registry MISS — every tx not in the pool, the common case when `remove` runs per-tx on block application — so it's paid on the NVH hot path even with no inconsistency.

Suggestion: add a third case — registry miss → `this` (no scan), leaving the fallback only for registry-hit + ordered-miss. 2-line change, orthogonal to the deferred fix.

ergo-core/src/test/scala/org/ergoplatform/mining/AutolykosPowSchemeSpec.scala
jozanek: Agreed — that invalidates my suggestion too (at low difficulty a tampered-msg solution can still pass, so asserting the first fails would be flaky). Minor: the added re-validate line is tautological (`.find` already selected it for failing) — fine to track a proper difficulty-gated rework separately.

src/test/scala/org/ergoplatform/mining/ErgoMinerSpec.scala
jozanek: Fair — v2 mining is covered by every live block and the v1→v2 boundary can't recur, so fine to leave deleted; downgrading. Feel free to resolve.

ergo-core/src/main/scala/org/ergoplatform/settings/LaunchParameters.scala
jozanek: Understood — the ask was just a release-notes line so operators wipe incompatible old testnet data dirs (new magic bytes/port). Feel free to resolve.

src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
jozanek: Makes sense as a forward-looking seam — a one-line comment saying so would prevent a future "simplification". Feel free to resolve.

src/main/scala/org/ergoplatform/http/api/BlocksApiRoute.scala
```diff
@@ -125,13 +126,21 @@ case class BlocksApiRoute(viewHolderRef: ActorRef, readersHolder: ActorRef, ergo
...
         viewHolderRef ! LocallyGeneratedModifier(_)
       }
 
+      // Immediately announce newly mined block to network BEFORE local application.
+      // This reduces propagation latency by avoiding the wait for NodeViewHolder
+      // to fully validate and apply the block. LocalBlockApplied arrives later
+      // and skips broadcast since the block was already announced.
+      // TODO: Consider switching to direct actor message for lower latency
+      //       instead of event bus publish.
+      context.eventStream.publish(NewBlockMined(block.header))
```
jozanek, line 142: [MAJOR] Broadcasts inv to all peers (via NewBlockMined) before the block is applied, on an unauthenticated endpoint that only checks self-declared-difficulty PoW — so a cheap low-difficulty POST is announced network-wide but never applies, and peers penalize us for non-delivery.

Suggestion: announce post-apply (on LocalBlockApplied), or add `withAuth` + full validation first.

build.sbt
```diff
@@ -174,18 +174,38 @@ inConfig(Linux)(
...
 
+// Cap how many forked test JVMs run concurrently.  sbt's default
+// concurrentRestrictions already contains Tags.limit(Tags.ForkedTestGroup, 1);
+// simply += adds a second, looser rule and the strictest wins, so the
+// 2-way parallelism would be a no-op.  We replace the default rule by
+// keeping the other defaults and setting ForkedTestGroup to 2.
+Global / concurrentRestrictions := Seq(
+  Tags.limitAll(8),
+  Tags.limit(Tags.ForkedTestGroup, 2),
+  Tags.limit(Tags.Test, 1)
```
jozanek: [NIT] `Tags.limit(Tags.Test, 1)` is inert (test tasks aren't tagged `Tags.Test` by default); if it ever bit, it'd re-serialize tests and defeat `ForkedTestGroup = 2`. Also `limitAll(8)` is hardcoded vs the CPU-scaled default, and `:=` drops `Tags.exclusiveGroup(Tags.Clean)`.

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -117,25 +118,42 @@ class ErgoMiner(
...
-        (1 to ergoSettings.nodeSettings.internalMinersCount) foreach { _ =>
-          ErgoMiningThread(
-            ergoSettings,
-            minerState.candidateGeneratorRef,
-            minerState.secretKeyOpt.get.w
-          )(context)
+      // Check if blockchain is synced before starting mining
+      readersHolderRef ! GetReaders
+
+    case Readers(historyReader, _, _, _) =>
```
jozanek, line 133: [NIT] If multiple StartMining/FullBlockApplied fire before sync, extra `Readers` replies arrive after `become(started)` and fall through `started`'s catch-all (L193-194), logging the whole `Readers(h,s,m,w)` tuple as "Unexpected message".

Suggestion: add `case _: Readers =>` in `started`.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
odiseusme: Thanks @upwindstrategy. Splitting the two so neither gets lost: “announce once servable” is the synchronizer/broadcast side. The piece I flagged is storage-path and independent of announce — regenerated ADProofs for peer-received near-tip blocks reach pmodModify with local = false (the remote path hardcodes it, ErgoNodeViewHolder.scala:348), so at L472 they fall out of the mainnet dumpToDb fast path into history().append(). On master that callback was local = true, so remote ADProofs used dumpToDb too — it’s a concrete change in the write path, not announce timing.
To pre-empt “does it matter”: no correctness impact — append still routes ADProofs to the same historyStorage.insert via justPutToHistory (same key, retrievable identically). The cost is that append re-runs applicableTry + PayloadValidator + serialization on a proof the node just generated and trusts, which dumpToDb deliberately skips. So it’s a per-block perf regression on the mainnet hot path, plus an asymmetry where self-mined ADProofs keep the fast path but peer-received ones don’t. Severity MINOR, not a blocker.

src/main/scala/org/ergoplatform/settings/ErgoSettings.scala
kushti: comment added

src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
kushti: added

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
kushti: added

build.sbt
kushti: replaced with 

```
Global / concurrentRestrictions := Seq(
  Tags.limitAll(math.max(1, java.lang.Runtime.getRuntime.availableProcessors())),
  Tags.limit(Tags.ForkedTestGroup, 2),
  Tags.exclusiveGroup(Tags.Clean)
)
```

src/main/scala/org/ergoplatform/http/api/BlocksApiRoute.scala
kushti: authentication added. I think it is node operator's responsibility to post a valid block (by eg following candidategenerator logic),

kushti:
> [MINOR] Release notes in the PR description are stale: stats claim 26 files/+1,133 while the diff is 75 files/+3,357, and the notes omit the `localOnly` → `allowLocal` rename, the incoming-connection cap (`NetworkController.scala`, hardcoded `OutgoingConnections = 8`), the 16 MB frame limit, fast block propagation, the miner sync gate, and the mempool replace-by-fee fix — operators read these notes.
> 
> Suggestion: regenerate the notes from the actual commit range.
description updated

regenerated

ergo-core/src/test/scala/org/ergoplatform/mining/AutolykosPowSchemeSpec.scala
kushti: tautological check removed

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
kushti: local = true on line 235 got back 

I also made a new issue , https://github.com/ergoplatform/ergo/issues/2413 , to simplify logic around ADProofs dumping

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
kushti: the assumption here is that local processing is finishing before network request (which takes at least tens of ms). But it would be good to cache anyway, made an issue for that https://github.com/ergoplatform/ergo/issues/2414

src/main/scala/org/ergoplatform/network/message/MessageSerializer.scala
kushti: Updated (MessageConstants , line 16-26)

kushti:
> [MINOR] Release notes in the PR description are stale: stats claim 26 files/+1,133 while the diff is 75 files/+3,357, and the notes omit the `localOnly` → `allowLocal` rename, the incoming-connection cap (`NetworkController.scala`, hardcoded `OutgoingConnections = 8`), the 16 MB frame limit, fast block propagation, the miner sync gate, and the mempool replace-by-fee fix — operators read these notes.
> 
> Suggestion: regenerate the notes from the actual commit range.

regenerated, stats removed also

jozanek (approved):
LGTM!

PR #2293, opened 2026-01-28
Candidate for 6.0.4 release

kushti:
* #2298 
* #2323 
* #2344
* #2376 
* #2383
* #2387 
* #2388 
* #2391
* #2401
* #2415
* #2439
* #2442 
* #2445 
* #2448
* #2459
* #2460 
* #2471
* #2472


jozanek (commented):
Reviewed the full 6.0.4 candidate diff. No protocol/consensus concerns: no serializer or validation-rule changes; the new `txIds` on `LocalBlockApplied`/`RemoteBlockApplied` is internal event-stream only, never on the wire; `appVersion`/openapi version stamps are consistent.

Findings: **1 MAJOR, 3 MINOR, 2 NIT** — inline. The same findings are posted on #2416, which builds on this branch, so fixing them once on `v6.0.4` and syncing forward resolves both.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -470,10 +478,18 @@ trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
...
+      val extendsIndexedTip = nextHeaderOpt.forall { header =>
+        state.indexedHeight == 0 || state.indexedHeaderId.contains(header.parentId)
+      }
+      if (extendsIndexedTip) {
+        val newState = index(state.incrementIndexedHeight)
+        if (modCount >= saveLimit) saveProgress(newState)
+        context.become(receive.orElse(loaded(newState)))
+        self ! Index()
+      } else {
+        log.info("Deferring catch-up because the next header does not extend the indexed tip")
```
jozanek, line 491: **MAJOR** After "Deferring catch-up" no retry is scheduled, and while `caughtUp = false` the actor has no `FullBlockApplied` handler — only a `Rollback` event can resume indexing. If a headers-only fork briefly becomes the best header chain and then loses without the full-block chain ever switching (the losing side of a miner race), no `Rollback` is published and the indexer stays stalled until restart, dropping all further `FullBlockApplied` events.

Suggestion: `case _: FullBlockApplied if !state.caughtUp && !state.rollbackInProgress => self ! Index()`, or re-schedule `Index()` with a delay instead of only logging.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -470,10 +478,18 @@ trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
   protected def loaded(state: IndexerState): Receive = {
 
     case Index() if !state.caughtUp && !state.rollbackInProgress =>
-      val newState = index(state.incrementIndexedHeight)
-      if (modCount >= saveLimit) saveProgress(newState)
-      context.become(receive.orElse(loaded(newState)))
-      self ! Index()
+      val nextHeaderOpt = history.bestHeaderAtHeight(state.indexedHeight + 1)
```
jozanek, line 481: **MINOR** This adds a `bestHeaderAtHeight(h + 1)` read per block during catch-up, and `index()` re-reads `bestHeaderIdAtHeight(height)` at line 381 (`headerOpt` is `None` on this path) — two redundant storage reads per block on the reindex hot path. Passing the already-fetched `nextHeaderOpt` into `index()` would reuse a single read.

src/main/scala/org/ergoplatform/http/api/InfoApiRoute.scala
```diff
@@ -28,12 +27,6 @@ case class InfoApiRoute(statsCollector: ActorRef,
         "lastMemPoolUpdateTime" -> nodeInfo.lastMemPoolUpdateTime.asJson
       ))
     })
-    } ~
-    (path(".well-known" / "ai-plugin.json") & get) {
```
jozanek, line 32: **MINOR** Removing `/openapi.yaml` and `/.well-known/ai-plugin.json` is clearly intentional (ChatGPT-plugin retirement), but it is a breaking removal of public endpoints — anything still fetching them gets a 404 after upgrade. Worth an explicit line in the 6.0.4 release notes.

src/it/scala/org/ergoplatform/it/OpenApiSpec.scala
```diff
@@ -25,6 +25,9 @@ class OpenApiSpec extends AnyFlatSpec with IntegrationSuite {
     .withFallback(nodeSeedConfigs.head)
     .withFallback(allowLocalConfig)
 
+  // `lazy` so the container is only started when a test actually touches `node`.
+  // The single test below is currently `ignore`d (the openapi-checker image is gone),
+  // so without `lazy` we would start and tear down a node for nothing.
   lazy val node: Node = docker.startDevNetNode(offlineGeneratingPeer).get
```
jozanek, line 31: **MINOR** The spec's only test remains `ignore`d (the openapi-checker image is gone), so this suite passes CI while providing zero signal — the new comment documents the situation but keeps the dead spec. Consider deleting the spec or reviving the check with a maintained validator image.

src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
```diff
@@ -41,6 +41,32 @@ class OrderedTxPool(val orderedTransactions: TreeMap[WeightedTxId, UnconfirmedTr
 
   private val mempoolCapacity = settings.nodeSettings.mempoolCapacity
 
+  private def withoutTransaction(id: ModifierId): TreeMap[WeightedTxId, UnconfirmedTransaction] = {
+    // Keep healthy mutations logarithmic; scan by ID only after cardinality diverges.
```
jozanek, line 45: **NIT** The size-equality heuristic takes the fast path on *compensating* corruption — one duplicate key plus one orphaned entry leaves the sizes equal, so a duplicate would survive `withoutTransaction`. Fine as best-effort self-healing, but worth extending the comment to note that limitation.

src/test/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexerSpecification.scala
```diff
@@ -319,16 +352,15 @@ class ExtraIndexerSpecification extends ErgoCorePropertyTest {
...
 
-  property("tokens dont disappear when rolling back with orphan block") {
+  property("indexes replacement blocks after rolling back an orphan block") {
     indexer ! CreateDB(HEIGHT)
     indexer ! Index()
-    lock.lock()
-    done.await()
+    awaitCondition(done)
     indexer ! GenerateBetterChainTip()
     lock.lock()
```
jozanek, line 361: **NIT** `awaitCondition` fixes the lock discipline (the old pattern never unlocks after `await()`), but this test still uses bare `lock.lock(); created.await()` in two places. Worth finishing the migration to `awaitCondition(created)` here too.

PR #2304, opened 2026-05-02
Docs: Comprehensive Tor Manual & JVM NIO SOCKS Bypass Disclosure (Closes #970)

### LINKED ISSUE #970: Check how to run the node behind Tor and write a manual on that
Manual on how to run Bitcoin node behind Tor https://en.bitcoin.it/wiki/Tor .

Check how to run Ergo node behind Tor and write a how-to manual on that.

a-shannon:
## Summary

Comprehensive Tor manual for running an Ergo node with network-level privacy. Addresses **#970**.

## ⚠️ Critical Security Discovery

**The commonly recommended `-DsocksProxyHost` JVM flags — including those in `ergo-installer.sh --tor` (line 149) — do NOT secure P2P traffic.**

Ergo uses Akka IO TCP → Java NIO `SocketChannel`, which silently ignores JVM SOCKS proxy properties. All P2P gossip leaks in plaintext. See [JDK-8033696](https://bugs.openjdk.org/browse/JDK-8033696).

## Deployment Options

| Option | Method | Security Level |
|--------|--------|----------------|
| **A (Recommended)** | Docker shared namespace + Tor TransPort + iptables | Production-grade |
| **B** | Whonix / Tor Gateway VM | Gold standard |
| **C** | iptables + dedicated user (with caveats) | Advanced / Leak-prone |

## Defense-in-Depth Architecture (Option A)

```
NAT OUTPUT chain (order critical):
  1. uid=tor     → RETURN     Tor reaches internet directly
  2. DNS :53     → REDIRECT   BEFORE loopback (prevents 127.0.0.11 Docker DNS leak)
  3. loopback    → RETURN     REST API on 127.0.0.1:9053 works
  4. ALL TCP     → REDIRECT   P2P forced through Tor TransPort :9040

FILTER OUTPUT chain:
  5. ESTABLISHED → ACCEPT     API responses via bridge IP survive
  6. uid=tor     → ACCEPT     Tor's own traffic
  7. loopback    → ACCEPT     REDIRECT'd packets to 127.0.0.1
  8. ALL ELSE    → DROP       Leak prevention safety net

IPv6 Kill-Switch:
  9. ip6tables   → DROP ALL   JVM can't bypass via IPv6
  10. _JAVA_OPTIONS: -Djava.net.preferIPv4Stack=true

Production Hardening:
  11. healthcheck gate         Prevents startup race condition IP leak
  12. su-exec (not sudo)       Alpine PID 1 security standard
  13. Valid apiKeyHash          Blake2b256("hello") placeholder prevents crash
```

## Key Security Findings Documented

| Finding | Impact | Section |
|---------|--------|---------|
| Java NIO ignores `-DsocksProxyHost` | P2P leaks in plaintext | §2 |
| `torsocks` unreliable with JVM/JNI | Silent traffic leaks | §2 |
| `.onion` in `declaredAddress` → crash | `UnknownHostException` | §7 |
| UPnP exposes real IP | Auto-doxxing | §8, §9 |
| Docker DNS at `127.0.0.11` | DNS leak via loopback exemption | §4 |
| `systemd-resolved` IPC bypasses UID rules | DNS leak on bare-metal | §6 |
| Docker `depends_on` race condition | 2-3 sec startup IP leak | §4 |
| IPv6 bypasses `iptables` entirely | Full traffic leak | §4, §6 |
| `iptables` rules are ephemeral | Post-reboot full leak | §6 |

## Scope

- **1 new file:** `docs/tor.md` (630 lines)
- **Zero code changes** — documentation only

## Note to Maintainers

The `-DsocksProxyHost` flags in `ergo-installer.sh` (line 149) should be reviewed. I can open a follow-up PR for the installer fix.

Closes #970

### DIFF
--- a/docs/tor.md
+++ b/docs/tor.md
@@ -1,0 +1,656 @@
+# Running an Ergo Node Behind Tor
+
+This guide explains how to run an Ergo node with its network traffic routed through
+[Tor](https://www.torproject.org/), providing network-level privacy for node operators.
+
+It covers threat modeling, deployment options, Tor Hidden Services, and critical
+operational security (OpSec) considerations specific to the Ergo JVM/Scorex/Akka stack.
+
+> **Reference:** This guide addresses [issue #970](https://github.com/ergoplatform/ergo/issues/970)
+> and is modeled after Bitcoin Core's [`doc/tor.md`](https://github.com/bitcoin/bitcoin/blob/master/doc/tor.md).
+
+---
+
+## Table of Contents
+
+1. [Why Run Behind Tor?](#1-why-run-behind-tor)
+2. [⚠️ CRITICAL: The JVM SOCKS Proxy Trap](#2--critical-the-jvm-socks-proxy-trap)
+3. [Prerequisites](#3-prerequisites)
+4. [Option A — Docker Network Isolation (Recommended)](#4-option-a--docker-network-isolation-recommended)
+5. [Option B — Whonix / Tor Gateway VM](#5-option-b--whonix--tor-gateway-vm)
+6. [Option C — iptables Transparent Proxy (Advanced)](#6-option-c--iptables-transparent-proxy-advanced)
+7. [Tor Hidden Service (Onion Address)](#7-tor-hidden-service-onion-address)
+8. [Ergo-Specific Configuration Reference](#8-ergo-specific-configuration-reference)
+9. [Security Considerations](#9-security-considerations)
+10. [Troubleshooting](#10-troubleshooting)
+11. [Further Reading](#11-further-reading)
+
+---
+
+## 1. Why Run Behind Tor?
+
+Running an Ergo node behind Tor provides several benefits:
+
+- **IP Privacy:** Your node's real IP address is hidden from peers, preventing observers
+  from linking your node to your physical location or identity.
+- **Censorship Resistance:** In jurisdictions that restrict blockchain network traffic,
+  Tor can bypass ISP-level filtering.
+- **Eclipse Attack Mitigation:** When used in combination with clearnet connections
+  (dual-stack), Tor makes it harder for an attacker to isolate your node.
+
+> **Note:** Tor adds latency to network communications (typically 200–800 ms per hop).
+> For **mining nodes** that need minimal latency for block propagation, this can lead to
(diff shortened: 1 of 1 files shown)


a-shannon:
## ⚠️ Force-Pushed Update — Critical Security Discovery

During a final deep-dive OpSec review, I discovered that the original version of this guide (and the current `ergo-installer.sh --tor` script) recommends a mechanism that **does not secure P2P traffic**.

### The Problem

Ergo uses Akka IO TCP, which internally operates on **Java NIO** (`java.nio.channels.SocketChannel`). The standard JVM SOCKS flags (`-DsocksProxyHost`) only apply to the legacy blocking `java.net.Socket` API. Java NIO channels bypass the SOCKS implementation entirely and connect via direct native OS system calls.

**Result:** When a user runs Ergo with `-DsocksProxyHost=127.0.0.1` (as in `ergo-installer.sh --tor`), HTTP API calls may be proxied, but **all P2P gossip traffic — blocks, transactions, peer discovery — connects directly to peers in plaintext**, exposing the operator's real IP.

This is a well-documented JDK architectural limitation: [JDK-8033696](https://bugs.openjdk.org/browse/JDK-8033696).

### What Changed in This Force-Push

The guide has been completely rewritten to:

1. **Ban `-DsocksProxyHost`** with a prominent warning explaining the NIO bypass.
2. **Ban `torsocks`** — `LD_PRELOAD` hooks are unreliable with the JVM (JNI native calls can bypass libc interception, causing silent leaks or crashes).
3. **Mandate network-level isolation** as the only safe approach:
   - **Option A (Recommended):** Docker with Tor `TransPort` transparent proxy
   - **Option B:** Whonix / Tor Gateway VM (gold standard)
   - **Option C:** `iptables` transparent proxy with a dedicated system user

These approaches enforce Tor routing **below** the JVM, making it impossible for Java NIO to bypass the proxy regardless of which socket API it uses internally.

The `.onion` crash warning, UPnP disable, and Hidden Service documentation are preserved from the original version.

---

*Note to maintainers: The `-DsocksProxyHost` flags in `ergo-installer.sh` (line 149) should also be reviewed, as they currently provide a false sense of security for `--tor` users. I can open a separate PR for the installer fix if desired.*

PR #2306, opened 2026-05-12
P2P logic improvements

kushti:
In this PR : 
* removing local addresses from database (if scorex.network.localOnly = false, which is default setting)
* explicit limit for incoming connections (max(maxConnections - 8, maxConnections / 2))
* more sanity checks  

also, tests and speedups for testing code from #2336 and #2344 are pulled to this PR

### DIFF
--- a/ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
@@ -45,6 +45,14 @@ class RecoverableModifierError(val message: String, val modifierId: ModifierId, 
   def toThrowable: Throwable = this
 }
 
+/** Special case of recoverable error when parent header is not found in history.
+  * This allows the node to specifically request the missing parent header from peers.
+  */
+@SuppressWarnings(Array("org.wartremover.warts.Null"))
+class ParentHeaderNotFoundError(val parentId: ModifierId, modifierId: ModifierId, modifierTypeId: NetworkObjectTypeId.Value)
+    extends RecoverableModifierError(s"Parent header with id $parentId is not defined", modifierId, modifierTypeId, None) {
+  def parentHeaderId: ModifierId = parentId
+}
 
 /** Composite error class that can hold more than one modifier error inside. This was not made a `ModifierError` instance
   * intentionally to prevent nesting `MultipleErrors` to `MultipleErrors`
--- a/ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
@@ -50,9 +50,7 @@ class RecoverableModifierError(val message: String, val modifierId: ModifierId, 
   */
 @SuppressWarnings(Array("org.wartremover.warts.Null"))
 class ParentHeaderNotFoundError(val parentId: ModifierId, modifierId: ModifierId, modifierTypeId: NetworkObjectTypeId.Value)
-    extends RecoverableModifierError(s"Parent header with id $parentId is not defined", modifierId, modifierTypeId, None) {
-  def parentHeaderId: ModifierId = parentId
-}
+    extends RecoverableModifierError(s"Parent header with id $parentId is not defined", modifierId, modifierTypeId, None)
 
 /** Composite error class that can hold more than one modifier error inside. This was not made a `ModifierError` instance
   * intentionally to prevent nesting `MultipleErrors` to `MultipleErrors`
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -1869,10 +1869,10 @@ components:
       properties:
         pass:
           type: string
-          description: Password to encrypt wallet file with
+          description: Spending password used to encrypt the wallet secret storage on disk. Required on every /wallet/unlock afterwards. Not recoverable — losing it means restoring from the mnemonic.
         mnemonicPass:
           type: string
-          description: Optional pass to password-protect mnemonic seed
+          description: "Optional BIP-39 passphrase (a.k.a. 25th word) that is mixed into the seed derivation. Distinct from `pass`: it is NOT stored by the node and MUST be supplied again on restore. Empty/omitted means no passphrase."
 
     InitWalletResult:
       type: object
@@ -1892,16 +1892,16 @@ components:
       properties:
         pass:
           type: string
-          description: Password to encrypt wallet file with
+          description: Spending password used to encrypt the wallet secret storage on disk. Required on every /wallet/unlock afterwards. Not recoverable — losing it means restoring from the mnemonic.
         mnemonic:
           type: string
-          description: Mnemonic seed
+          description: BIP-39 mnemonic seed phrase (space-separated words) the wallet will be restored from.
         mnemonicPass:
           type: string
-          description: Optional pass to password-protect mnemonic seed
+          description: Optional BIP-39 passphrase (a.k.a. 25th word) mixed into the seed derivation. Distinct from `pass` and not stored by the node. Must match the value used when the mnemonic was originally generated, otherwise a different wallet will be restored silently.
         usePre1627KeyDerivation:
           type: boolean
-          description: use incorrect(previous) BIP32 key derivation (see https://github.com/ergoplatform/ergo/issues/1627 for details). It's recommended to set to 'true' if the original wallet was created by ergo node before v4.0.105.
+          description: Use the legacy (pre-EIP-3) BIP-32 key derivation that was the default before node v4.0.105. Set `true` only when restoring a wallet originally created on an older node; new wallets should use `false`. See https://github.com/ergoplatform/ergo/issues/1627 for details.
 
     CheckWallet:
       type: object
@@ -1931,7 +1931,7 @@ components:
       properties:
         derivationPath:
           type: string
-          description: Derivation path for a new secret to derive
+          description: BIP-32 derivation path of the secret to derive, relative to the wallet root. Apostrophe denotes hardened indices. Example below uses a short path; a typical EIP-3 path looks like `m/44'/429'/0'/0/3`.
           example: 'm/1/2'
 
     DeriveKeyResult:
@@ -1950,7 +1950,7 @@ components:
       properties:
         derivationPath:
           type: string
-          description: Derivation path of the resulted secret
+          description: BIP-32 derivation path that was picked for the newly derived key.
           example: 'm/1/2'
         address:
           $ref: '#/components/schemas/ErgoAddress'
@@ -4188,6 +4188,11 @@ paths:
       security:
         - ApiKeyAuth: [api_key]
       summary: Initialize new wallet with randomly generated seed
+      description: |
+        Initializes a brand new wallet by generating a fresh BIP-39 mnemonic seed phrase on the server.
+        The seed is encrypted on disk with `pass` and the returned mnemonic is the ONLY copy — the caller
+        must back it up, the node will not show it again. Fails if a wallet file already exists;
+        use `/wallet/restore` to overwrite from a known mnemonic instead.
       operationId: walletInit
       tags:
         - wallet
@@ -4216,6 +4221,11 @@ paths:
       security:
         - ApiKeyAuth: [api_key]
       summary: Create new wallet from existing mnemonic seed
+      description: |
+        Creates a wallet from a mnemonic the caller already has (e.g. one previously returned by
+        `/wallet/init`, or generated elsewhere). The mnemonic is converted to a BIP-32 seed
+        (optionally salted with `mnemonicPass`) and the resulting secret is encrypted on disk
+        with `pass`. Note the `usePre1627KeyDerivation` flag — see its field description.
       operationId: walletRestore
       tags:
         - wallet
@@ -4393,6 +4403,11 @@ paths:
       security:
         - ApiKeyAuth: [api_key]
       summary: Derive new key according to a provided path
+      description: |
+        Derives a new secret key at the BIP-32 path supplied by the caller and returns its P2PK address.
+        The wallet must be unlocked. The path is relative to the wallet's root and uses standard
+        BIP-32 notation (e.g. `m/44'/429'/0'/0/1`, with apostrophe denoting hardened indices).
+        The derived key is stored by the wallet so subsequent calls to `/wallet/addresses` will include it.
       operationId: walletDeriveKey
       tags:
         - wallet
@@ -4421,6 +4436,10 @@ paths:
       security:
         - ApiKeyAuth: [api_key]
       summary: Derive next key
+      description: |
+        Derives the next sequential key after the highest-index key already known to the wallet,
+        following EIP-3 (`m/44'/429'/0'/0/N`, where `N` is the next unused account-chain index).
+        Convenience wrapper over `/wallet/deriveKey` that picks the path for you. Wallet must be unlocked.
       operationId: walletDeriveNextKey
       tags:
         - wallet
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -66,6 +66,15 @@ class CandidateGenerator(
     log.info(
       s"New block ${newBlock.id} w. nonce ${Longs.fromByteArray(newBlock.header.powSolution.n)}"
     )
+
+    // Immediately announce newly mined block to network BEFORE local application.
+    // This reduces propagation latency by avoiding the wait for NodeViewHolder
+    // to fully validate and apply the block. LocalBlockApplied arrives later
+    // and skips broadcast since the block was already announced.
+    // TODO: Consider switching to direct actor message for lower latency
+    //       instead of event bus publish.
+    context.system.eventStream.publish(NewBlockMined(newBlock.header))
+
     viewHolderRef ! LocallyGeneratedModifier(newBlock.header)
     val sectionsToApply = if (ergoSettings.nodeSettings.stateType == StateType.Digest) {
       newBlock.blockSections
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -223,8 +223,7 @@ class CandidateGenerator(
             completeBlock(state.cachedPreviousCandidate.get.candidateBlock, solution)
           }
         log.info(s"New block mined, header: ${newBlock.header}")
-        ergoSettings.chainSettings.powScheme
-          .validate(newBlock.header) match {
+        ergoSettings.chainSettings.powScheme.validate(newBlock.header) match {
           case Success(_) =>
             sendToNodeView(newBlock)
             context.become(initialized(state.copy(solvedBlock = Some(newBlock))))
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -224,9 +224,8 @@ class CandidateGenerator(
           }
         log.info(s"New block mined, header: ${newBlock.header}")
         ergoSettings.chainSettings.powScheme
-          .validate(newBlock.header)
-          .map(_ => newBlock) match {
-          case Success(newBlock) =>
+          .validate(newBlock.header) match {
+          case Success(_) =>
             sendToNodeView(newBlock)
             context.become(initialized(state.copy(solvedBlock = Some(newBlock))))
             StatusReply.success(())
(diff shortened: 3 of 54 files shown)


src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
K-Singh: PoW validation is done twice in certain cases. Time difference is probably not as big as switching from event bus to direct messages, but maybe it will shave off a few ms if propagation needs to be done as fast as possible. Probably will only have an effect when N is at IncreasementHeightMax and block validation is successful (exception will end things earlier)

src/main/scala/scorex/core/utils/NetworkUtils.scala
```diff
@@ -24,4 +24,17 @@ object NetworkUtils {
       externalNodeAddress.contains(peerAddress)
   }
 
+  /**
+    * Check if a given address is a local address (site-local, link-local, or loopback).
+    * When localOnly is true, all addresses are allowed.
+    */
+  def checkLocalOnly(address: InetSocketAddress, localOnly: Boolean): Boolean = {
```
K-Singh: Small point but naming here is a bit confusing. I know the `localOnly` name of the parameter is coming from the config setting, but it makes either the if statement or the method seem wrong.

K-Singh (commented):
In general looks good to me, noticed some small issues imo but nothing huge as long as the failed tests are solved. Incoming connections limit also looked good.

src/main/resources/api/openapi.yaml
jozanek, line 1925: No update of description here?

src/main/scala/scorex/core/utils/NetworkUtils.scala
```diff
@@ -24,4 +24,17 @@ object NetworkUtils {
...
+    if (!localOnly) {
+      val addr = address.getAddress
+      addr.isSiteLocalAddress || addr.isLinkLocalAddress || addr.isLoopbackAddress
+    } else {
+      false
+    }
+  }
```
jozanek, line 38: Operators with a wallet/explorer/sibling node on `127.0.0.1` will be rejected after upgrade. Possible ideas:
- add loopback filter behind a setting
- flag as a breaking change in release notes
- add a test where a loopback peer is admitted when `localOnly=true` and rejected when `false`

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -1426,7 +1486,23 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+              val randomPeer = olderPeers(scala.util.Random.nextInt(olderPeers.size))
+              requestBlockSection(Header.modifierTypeId, Seq(parentId), randomPeer)
+            } else {
+              logger.warn(s"No older peer available to download missing parent header $parentId for modifier $modId")
+            }
+          }
+          deliveryTracker.setUnknown(modId, modTypeId)
+        case _ =>
+          deliveryTracker.setUnknown(modId, modTypeId)
+      }
```
jozanek, line 1502: [Minor] RecoverableFailedModification handler calls `deliveryTracker.status(...)`, which does `historyReader.modifierById(parentId)` (DB lookup) inside receive. Spammy orphans amplify dispatcher load. `historyReader` here could be in-memory view, or we could move the lookup to a Future on a dedicated dispatcher and pipeTo(self) the result.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -496,28 +494,45 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+        // Header not yet tracked — transition through Requested to Received
+        // (we already have the full object from the sync message, so no download needed).
+        // This allows the delivery tracker to properly manage the modifier lifecycle:
+        // Unknown -> Requested -> Received -> Held (on success) or
+        // Unknown -> Requested -> Received -> Unknown (on recoverable failure).
+        log.info(s"Applying valid syncInfoV2 header ${continuationHeader.encodedId}")
+        deliveryTracker.setRequested(Header.modifierTypeId, continuationHeader.id, peer, checksDone = 0) { _ =>
+          Cancellable.alreadyCancelled
+        }
+        deliveryTracker.setReceived(continuationHeader.id, Header.modifierTypeId, peer)
```
jozanek: You can add a `DeliveryTracker.setReceivedDirectly(id, type, peer)` that does the transition without the request/cancel dance.

ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
```diff
@@ -45,6 +45,14 @@ class RecoverableModifierError(val message: String, val modifierId: ModifierId,
   def toThrowable: Throwable = this
 }
 
+/** Special case of recoverable error when parent header is not found in history.
+  * This allows the node to specifically request the missing parent header from peers.
+  */
+@SuppressWarnings(Array("org.wartremover.warts.Null"))
+class ParentHeaderNotFoundError(val parentId: ModifierId, modifierId: ModifierId, modifierTypeId: NetworkObjectTypeId.Value)
+    extends RecoverableModifierError(s"Parent header with id $parentId is not defined", modifierId, modifierTypeId, None) {
+  def parentHeaderId: ModifierId = parentId
```
jozanek: Drop one of `parentId/parentHeaderId` if both must exist, make parentHeaderId a final def

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -873,12 +873,22 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
...
+        try {
+          result = Some(testProbe.expectMsgPF(candidateGenDelay) {
+            case StatusReply.Success(c: Candidate) => c
+          })
+        } catch {
+          case _: AssertionError =>
+            attempts -= 1
+        }
+      }
+      result.getOrElse(throw new AssertionError("Failed to get candidate after max retries"))
```
jozanek: Hides a real race. The comment admits "state not fully updated yet" and any genuine regression in candidate generation is silently swallowed up to 3 times. Replace with a deterministic wait: subscribe to whichever event signals the candidate refresh is ready, then send GenerateCandidate once.

src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
```diff
@@ -12,15 +12,17 @@ import org.ergoplatform.nodeView.state.wrapped.WrappedUtxoState
 import org.ergoplatform.nodeView.state.{StateType, UtxoState}
```
jozanek, line 12: Multiple `Thread.sleep(500) / Thread.sleep(100)` to "let the synchronizer initialize" or pace `CheckDelivery` floods. Wall-clock dependency, resulting in flaky under loaded CI.
Replace inits with `eventually { ... }` on an observable side-effect, replace pacing sleeps with `awaitAssert` keyed on tracker state.

src/test/scala/org/ergoplatform/network/DeliveryTrackerSpec.scala
```diff
@@ -71,4 +72,50 @@ class DeliveryTrackerSpec extends ErgoCorePropertyTest {
...
+  property("tracker should return Held status for modifiers in history") {
+    import org.ergoplatform.modifiers.history.header.Header
+    val tracker = DeliveryTracker.empty(settings)
+    val mid: ModifierId = ModifierId @@ "held_modifier"
+    val mTypeId: NetworkObjectTypeId.Value = NetworkObjectTypeId.fromByte(104)
+
+    // Create a mock ContainsModifiers that reports the modifier as held
+    val mockHistory = new ContainsModifiers[Header] {
+      override def modifierById(modifierId: ModifierId): Option[Header] =
+        if (modifierId == mid) Some(null) else None
```
jozanek: Can we use a real Header (from defaultHeaderGen.sample.get)?

src/test/scala/org/ergoplatform/network/DeliveryTrackerSpec.scala
```diff
@@ -71,4 +72,50 @@ class DeliveryTrackerSpec extends ErgoCorePropertyTest {
...
+    }
+
+    // Without history, modifier should be Unknown
+    tracker.status(mid, mTypeId, Seq.empty) shouldBe Unknown
+
+    // With history that contains the modifier, should be Held
+    tracker.status(mid, mTypeId, Seq(mockHistory)) shouldBe Held
+
+    // If modifier is in received cache, it should take precedence over Held
+    tracker.setReceived(mid, mTypeId, null)
```
jozanek: null could be replaced with `connectedPeerGen(ActorRef.noSender).sample.get`?

src/test/scala/scorex/core/network/NetworkControllerSpec.scala
```diff
@@ -0,0 +1,318 @@
...
+import java.net.InetSocketAddress
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+class NetworkControllerSpec extends ErgoCorePropertyTest {
+
+  import org.ergoplatform.network.peer.PeerManager.ReceivableMessages._
+  import org.ergoplatform.utils.ErgoNodeTestConstants._
+
+  private implicit val system: ActorSystem = ActorSystem("NetworkControllerSpec")
```
jozanek: never terminated

jozanek (commented):
2 generic questions related to PR as a whole:

- Was 16 MiB chosen against the max legitimate MessageSpec payload, or as a safety cap?
- Why split FullBlockApplied into three event types (LocalBlockApplied, RemoteBlockApplied, NewBlockMined) rather than a single event with a source flag? The split forces every existing subscriber to be re-examined; an enum keeps the surface stable.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
kushti: Thinking about fixing it in a separate PR, made an issue https://github.com/ergoplatform/ergo/issues/2346

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
kushti: done

src/test/scala/org/ergoplatform/network/DeliveryTrackerSpec.scala
kushti: done

src/test/scala/scorex/core/network/NetworkControllerSpec.scala
kushti: replaced with a fixture

kushti:
> 2 generic questions related to PR as a whole:
> 
>     * Was 16 MiB chosen against the max legitimate MessageSpec payload, or as a safety cap?
> 
>     * Why split FullBlockApplied into three event types (LocalBlockApplied, RemoteBlockApplied, NewBlockMined) rather than a single event with a source flag? The split forces every existing subscriber to be re-examined; an enum keeps the surface stable.



> 2 generic questions related to PR as a whole:
> 
>     * Was 16 MiB chosen against the max legitimate MessageSpec payload, or as a safety cap?
> 
>     * Why split FullBlockApplied into three event types (LocalBlockApplied, RemoteBlockApplied, NewBlockMined) rather than a single event with a source flag? The split forces every existing subscriber to be re-examined; an enum keeps the surface stable.

* the limit is reasonably bigger than max modifier size (8M), defined by `ModifiersSpec.maxMsgSizeWithReserve`. Also enough to have reasonable guaranteed max memory consumption  by p2p messaging buffers (30 peers by default * 16M = 480M). 

* NewBlockMined is just optimistic signal announcing block immediately after successful generation. Then the node is still sending the block through the full validation route, Local/Remote block applied is needed for p2p layer (which is accounting that local block was announced already). Made as two classes derived from a common trait for simpler pattern matching

src/test/scala/org/ergoplatform/network/DeliveryTrackerSpec.scala
kushti: done

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
kushti: hmm shoudn be the case as :
```
 def setUnknown(id: ModifierId, modifierTypeId: NetworkObjectTypeId.Value): Unit =
    tryWithLogging {
      val oldStatus = status(id, modifierTypeId, Seq.empty)
```      
 no historyReader provided, status works with Seq.empty just (so in-memory structures)

ergo-core/src/main/scala/org/ergoplatform/validation/ModifierError.scala
kushti: parentHeaderId dropped

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
kushti: fixed

src/test/scala/org/ergoplatform/network/ErgoNodeViewSynchronizerSpecification.scala
kushti: Thread.sleep eliminated

src/main/resources/api/openapi.yaml
kushti: this line is not changed in 6.0.3 / https://github.com/ergoplatform/ergo/pull/2318/changes as well

src/main/scala/scorex/core/utils/NetworkUtils.scala
kushti: ah but this check existed before, and it is for p2p networking only. 

Previously, connection to local addresses was prevented. Now, it is peers announcing local network addresses also not stored in a database.

src/main/scala/scorex/core/utils/NetworkUtils.scala
kushti: renamed the function to isLocal , and config setting to allowLocal

jozanek:
> > 2 generic questions related to PR as a whole:
> > ```
> > * Was 16 MiB chosen against the max legitimate MessageSpec payload, or as a safety cap?
> > 
> > * Why split FullBlockApplied into three event types (LocalBlockApplied, RemoteBlockApplied, NewBlockMined) rather than a single event with a source flag? The split forces every existing subscriber to be re-examined; an enum keeps the surface stable.
> > ```
> 
> > 2 generic questions related to PR as a whole:
> > ```
> > * Was 16 MiB chosen against the max legitimate MessageSpec payload, or as a safety cap?
> > 
> > * Why split FullBlockApplied into three event types (LocalBlockApplied, RemoteBlockApplied, NewBlockMined) rather than a single event with a source flag? The split forces every existing subscriber to be re-examined; an enum keeps the surface stable.
> > ```
> 
> * the limit is reasonably bigger than max modifier size (8M), defined by `ModifiersSpec.maxMsgSizeWithReserve`. Also enough to have reasonable guaranteed max memory consumption  by p2p messaging buffers (30 peers by default * 16M = 480M).
> * NewBlockMined is just optimistic signal announcing block immediately after successful generation. Then the node is still sending the block through the full validation route, Local/Remote block applied is needed for p2p layer (which is accounting that local block was announced already). Made as two classes derived from a common trait for simpler pattern matching

Both make sense, thanks.

On the 16 MiB cap: would it be worth pinning it as 2 * ModifiersSpec.maxMsgSizeWithReserve (or wrapping the constant near ModifiersSpec) so the relationship is documented in code and the per-peer × maxConnections worst-case is derivable rather than implicit? Not blocking — just a note for the next reader who wonders where 16 came from.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
jozanek: The snippet you quoted is `setUnknown` (which uses `Seq.empty` — agreed, no DB hit). My concern was the line just above it in the `ParentHeaderNotFoundError` branch:

```scala
if (deliveryTracker.status(parentId, Header.modifierTypeId, Seq(historyReader)) == ModifiersStatus.Unknown) { ... }
```

That one passes `Seq(historyReader)`, so `status` calls `historyReader.modifierById(parentId)`. Does that hit the in-memory header cache first, or always go through `historyStorage`? If it's always disk, an orphan flood (peer feeds us N orphans whose parents we don't have) lands N LevelDB lookups on the synchronizer's dispatcher thread.

Happy to drop this if it's cache-served. Otherwise, would `historyReader.contains(parentId)` (no value fetch) be enough here, or move the lookup off the actor with `pipeTo(self)`?

src/main/resources/api/openapi.yaml
jozanek: The line itself isn't touched by this PR. The reason I flagged it: this PR rewrites the `pass` and `mnemonicPass` descriptions on `InitWallet` / `RestoreWallet` to be much richer. The same fields on other wallet schemas still carry the old one-liner.

src/main/scala/scorex/core/utils/NetworkUtils.scala
jozanek: Got it on the scope expansion (DB filtering too) and the `isLocal` / `allowLocal` rename.

One thing I don't think is covered: the *predicate itself* also grew. Previously it was `isSiteLocalAddress || isLinkLocalAddress`; this PR adds `isLoopbackAddress`. So a setup with two Ergo nodes on the same host (or a wallet/explorer reaching the full node over `127.0.0.1`), running with the default `allowLocal = false`, will be filtered out after the upgrade — that wasn't true before.

Two questions:
1. Was the loopback addition intentional? (If yes, no objection — just please call it out in release notes.)
2. If yes, would it make sense to split into two settings (`allowSiteLocal` vs `allowLoopback`) so co-hosted setups can keep peering without also accepting LAN-broadcast peers? Or is that overkill — the typical operator on a single host will just set `allowLocal = true` and that's fine?

Happy with whatever resoluition, mostly want the change documented so existing operators aren't surprised.

kushti: Adding loopback address was intentional , and I think loopback address should not be different from other local addresses

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
kushti: In-memory structures checked first : 

```
    if (received.get(modifierTypeId).exists(_.contains(modifierId))) Received
    else if (requested.get(modifierTypeId).exists(_.contains(modifierId))) Requested
    else if (invalidModifierCache.mightContain(modifierId)) Invalid
    else if (modifierKeepers.exists(_.contains(modifierId))) Held
    else Unknown
```

but also, there was database read before during header's validation. Here we have double-check which can be useful for extra-safery w. multithreading

src/main/resources/api/openapi.yaml
kushti: ?? this PR (2306) is not introducing changes in openapi.yaml against v6.0.3

kushti:
> > > 2 generic questions related to PR as a whole:
> > > ```
> > > * Was 16 MiB chosen against the max legitimate MessageSpec payload, or as a safety cap?
> > > 
> > > * Why split FullBlockApplied into three event types (LocalBlockApplied, RemoteBlockApplied, NewBlockMined) rather than a single event with a source flag? The split forces every existing subscriber to be re-examined; an enum keeps the surface stable.
> > > ```
> > 
> > 
> > > 2 generic questions related to PR as a whole:
> > > ```
> > > * Was 16 MiB chosen against the max legitimate MessageSpec payload, or as a safety cap?
> > > 
> > > * Why split FullBlockApplied into three event types (LocalBlockApplied, RemoteBlockApplied, NewBlockMined) rather than a single event with a source flag? The split forces every existing subscriber to be re-examined; an enum keeps the surface stable.
> > > ```
> > 
> > 
> > 
> > * the limit is reasonably bigger than max modifier size (8M), defined by `ModifiersSpec.maxMsgSizeWithReserve`. Also enough to have reasonable guaranteed max memory consumption  by p2p messaging buffers (30 peers by default * 16M = 480M).
> > * NewBlockMined is just optimistic signal announcing block immediately after successful generation. Then the node is still sending the block through the full validation route, Local/Remote block applied is needed for p2p layer (which is accounting that local block was announced already). Made as two classes derived from a common trait for simpler pattern matching
> 
> Both make sense, thanks.
> 
> On the 16 MiB cap: would it be worth pinning it as 2 * ModifiersSpec.maxMsgSizeWithReserve (or wrapping the constant near ModifiersSpec) so the relationship is documented in code and the per-peer × maxConnections worst-case is derivable rather than implicit? Not blocking — just a note for the next reader who wonders where 16 came from.

done 

```

  val MaxMessageSize: Int = ModifiersSpec.maxMsgSizeWithReserve * 2
```

per-peer × maxConnections is implicit as there is no easy way to check cumulative size of all the peer buffers I think

src/main/resources/api/openapi.yaml
jozanek: I am seeing +27-8 changes in openapi.yaml file 🤷

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
jozanek: Great, in-memory short-circuit makes the orphan-flood case fine. Thanks for sharing the code!

jozanek (approved):

Issue #331, opened 2018-06-29
Links to different best headers


catena2w:
Hide `bestHeader` from API, use `bestFullBlock` at `ErgoSyncInfo`, `PoPoW`, etc. creation.

duplicate of https://github.com/ergoplatform/ergo/issues/328

catena2w:
See https://github.com/ergoplatform/ergo/issues/525 for actual description

PR #2313, opened 2026-05-18
prevent bestFullBlock/bestHeader divergence on sibling forks

### LINKED ISSUE #525: Different FullBlock and Header at the same height.
Now it's possible to reproduce situation when at height `x` we have different best `FullBlock` and best `Header`.

jozanek:
Fixes #525. `processBetterChain` promoted any full block whose score beat the current `bestFullBlock`, even when its header wasn't on the best-header chain — leaving `bestFullBlock` and `bestHeader` on different forks at the same height.

Adds an `isInBestChain` precondition. Off-chain full blocks fall through to `nonBestBlock` (stored, not promoted) and get picked up if the header chain later switches to their fork. Synced and reorg paths unaffected.

## Follow-ups

- `commonBlock` reader (#328 option 1) — prior attempt in #330.
- Linearize header chain by full-block work (#331 / #328 option 2) — prior attempt in #333.
- Port `BestChainInconsistencySpec` from `origin/i525` as an `it:test`.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/FullBlockProcessor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/FullBlockProcessor.scala
@@ -82,7 +82,10 @@ trait FullBlockProcessor extends HeadersProcessor {
 
   private def processBetterChain: BlockProcessing = {
     case ToProcess(fullBlock, newModRow, Some(newBestBlockHeader), _)
-      if bestFullBlockOpt.nonEmpty && isBetterChain(newBestBlockHeader.id) && isLinkable(fullBlock.header) =>
+      if bestFullBlockOpt.nonEmpty &&
+        isBetterChain(newBestBlockHeader.id) &&
+        isInBestChain(newBestBlockHeader.id) &&
+        isLinkable(fullBlock.header) =>
 
       val prevBest = bestFullBlockOpt.get
 
--- a/src/test/scala/org/ergoplatform/nodeView/history/VerifyADHistorySpecification.scala
+++ b/src/test/scala/org/ergoplatform/nodeView/history/VerifyADHistorySpecification.scala
@@ -63,6 +63,31 @@ class VerifyADHistorySpecification extends ErgoCorePropertyTest with NoShrink {
   }
 
 
+  property("bestFullBlock must stay on the best-header chain when a sibling full block arrives") {
+    var (history, _) = genHistory()
+
+    val commonChain = genChain(2, history)
+    history = applyChain(history, commonChain)
+    val ancestor = commonChain.last
+    history.bestFullBlockOpt.get.header shouldBe ancestor.header
+    history.bestHeaderOpt.get shouldBe ancestor.header
+
+    val a3 = genChain(1, ancestor).tail.head
+    Thread.sleep(2)
+    val b3 = genChain(1, ancestor).tail.head
+    b3.header.id should not be a3.header.id
+
+    history = applyHeaderChain(history, HeaderChain(Seq(a3.header)))
+    history.bestHeaderOpt.get shouldBe a3.header
+    history.bestFullBlockOpt.get.header shouldBe ancestor.header
+
+    history = applyBlock(history, b3)
+
+    history.bestHeaderOpt.get shouldBe a3.header
+    history.bestFullBlockOpt.get.header shouldBe ancestor.header
+    history.isInBestChain(history.bestFullBlockOpt.get.header) shouldBe true
+  }
+
   property("ErgoModifiersCache.findCandidateKey() should find headers in case of forks") {
     val modifiersCache = new ErgoModifiersCache(Int.MaxValue)
 


cafebedouin:
Prepared with Claude (Anthropic), model Claude Opus 5, building on earlier test-setup work with Claude Opus 4.8.

@jozanek, this is an external multi-node reproduction of #525, which the PR lists as a follow-up. It is not a port of `BestChainInconsistencySpec` and is offered only as evidence. A node's best full block and best header ended up on different forks at the same height in 8 of 9 runs on the v6.0.5 release jar, and in 0 of 9 runs with this PR's guard applied to v6.0.5. The setup and a script are below, in case they help with the `it:test` follow-up.

**Setup.** Four node processes run with `--devnet` on an isolated network.
- Two miners mine from a shared genesis, with internal-miner polling intervals of 4 s and 5 s and 500 ms of latency each way between them, so equal-height sibling blocks are common.
- Two non-mining followers each have no added latency to one miner and 250 ms each way to the other.
- No node is restarted.
- With equal polling intervals the two miners stayed on their own forks, because an equal-score tie keeps the first-seen block. Unequal intervals were used for that reason. These settings were chosen in a short calibration run on the release jar and fixed before any guarded run.

For 150 s, `/info` is polled on all four nodes as fast as they answer. A sample counts as divergent when `fullHeight == headersHeight` but `bestFullHeaderId != bestHeaderId`. An episode is one distinct divergent (best full block, best header) pair on one node.

| build | runs with divergence | divergent episodes | longest stretch with `fullHeight < headersHeight` | nodes agree 5 blocks below the lowest tip at the end |
|---|---|---|---|---|
| v6.0.5 release jar | 8 of 9 | 21 (0–6 per run), on a miner and on both followers | 24.1 s | 9 of 9 |
| v6.0.5 + this PR's guard | 0 of 9 | 0 (no divergent sample) | 14.5 s | 9 of 9 |

Each run is a single 150 s observation on one topology. The counts are descriptive, not a statistical test.

On the release jar, divergent episodes lasted from under a second to 11 s (median 4 s). All 21 match node-log lines in which a block whose header had been logged as orphaned became the best full block. On the guarded nodes such a block became the best full block three times, each time a few milliseconds after a new best header was logged. The guard only allows that promotion once the block is on the best-header chain, which is the "picked up if the header chain later switches to their fork" path in the PR description. With the guard, the longest lag of the full chain behind the header chain (14.5 s) was within the release jar's range (up to 24.1 s), and no stall was seen in these runs.

**Build.** Only the PR's production change (the `isInBestChain(newBestBlockHeader.id)` condition in `processBetterChain`) was applied, as one commit on the v6.0.5 tag, and built with `sbt assembly` on JDK 8.
- `/info.appVersion` reported `6.0.5-1-103c0c03-SNAPSHOT` on all four nodes in every guarded run, and `6.0.5` on all four in every release run. The nodes ran on OpenJDK 21.
- This tests the guard on v6.0.5, not the PR branch as submitted, and the PR's test changes were not built.
- `FullBlockProcessor.scala` and `VerifyADHistorySpecification.scala` are unchanged from v6.0.3-RC1 through v6.0.7, so the PR's diff should apply as-is to the current release branches.
- At the time of writing, the guard is not on master, the v6.0.6 or v6.0.7 branches, or the v6.1.5 or v6.5.0-RC3 tags.
- The failing "Run node tests" job on this PR also fails on other PRs based on v6.0.3-RC1, including ones that change only documentation, so it may come from the base branch rather than from this change.

**Reproduce.** The script below is self-contained. It needs:
- Linux with unprivileged user namespaces
- util-linux (`unshare`, `mount`)
- iproute2 (`ip`, `tc`) with the `sch_netem` module
- `jq`, `curl`, `unzip`, `awk`, `pkill`, and bash 4
- a Java runtime and four JVMs' worth of memory at `-Xmx512m` (a similar four-node script peaked at about 1.4 GB)

No Docker is required, and no root either, except `sudo modprobe sch_netem` once if netem isn't loaded. A run takes about 3–5 minutes. Use a fresh `WORKDIR` (the default) for each run. If the script is killed hard, `pkill -f <workdir>/conf_` stops the nodes and `pkill -f ergo-sibling-fork.sh` stops its poller.

```
bash ergo-sibling-fork.sh <ergo-node.jar>   # prints D (samples), E (episodes) and G (longest stretch with fullHeight < headersHeight) per node
```

All 18 runs used this script. The first 12 used a revision that differed only in printing one extra output line that is not used in these results. Builds were alternated run by run.

<details><summary><code>ergo-sibling-fork.sh</code></summary>

```bash
#!/usr/bin/env bash
# ergo-sibling-fork.sh: do a node's best full block and best header ever sit on different forks at the same
# height (ergoplatform/ergo issue #525)?
#
#   bash ergo-sibling-fork.sh <ergo-node.jar>
#
# Four real node processes (--devnet, private magic), each in its own network namespace, with one veth pair
# per link. No root is needed: the script runs itself under `unshare -Urmn`. The one exception: if the
# sch_netem kernel module is not loaded and cannot be auto-loaded, run `sudo modprobe sch_netem` once.
#
#   A (miner) ---D_AC--- C (miner)      Miners poll at P (A) and P_C (C); unequal rates let forks resolve. The
#     |  \              /  |            latency between them makes equal-height sibling blocks common. Followers L and S are
#     |   \--D_X-- --D_X/  |            close to one miner and far from the other. Nobody is restarted.
#     L (follower)    S (follower)
#
# For W seconds, /info is polled on all four nodes as fast as they answer. Reported per node:
#   D = samples with fullHeight == headersHeight but bestFullHeaderId != bestHeaderId (full block and header
#       on different forks at the same height); E = distinct such episodes (independent of polling speed);
#   G = the longest continuous stretch with fullHeight < headersHeight (the full chain lagging the header chain).
# At the end: whether all four nodes agree on the header id at min(fullHeight) - 5.
#
# Requires: Linux with unprivileged user namespaces, util-linux (unshare, mount), iproute2 (ip, tc) with
#           sch_netem, jq, curl, unzip, awk, bash 4+, and a Java runtime the jar supports on PATH.
# Env: D_AC (ms, default 500), D_X (ms, default D_AC/2), P (miner A polling, default 4s), P_C (miner C, default 5s),
#      W (seconds, 150),
#      WORKDIR (default: mktemp -d). Use a fresh WORKDIR for each run.
# Runtime: about W + 2 minutes. RAM: 4 JVMs at -Xmx512m.
set -uo pipefail

if [[ "${FC_INNER:-0}" != 1 ]]; then
  JAR="$(readlink -f "${1:?usage: $0 <ergo-node.jar>}")"
  [[ -f "$JAR" ]] || { echo "no jar: $JAR" >&2; exit 2; }
  for c in unshare ip tc jq curl unzip awk java; do command -v "$c" >/dev/null || { echo "missing: $c" >&2; exit 2; }; done
  WORKDIR="${WORKDIR:-$(mktemp -d)}"; mkdir -p "$WORKDIR"; WORKDIR="$(readlink -f "$WORKDIR")"
  [[ -e "$WORKDIR/node_A.log" ]] && { echo "WORKDIR $WORKDIR was already used; pass a fresh one" >&2; exit 2; }
  unshare -Urmn true 2>/dev/null || { echo "unprivileged user namespaces are unavailable here (on Ubuntu 23.10+:" \
    "sudo sysctl kernel.apparmor_restrict_unprivileged_userns=0, or use a VM)" >&2; exit 2; }
  export FC_INNER=1 JAR WORKDIR
  exec unshare -Urmn bash "$(readlink -f "${BASH_SOURCE[0]}")"
fi

# ---------------- inside the namespaces ----------------
D_AC=${D_AC:-500}; D_X=${D_X:-$((D_AC/2))}; P=${P:-4s}; P_C=${P_C:-5s}; W=${W:-150}
MAGIC='[99,97,102,101]'
# Ergo's own public test mnemonic (src/main/resources/nodeTestnet/application.conf); gives miners a reward address.
MNEMONIC="ozone drill grab fiber curtain grace pudding thank cruise elder eight picnic"
API_HASH=324dcf027dd4a30a932c441f365a25e86b173defa4b8e58948253471b81b72cf
echo "[sf] jar=$JAR sha256=$(sha256sum "$JAR" | cut -c1-16) D_AC=${D_AC}ms D_X=${D_X}ms P=$P P_C=$P_C W=${W}s workdir=$WORKDIR"
echo "[sf] java: $(java -version 2>&1 | head -1)"

mount --make-rprivate / 2>/dev/null || true
mount -t tmpfs tmpfs /run || { echo "FAIL: cannot mount tmpfs on /run in the user namespace"; exit 10; }
mkdir -p /run/netns
# The node resolves some fallback configs relative to its CWD; give it the jar's own copies.
RT="$WORKDIR/rt"; mkdir -p "$RT/src/main/resources"
unzip -o -q "$JAR" application.conf devnet.conf mainnet.conf testnet.conf -d "$RT/src/main/resources"

NODES=(A C L S)
declare -A LIP PRIMARY PID
for n in "${NODES[@]}"; do ip netns add "ns_$n"; ip -n "ns_$n" link set lo up; done
LINKS=("A C" "A L" "A S" "C L" "C S")          # link i -> subnet 10.9.(20+i).0/30; first node .1, second .2
for i in "${!LINKS[@]}"; do
  read -r a b <<< "${LINKS[$i]}"; s="10.9.$((20+i))"
  ip link add "ve_${a}_${b}" type veth peer name "ve_${b}_${a}"
  ip link set "ve_${a}_${b}" netns "ns_$a"; ip link set "ve_${b}_${a}" netns "ns_$b"
  ip -n "ns_$a" addr add "$s.1/30" dev "ve_${a}_${b}"; ip -n "ns_$b" addr add "$s.2/30" dev "ve_${b}_${a}"
  ip -n "ns_$a" link set "ve_${a}_${b}" up; ip -n "ns_$b" link set "ve_${b}_${a}" up
  LIP["$a,$b"]="$s.1"; LIP["$b,$a"]="$s.2"
  [[ -z "${PRIMARY[$a]:-}" ]] && PRIMARY[$a]="$s.1"; [[ -z "${PRIMARY[$b]:-}" ]] && PRIMARY[$b]="$s.2"
done
netem(){ ip netns exec "ns_$1" tc qdisc replace dev "ve_${1}_${2}" root netem $3; }   # $1->$2 direction
delay(){ netem "$1" "$2" "delay ${3}ms" && netem "$2" "$1" "delay ${3}ms"; }
{ delay A C "$D_AC" && delay A S "$D_X" && delay C L "$D_X"; } || { echo "FAIL: tc netem unavailable (try: sudo modprobe sch_netem)"; exit 10; }

declare -A MINING=([A]=true [C]=true [L]=false [S]=false) POLL=([A]=$P [C]=$P_C)
declare -A PEERS=([A]="C L S" [C]="A L S" [L]="A C" [S]="A C")
declare -A EXTRA
conf(){ local n=$1; local f="$WORKDIR/conf_$n.conf" kp="" p
  for p in ${PEERS[$n]}; do kp+="\"${LIP[$p,$n]}:9021\","; done
  { echo "ergo.directory=\"$WORKDIR/data_$n\""
    echo "ergo.node.mining=${MINING[$n]}"; echo "ergo.node.offlineGeneration=${MINING[$n]}"
    echo "ergo.node.useExternalMiner=false"
    [[ ${MINING[$n]} == true ]] && echo "ergo.node.internalMinerPollingInterval=${POLL[$n]}"
    echo "ergo.wallet.testMnemonic=\"$MNEMONIC\""; echo "ergo.wallet.testKeysQty=5"
    echo "scorex.network.bindAddress=\"0.0.0.0:9021\""
    echo "scorex.network.declaredAddress=\"${PRIMARY[$n]}:9021\""
    echo "scorex.network.knownPeers=[${kp%,}]"
    echo "scorex.network.allowLocal=true"; echo "scorex.network.nodeName=\"$n\""
    echo "scorex.network.magicBytes=$MAGIC"
    echo "scorex.restApi.bindAddress=\"127.0.0.1:9052\""; echo "scorex.restApi.apiKeyHash=\"$API_HASH\""
    # settings from Ergo's src/it/resources/devnetTemplate.conf, so a sole-peer follower fully syncs
    echo "ergo.node.stateType=utxo"; echo "ergo.node.verifyTransactions=true"; echo "ergo.node.blocksToKeep=-1"
    echo "ergo.node.PoPoWBootstrap=false"; echo "ergo.node.minimalSuffix=10"; echo "ergo.node.mempoolCapacity=10000"
    echo "ergo.chain.powScheme.powType=autolykos"; echo "ergo.chain.powScheme.k=32"; echo "ergo.chain.powScheme.n=26"
    [[ -n "${EXTRA[$n]:-}" ]] && echo "${EXTRA[$n]}"
  } > "$f"; echo "$f"; }
# Each node gets its own java.io.tmpdir so that four nodes starting at once on one host do not interfere at startup.
launch(){ local n=$1 c; c=$(conf "$n"); mkdir -p "$WORKDIR/data_$n" "$WORKDIR/tmp_$n"
  ip netns exec "ns_$n" bash -c "cd '$RT' && exec java -Xmx512m -Djava.io.tmpdir='$WORKDIR/tmp_$n' -jar '$JAR' --devnet -c '$c'" >> "$WORKDIR/node_$n.log" 2>&1 &
  PID[$n]=$!; }
rest(){ ip netns exec "ns_$1" curl -s --max-time 4 "http://127.0.0.1:9052$2"; }
f(){ local v; v=$(rest "$1" /info | jq -r ".$2 // \"null\"" 2>/dev/null); echo "${v:-null}"; }   # "null" if unreachable
header_at(){ rest "$1" "/blocks/at/$2" | jq -r '.[0] // empty'; }
wait_up(){ local end=$((SECONDS+90)); while [[ $SECONDS -lt $end ]]; do rest "$1" /info | jq -e .appVersion >/dev/null 2>&1 && return 0; sleep 2; done; return 1; }
POLLER=""
cleanup(){ [[ -n "$POLLER" ]] && kill "$POLLER" 2>/dev/null
           for n in "${NODES[@]}"; do [[ -n "${PID[$n]:-}" ]] && kill "${PID[$n]}" 2>/dev/null; done
           pkill -f "$WORKDIR/conf_" 2>/dev/null; wait 2>/dev/null; }
trap cleanup EXIT
die(){ echo "[sf] INCONCLUSIVE: $*"; exit 3; }

for n in A L S; do launch $n; done
for n in A L S; do wait_up $n || die "$n did not start"; done
end=$((SECONDS+120)); until [[ -n "$(header_at A 3)" ]] || [[ $SECONDS -ge $end ]]; do sleep 2; done
G=$(header_at A 1); [[ -n "$G" && -n "$(header_at A 3)" ]] || die "A did not mine 3 blocks"
EXTRA[C]="ergo.chain.genesisId=\"$G\""; launch C; wait_up C || die "C did not start"
end=$((SECONDS+120)); until [[ "$(header_at C 1)" == "$G" ]] || [[ $SECONDS -ge $end ]]; do sleep 2; done
[[ "$(header_at C 1)" == "$G" ]] || die "C did not sync A's genesis"
echo "[sf] versions: A=$(f A appVersion) C=$(f C appVersion) L=$(f L appVersion) S=$(f S appVersion); genesis=${G:0:16}"

# observe: one line per node per round: epoch_ms node fullHeight headersHeight bestFullHeaderId bestHeaderId
S0=$(date +%s%3N)
( while :; do for n in "${NODES[@]}"; do
    rest "$n" /info | jq -r --arg n "$n" --arg t "$(date +%s%3N)" \
      '"\($t) \($n) \(.fullHeight // "null") \(.headersHeight // "null") \(.bestFullHeaderId // "null") \(.bestHeaderId // "null")"' 2>/dev/null
  done; done ) > "$WORKDIR/samples.txt" &
POLLER=$!
sleep "$W"
kill "$POLLER" 2>/dev/null; wait "$POLLER" 2>/dev/null; POLLER=""

# agreement at min(fullHeight) - 5
minf=""; for n in "${NODES[@]}"; do h=$(f $n fullHeight); [[ "$h" =~ ^[0-9]+$ ]] || die "$n not answering at the end"
  [[ -z "$minf" || $h -lt $minf ]] && minf=$h; done
ah=$((minf-5)); ids=""; for n in "${NODES[@]}"; do ids+="$(header_at $n $ah) "; done
agree=DIFF; [[ $(echo $ids | tr ' ' '\n' | sort -u | grep -c .) == 1 ]] && agree=SAME
maxh=$(awk '$3 ~ /^[0-9]+$/ && $3>m {m=$3} END {print m+0}' "$WORKDIR/samples.txt")

echo "[sf] samples=$(wc -l < "$WORKDIR/samples.txt") over ${W}s; max fullHeight=$maxh; agreement at h=$ah: $agree"
awk -v s0="$S0" '
  $3 ~ /^[0-9]+$/ && $4 ~ /^[0-9]+$/ {
    n=$2; seen[n]++
    if ($3 == $4 && $5 != $6) { D[n]++; k=n SUBSEP $5 SUBSEP $6; if (!(k in ep)) { ep[k]=1; E[n]++ } }
    if ($3 < $4) { if (!(n in st)) st[n]=$1; len=$1-st[n]; if (len>G[n]) G[n]=len } else delete st[n]
  }
  END { for (n in seen) printf "[sf] node %s: samples=%d D=%d E=%d G=%.1fs\n", n, seen[n], D[n]+0, E[n]+0, (G[n]+0)/1000 }' "$WORKDIR/samples.txt" | sort
Dtot=$(awk '$3 ~ /^[0-9]+$/ && $3==$4 && $5!=$6 {d++} END {print d+0}' "$WORKDIR/samples.txt")
Etot=$(awk '$3 ~ /^[0-9]+$/ && $3==$4 && $5!=$6 {k=$2 SUBSEP $5 SUBSEP $6; if (!(k in e)) {e[k]=1; c++}} END {print c+0}' "$WORKDIR/samples.txt")
[[ $maxh -ge 128 ]] && echo "[sf] NOTE: heights reached $maxh (>=128: devnet difficulty step; treat as inconclusive)"
echo "RESULT D_total=$Dtot E_total=$Etot agreement=$agree max_height=$maxh"
echo "[sf] samples: $WORKDIR/samples.txt  node logs: $WORKDIR/node_{A,C,L,S}.log"
```
</details>

**Not covered.**
- equal mining rates, other latencies, more than two miners
- mainnet heights (these chains stayed below 50 blocks)

The run can be repeated against the branch if it is retargeted.

Issue #1380, opened 2021-06-01
[Soft-Fork] Consider maxTarget, not block target in interlink vector calculation


kushti:
Currently, max level for a block in interlink vector is calculated as level = log2(requiredTarget) - log2(realTarget), where required target is calculated from block difficulty (nbits). 

From one of the protocol devs:

"...it seems to be OK, under one assumption:
The difficulty is, generally, non-decreasing"

also,

"If you want to have strong resilience to significantly decreasing difficulty, it is better to use genesis targets in the interlink too."

So to calculate  level as log2(initTarget) - log2(realTarget), where initTarget is target at the moment of network launch (corresponding to ~10 Gh/s).  

The change can be done via soft-fork, where are some of interlink validation rules (401,402,413) should be switched off. Then new rules can be introduced in the same fork, or we can move interlink calculation outside of consensus (so it will be velvet-forked then).

jozanek:
Worth folding one more thing into this fork, since it opens the same formula.

`maxLevelOf` computes `(log2(T) − log2(hit)).toInt` in floating point, and per #2556 the result is not bit-reproducible: `math.log` is specified to within 1 ulp, so a hit within a few hundred ulps of a level boundary truncates differently depending on the JDK and the CPU architecture (`11 30 25` on x86_64 JDK 8 vs `11 30 24` on aarch64, same inputs). Rule 402 is fatal, so if it ever fired it would be a permanent partition along runtime lines, recurring on every resync — not a transient reorg. And the runtime isn't pinned for operators today: CI builds on JDK 8, the Docker image ships JRE 11 for two architectures, and the release binaries bundle JDK 21.

Nothing has come close (nearest header is ~1.7e9 ulps from a boundary) and it isn't grindable, so it doesn't justify a fork on its own. But this fork already disables 401/402/413 and rewrites the level calculation, so switching to an integer form — largest `k` with `hit·2^k ≤ T` on BigInts — costs nothing extra here and removes the class of problem entirely. #2556's scan shows that integer form is identical to every float class on all 1,876,045 headers to date.

It composes with the `initTarget` change rather than competing with it: same formula, exact instead of float.

PR #2332, opened 2026-05-22
make NipopowAlgos.log2 total, add level-boundary canary

### LINKED ISSUE #1039: Incorrect calculation of maxLevelOf because of lBigInt->Double conversion
Rounding caused by conversion of 256-bit value to double in PoPowAlgos.maxLevelOf function cause incorrect calculation of  max level (μ) of the given header.

jozanek:
Refs #1039, #2556. Does **not** close #1039.

Moves the `Double` conversion inside a total `log2(BigInt)`: negative → `NaN`, zero → `-Infinity`, oversized values shifted down first.

**No behavioural change** — the overflow branch is unreachable, both operands being bounded by the group order (256 bits) against a 977-bit threshold. µ and interlinks are unchanged for every header. The value is that `log2` is now total and testable.

The original description was wrong twice: #1039 is about rounding, not overflow, and "cannot flip µ in practice" is false — see #2556. `log2` still bottoms out in `math.log`, so that stands. Fixing it needs `StrictMath.log` or an integer form, both consensus changes → #1380.

Second commit adds the boundary test #2556 says is missing. It asserts the triple is one of three recorded classes rather than pinning a value, which would fail on arm64 while passing CI's x86_64 JDK 8.

Rebased onto master; was on `v6.0.3-RC1`, 207 commits behind.

### DIFF
--- a/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowAlgos.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowAlgos.scala
@@ -68,8 +68,8 @@ class NipopowAlgos(val chainSettings: ChainSettings) {
   def maxLevelOf(header: Header): Int =
     if (!header.isGenesis) {
       val requiredTarget = org.ergoplatform.mining.q / DifficultySerializer.decodeCompactBits(header.nBits)
-      val realTarget = powScheme.powHit(header).doubleValue
-      val level = log2(requiredTarget.doubleValue) - log2(realTarget.doubleValue)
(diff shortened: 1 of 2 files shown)


jozanek:
Reworked after #2556.

Two claims in the old description were wrong: this doesn't close #1039 (rounding, not overflow), and "cannot flip µ in practice" is false. The overflow branch is also unreachable — operands bounded at 256 bits, threshold 977 — so behaviour is unchanged. The payload is a total, testable `log2` plus the boundary canary.

Rebased and retargeted; was on `v6.0.3-RC1`, 207 behind, hence the red CI. Green locally (805/805).

PR #2336, opened 2026-05-26
Stabilize P2P test coverage and IT harness

a-shannon:
Follow-up patch for #2306.

## Summary
- Count pending inbound confirmations against the incoming connection limit.
- Close duplicate/stale inbound confirmations and denials deterministically.
- Keep local/self peers out of the persisted peer database when local peer discovery is disabled.
- Add message payload length boundary coverage for negative, max-size, and over-limit frames.
- Make the Docker IT harness more stable around host REST fallback, temp directory fallback, and bounded log capture.
- Make the fork-resolution and deep-rollback specs verify actual convergence instead of relying only on height progress.
- Fix the off-chain registry test so generated data cannot select a reserved scan id.

## Checks
- `sbt -Denv=test ++2.12.20 "testOnly scorex.core.network.NetworkControllerSpecification org.ergoplatform.network.peer.LocalPeerFilteringSpecification org.ergoplatform.network.MessageSerializerSpecification org.ergoplatform.nodeView.wallet.persistence.OffChainRegistrySpec"`
- `sbt -Denv=test ++2.12.20 "it:testOnly org.ergoplatform.it.ForkResolutionSpec org.ergoplatform.it.DeepRollBackSpec"`

### DIFF
--- a/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
+++ b/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
@@ -1,12 +1,14 @@
 package org.ergoplatform.it
 
 import java.io.File
+import java.util.concurrent.TimeoutException
 import com.typesafe.config.Config
+import org.ergoplatform.it.api.NodeApi.NodeInfo
 import org.ergoplatform.it.container.{IntegrationSuite, Node}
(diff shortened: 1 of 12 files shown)


kushti:
incorporated into 6.0.3 canidate via #2306

PR #2344, opened 2026-05-30
Speed up integration tests and fix Docker Desktop compatibility

jozanek:
# Speed up the integration test suite (and make it runnable on Docker Desktop)

## Summary

The `it:` integration suite was slow. While speeding it up I found the suite also couldn't **build or run** on Docker Desktop on this branch, so this PR has two parts: the speedups, and the fixes that make the suite runnable at all.

## Speedups (no change to what any test verifies)

- **devnet mining poll `5s → 1s`** (`devnetTemplate.conf`). This is the per-block gate on devnet: the mining thread polls the candidate generator every `internalMinerPollingInterval` and idles between polls, so block rate ≈ one block per interval (`blockInterval` only matters at epoch boundaries these short chains never reach). Speeds every spec on the default.
- **`StateRecoveryDigestNodeSpec`**: `10s → 500ms` mining poll (single miner).
- **`ForkResolutionSpec`**: dropped the redundant `Thread.sleep(nodeConfigs.size * 2000)` (~8s × 3 restarts) — the following `eventually { waitForStartup }` already gates readiness.
- **`PrunedDigestNodeSync2Spec`**: tightened a coarse `20s` height poll to `1s`.
- **`OpenApiSpec`**: node made `lazy` so the permanently-`ignore`d test no longer starts/tears down a container.
- **`build.sbt`**: each IT suite forks into its own JVM with bounded parallel execution. Gated via `Tags.ForkedTestGroup` so it can only ever *reduce* concurrency — safe against host-RAM exhaustion. (With sbt's default forked-group limit it stays sequential until the limit is raised, so this is scaffolding for parallel runs on higher-RAM CI.)

## Fixes required to run the suite

- **Docker base image**: `openjdk:11-jre-slim` was removed from Docker Hub → `eclipse-temurin:11-jre-jammy`. The IT image could not build without this.
- **Node REST addressing** (`Node.scala`): the harness targeted the container's default-bridge IP, which is empty on a custom Docker network (Docker Desktop), so every API poll hung. Now uses `localhost` + the host-mapped REST port, mirroring how the network port is already handled.

## it2 mainnet specs

- Persist each spec's synced chain in a stable host directory (separate per state type) so reruns resume instead of syncing from genesis. Delete the directory to force a clean sync.

## Left slow on purpose

- `PrunedDigestNodeSyncSpec` keeps its 10s mining cadence: its assertion that an old block is *already* pruned immediately after sync depends on that slow cadence — faster mining makes the one-shot check race ahead of asynchronous pruning. Left it slow rather than weaken the assertion.


a-shannon:
I had a look because this is currently the blocker behind the IT failures on the hardening PRs. The Docker base image change (openjdk:11-jre-slim -> eclipse-temurin:11-jre-jammy) matches the failure in current PR CI, and the REST address change to host-mapped localhost is consistent with Docker Desktop networking. CI is green.\n\nOne non-blocking cleanup before merge: uild.sbt mentions CLAUDE.md, but I do not see that file in the repo. I would drop that parenthetical so the comment stays repo-local. Otherwise this looks good to me.

kushti (approved):

kushti:
@jozanek please resolve conflicts

PR #2352, opened 2026-06-03
Avoid lazy mempool streams during candidate assembly

### LINKED ISSUE #1551: Transactions returned from memory pool are Stream collection
`TreeMap.values.toSeq` resolves to `Stream` which is later used in `CandidateGenerator` and `CleanupWorker` ... `CandidateGenerator` is very inefficient ... multiple `++` of `List` of transactions ... Generating a block from transactions could be much more efficient

BenItBuhner:
## Summary
- materialize ordered mempool values with `valuesIterator.toVector` instead of `values.toSeq`
- build candidate transaction input with `ArrayBuilder` before collection
- replace repeated `headOption`/`tail`/`Seq` growth in `collectTxs` with iterator + `Vector` accumulators

Fixes #1551

## Validation
- `git diff --check`
- `/home/bennett/projects/codex-poly-bot-bounty-work/cs launch sbt -- 'Compile/compile'`\n- `/home/bennett/projects/codex-poly-bot-bounty-work/cs launch sbt -- 'set Test / scalacOptions -= "-Xfatal-warnings"' 'testOnly org.ergoplatform.nodeView.mempool.ErgoMemPoolSpec org.ergoplatform.local.MempoolAuditorSpec'` (21 tests passed)\n\nNote: the selected test run relaxes `Test / scalacOptions` only because local OpenJDK 21 fatal-warnings trips unrelated URL deprecation warnings in `ErgoSettingsSpecification` during broader test compilation. `Compile/compile` passed with the repository normal fatal-warning settings.\n\nThis PR is based on current upstream `master` at `2cdbb8c` (`v6.0.2`). I noticed #1551 has an old assigned PR (#1557) that is currently conflicting/stale; this is a small current-master alternative focused on the mempool/candidate-generation lazy collection path.

### DIFF
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -36,6 +36,7 @@ import sigma.validation.ReplacedRule
 import sigma.{Coll, Colls}
 
 import scala.annotation.tailrec
+import scala.collection.mutable
 import scala.concurrent.duration._
 import scala.util.{Failure, Random, Success, Try}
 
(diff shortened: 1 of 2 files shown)


jozanek:
@BenItBuhner wouldn't it make more sense to update to scala 2.13 which behaves correctly in this case?

BenItBuhner:
Thanks, I checked that path. I agree Scala 2.13 fixes the specific `TreeMap.values.toSeq` behavior, but in the current build the root `ergo` project is still on Scala 2.12.20 and is not included in the 2.13 cross-build.

For example, `++ 2.13.16 -v` switches only `ergoWallet`, `ergoCore`, and `avldb`, and explicitly excludes the root `ergo` project where `CandidateGenerator` lives. A direct `++ 2.13.16` root compile attempt did not get to a useful compile either; it still tried resolving local snapshot root dependencies as `_2.12` artifacts.

So I kept this PR as a narrow current-master fix for the default/root Scala 2.12 path: materialize mempool values explicitly and avoid lazy/linear collection behavior in candidate assembly. If you would prefer to solve this by making the root project compile/run on 2.13 instead, I can close this as superseded, but that looked like a broader build/release migration than #1551 itself.

kushti (approved):

jozanek (approved):

PR #2354, opened 2026-06-05
Optimize mempool transaction collection

### LINKED ISSUE #1551: Transactions returned from memory pool are Stream collection

Peter7896:
## Summary

- Materialize ordered mempool transaction snapshots with `valuesIterator.toVector` instead of `TreeMap.values.toSeq`, avoiding lazy `Stream` results on Scala 2.12.
- Build the candidate transaction list with a `Vector` builder before block assembly.
- Use `Vector` accumulators in `collectTxs` to avoid repeated generic `Seq` appends while preserving transaction order.
- Extend the mempool ordering test to assert the public prioritized transaction snapshots are strict `Vector`s.

Closes #1551

## Validation

- `git diff --check`
- `sbt "testOnly org.ergoplatform.nodeView.mempool.ErgoMemPoolSpec"` - 19 tests passed
- `sbt "testOnly org.ergoplatform.mining.CandidateGeneratorPropSpec"` - 8 tests passed

### DIFF
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -589,13 +589,21 @@ object CandidateGenerator extends ScorexLogging {
         500000
       }
 
+      val transactionsToCollect = {
+        val builder = Vector.newBuilder[ErgoTransaction]
+        builder ++= emissionTxs
+        builder ++= prioritizedTransactions
+        builder ++= poolTxs.iterator.map(_.transaction)
+        builder.result()
+      }
+
       val (txs, toEliminate) = collectTxs(
         minerPk,
         state.stateContext.currentParameters.maxBlockCost - safeGap,
         state.stateContext.currentParameters.maxBlockSize,
         state,
         upcomingContext,
-        emissionTxs ++ prioritizedTransactions ++ poolTxs.map(_.transaction)
+        transactionsToCollect
       )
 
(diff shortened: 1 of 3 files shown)


src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -589,13 +589,21 @@ object CandidateGenerator extends ScorexLogging {
         500000
       }
 
+      val transactionsToCollect = {
+        val builder = Vector.newBuilder[ErgoTransaction]
+        builder ++= emissionTxs
+        builder ++= prioritizedTransactions
+        builder ++= poolTxs.iterator.map(_.transaction)
+        builder.result()
+      }
```
jozanek, line 599: what's complexity of these concats?

Peter7896: Good point. I pushed 10b384e97 to make this explicit: the candidate transaction list is built with a size-hinted `Vector` builder and three `foreach` passes, so this part is O(emissionTxs.size + prioritizedTransactions.size + poolTxs.size) and does not allocate intermediate collections via chained `++`. `CandidateGeneratorPropSpec` passes locally.

PR #2355, opened 2026-06-05
Add node version to block extensions

### LINKED ISSUE #962: Save version of the node in the Block Extensions section

Peter7896:
## Summary
- reserve optional extension key `0x03/0x00` for the node application version
- append `Version.VersionString` to locally generated block candidates while leaving extension validation semantics unchanged
- cover candidate generation so the produced extension contains exactly one node-version field

## Notes
The new field is included in the extension root for blocks produced by upgraded miners, but validators still treat it as a generic extension field: parameter parsing continues to read prefix `0x00`, NiPoPoW interlinks continue to read prefix `0x01`, and validation settings continue to read prefix `0x02`. Existing extension validation only enforces field size, duplicate-key, non-empty, and interlink structure rules for unknown keys.

The candidate-generator test uses an isolated data directory because `ErgoNodeViewRef` initializes local node state during the test.

## Validation
- `git diff --check`
- `sbt "testOnly org.ergoplatform.mining.CandidateGeneratorSpec -- -z node"`

Closes #962

### DIFF
--- a/src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
+++ b/src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
@@ -1,5 +1,8 @@
 package org.ergoplatform.mining
 
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+
 import akka.actor.{ActorRef, ActorSystem}
 import akka.pattern.{StatusReply, ask}
 import akka.testkit.{TestKit, TestProbe}
@@ -7,6 +10,7 @@ import akka.util.Timeout
 import org.bouncycastle.util.BigIntegers
 import org.ergoplatform.mining.CandidateGenerator.{Candidate, GenerateCandidate}
 import org.ergoplatform.modifiers.ErgoFullBlock
+import org.ergoplatform.modifiers.history.extension.Extension
 import org.ergoplatform.modifiers.history.header.Header
 import org.ergoplatform.modifiers.mempool.{ErgoTransaction, UnconfirmedTransaction, UnsignedErgoTransaction}
 import org.ergoplatform.network.ErgoNodeViewSynchronizerMessages.FullBlockApplied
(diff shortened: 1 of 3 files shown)


Peter7896:
：）

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -56,6 +60,35 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
 
   private val defaultSettings60 = defaultSettings.copy(networkType = DevNet60, directory = defaultSettings.directory + "60")
 
+  it should "include node version in generated block extension" in new TestKit(ActorSystem()) {
+    val testProbe = new TestProbe(system)
+    val ergoSettings: ErgoSettings = defaultSettings.copy(
+      directory = Files.createTempDirectory("ergo-node-version").toFile.getAbsolutePath
+    )
```
jozanek, line 68: what's wrong with default settings?

jozanek:
This looks like AI slop, do you understand what is going on and what you're trying to solve?

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
Peter7896: You are right to ask. This test starts `ErgoNodeViewRef`, which initializes local node state; using the shared `defaultSettings` data directory can fail when state from another run is already present. I added a short note beside the isolated directory setup and re-ran `CandidateGeneratorSpec -- -z node` successfully.

Peter7896:
Thanks for the direct feedback. I tightened the PR description because the initial draft did not spell out the protocol boundary clearly enough. The intended scope is: add one optional generic extension field for locally mined blocks, keep parameter / interlinks / validation-settings parsing unchanged, and rely on the existing generic extension validation rules for unknown 2-byte keys. I also pushed c335e6c1f to explain why the candidate-generator test uses an isolated data directory. I will keep this as a draft until the key-space/design direction is acceptable.

PR #2360, opened 2026-06-07
Advance wallet past unavailable blocks in pruned mode

jozanek:
In digest/pruning mode a node can prune block sections before the wallet has scanned them. When the wallet's catch-up loop (`ScanInThePast`) fetched such a block it received `None` and skipped it **without advancing the wallet's scanned height**. Because `expectedNextBlockHeight` then stayed put, the guard never matched again, the message was dropped, and the wallet dead-looped on that height forever — it stopped tracking balances. (Reported in issue 1187.)

This advances the wallet's scanned height past an unavailable block instead of getting stuck:

- `WalletRegistry.skipBlock` moves the digest height forward without touching balances, transacting a new version (keyed by the still-present header id) so rollback keeps working.
- `ErgoWalletService.skipBlockUpdate` exposes it on the service.
- `ErgoWalletActor.ScanInThePast` uses it for the `None` case in pruned mode; the existing recursion then proceeds and the loop self-heals. Non-pruned mode keeps its previous behaviour.

Transactions in a block that was already pruned are unrecoverable on a node that fell more than `blocksToKeep` behind, but the wallet no longer gets stuck. A follow-up PR keeps blocks from being pruned before the wallet scans them, avoiding that loss.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletActor.scala
@@ -248,8 +248,27 @@ class ErgoWalletActor(settings: ErgoSettings,
                 case Success(updatedState) =>
                   updatedState
               }
+            case None if settings.nodeSettings.isFullBlocksPruned =>
+              // The block data is unavailable - it was pruned before the wallet could scan it.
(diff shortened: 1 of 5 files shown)


jozanek:
Superseded by #2361 (recreated from an upstream branch so the prevention PR #2362 can stack on it natively).

PR #2364, opened 2026-06-09
Fix mining candidate cache and PoW boundary consistency

a-shannon:
# Fix mining candidate cache and PoW boundary consistency

## Summary

This patch tightens two mining-side consistency checks:

- Candidate cache reuse now also requires the cached candidate to use the requested effective miner public key.
- `checkNonces` now uses the same strict target inequality as block validation.

## Why

`GenerateCandidate(..., Some(customPk))` can be used by the mining API to request work for a specific miner key. The previous cache check only compared requested transactions, so an empty or matching transaction set could reuse a candidate generated for a different key. In that case, the returned candidate could carry a stale `externalVersion.pk`.

`checkOrderingBlockPoW` and `checkInputBlockPoW` validate hits with strict `< target` semantics. `checkNonces` used `<=`, so it could report a solution at the exact boundary that validation would reject.

## Changes

- Compute `effectiveMinerPk = optPk.getOrElse(minerPk)` before checking the cached candidate.
- Require `cachedFor(...)` to match both transactions and `candidate.externalVersion.pk`.
- Change `checkNonces` ordering and input thresholds from `<=` to `<`.
- Align existing PoW parameter tests with the strict validation ranges.

## Tests

- Strengthened `CandidateGeneratorSpec` so the custom-key candidate must expose the custom public key.
- Added a deterministic `checkNonces` boundary regression: the first accepted target must satisfy `d < target`.
- Updated `AutolykosPowSchemeParametersSpec` assertions to match the validator ranges:
  - ordering block: `d < b`
  - input block: `b <= d < b * subblocksPerBlock`

## Validation

- `git diff --check`
- Targeted sbt tests were attempted, but dependency resolution stopped before test execution because `org.scorexfoundation:sigma-state_2.13:6.0.2-30-59782d92-SNAPSHOT` is not available from the configured resolvers.


a-shannon:
Closing this draft. The actionable mining-cache fix was split into #2373 and merged into v6.0.3; this draft also carried unrelated Matrix base history and should not be reviewed.

PR #2367, opened 2026-06-09
Bind Matrix previous transaction digest in extension fields

a-shannon:
Bind `PreviousInputBlockTransactionsDigestKey` to the accumulated previous input-block transaction digest. The regression requires distinct current/previous digests and checks both candidate extension fields against their corresponding input-block fields.

Current head: `4ffa7369da2a6d63785f99e2aa992d43860f6d43`, targeting `weak-blocks` at `c216c5b6e3310911d5f3822b40c5e57eee6fc45a`. [Review the preceding update](https://github.com/a-shannon/ergo/compare/43f2c72a142ce364c6eb791774d4b598ce75d149...1fbab38f0fccbf68680f371d911ac90016152e01) from the previous head: ten paths reuse the existing mining prerequisites. The original one-argument production correction and complete digest-binding regression remain unchanged; the only fixture conflict combined the required imports.

Review/integrate the shared increments before the owned correction: [#2501](https://github.com/ergoplatform/ergo/pull/2501) owns bootstrap S `13352c1e`, Scala compatibility C `1bf12ba7`, and mining fixtures `e3493d84` → `b43a21c4`; [#2504](https://github.com/ergoplatform/ergo/pull/2504) owns mining selection E `e77f77c2`; [#2505](https://github.com/ergoplatform/ergo/pull/2505) owns solution forwarding F `13d79b04`. These exact prerequisites are reused rather than duplicated. The cumulative target diff includes their files alongside the two owned candidate files.

Normal main and complete test-source compilation passed at the preceding public head. At that preceding head, the full CandidateGeneratorSpec passed **20/20**, including the owned digest case and the 19 existing candidate cases, with no ignored or aborted tests. Independent source review found no blocker. Unchanged S/C validation remains applicable; the target still declares Sigma `6.0.5-22-368a860b-SNAPSHOT` with its pinned bootstrap source.

The previous head's native node failures motivated the existing fixture prerequisites; this selected local result does not establish that every historical CI failure has the same cause. Exact-head native CI remains a separate gate. [Integration tracker](https://github.com/ergoplatform/ergo/issues/2533).

The [new review increment](https://github.com/a-shannon/ergo/compare/1fbab38f0fccbf68680f371d911ac90016152e01...4ffa7369da2a6d63785f99e2aa992d43860f6d43) adds the existing Fork observation/diagnostic fixtures, exact shared [P/M/funding support](https://github.com/a-shannon/ergo/commit/29e6b6c641c6b2f8e179a9a596fc566c494a8a7d), and [G](https://github.com/a-shannon/ergo/commit/d62c945a4d6314c54a7b57f18a94aee8fd258aa7) with its [weak test compatibility follow-up](https://github.com/a-shannon/ergo/commit/ec9c33682f3cd07266ef6c4c4cd1a763f71200d0). [R](https://github.com/a-shannon/ergo/commit/76e5ec4bfc377334fdced4bd60a61df06e6b4cb6), owned by [#2535](https://github.com/ergoplatform/ergo/pull/2535), precedes G, owned by [#2511](https://github.com/ergoplatform/ergo/pull/2511). Review these shared commits once before integrating this consumer. G adds available genesis to sparse summaries, keys their cache by selected tip and full/reduced mode, and schedules downloads through retained common ancestry. No X holder wake-up increment is added here.

The final seven-file G join exactly matches the independently reviewed tree. Normal Scala 2.12 compilation and 109/109 checks passed: weak synchronizer 88, cache 4, stored histories 6 and scheduling 11. The original large-mempool case 1 and pure convergence/diagnostic cases 37 passed in two separate fresh JVMs before G; their source inputs remain unchanged by that join. Prior core5, mempool13 and funding2 XML showed zero failures, but their combined launcher later exited2 during a second node test-task classloader reload. That original exit is retained separately; it is not reported as a wholly successful launcher. The new 109/38 gates exited0 and have fresh suite identities.

The owned previous-digest production correction and complete candidate regression remain unchanged. Independent review confirmed owner preservation and all 88 synchronizer assertions. The historical Fork timeout does not exclusively establish a cause; actual fork/recovery behavior and remaining matrix-wide checks require this head's native CI.


src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -1250,6 +1273,80 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
...
+      case scala.util.Success((candidate: CandidateGenerator.Candidate, _)) =>
+        val candidateBlock = candidate.candidateBlock
+        val extensionFields = candidateBlock.extension.fields
+
+        def extensionValue(key: Array[Byte]): Seq[Byte] =
+          extensionFields.find(_._1.sameElements(key)).map(_._2.toSeq).get
+
+        extensionValue(Extension.InputBlockTransactionsDigestKey) shouldBe
+          candidateBlock.inputBlockFields.transactionsDigest.toSeq
+        extensionValue(Extension.PreviousInputBlockTransactionsDigestKey) shouldBe
```
cafebedouin, line 1340: **[Nit] Give the previous-digest assertion an independent expected value.**

Both sides of this equality come from the one variable the producer computed: the extension slot and `inputBlockFields.prevTransactionsDigest` are each `previousInputBlocksTransactionsDigest` (`CandidateGenerator.scala:756`, `:762`). The spec therefore catches "the current digest written twice", which is this PR's bug. It would still pass if that variable were computed over the wrong transactions.

The fixture already holds the one transaction the previous digest should cover.

Recommended: also assert the slot against a root recomputed in the spec. Not compiled here:

```scala
extensionValue(Extension.PreviousInputBlockTransactionsDigestKey) shouldBe
  Algos.merkleTreeRoot(Seq(LeafData @@ idToBytes(tx.id))).toSeq
```

cafebedouin (commented):
Review carried out by Claude (Anthropic, Claude Fable 5.1) for cafebedouin, with independent passes by Grok (grok-4.7-build) and Gemini (3.8 Flash). Taken from the #2533 review table — Protocol commitments, the #2367 unit.

**Commit:** `3c968622d` (one production line in `CandidateGenerator`, one new spec), reconciled with the head `4ffa7369d`.
**Scope:** the digest arguments at `CandidateGenerator.scala:751-762` and the new `CandidateGeneratorSpec` case. The rest of this PR's history was not reviewed.
**Answer to the unit's question:** the producer now writes the same two values into the authenticated leaves and into the announcement. This PR does not add a receiver-side comparison; `valid` at this head never compares declared digests with leaves. That comparison is #2368's.

- The extension leaves (`:756`) and the announcement's `InputBlockFields` (`:762`) are built from the same two values in the same order. Before the change the leaves carried the current digest in both slots while the announcement carried the previous digest.
- Current digest (`:751`): Merkle root over the ids of the new input-block transactions. Previous digest (`:754`): Merkle root over `previousOrderingBlockTransactionIds`, described at `:753` as the transactions since the ordering block. The history method that fills that list was not read.
- The two digests use different leaf expressions, `tx.serializedId` and `idToBytes(id)`. They are the same bytes: `ErgoTransaction.id` is `bytesToId(serializedId)` (`ErgoTransaction.scala:68-70`).
- The new spec can fail: with the old argument its second equality compares the empty root with the root over the fixture's transaction. By reading; the revert was not executed here.

**Tests run:** no local build. In the head's CI run [34751079292](https://github.com/ergoplatform/ergo/actions/runs/34751079292), the node job's log shows "should bind input block transaction digests to the matching extension fields" passing, within 1101 succeeded / 0 failed.

**Finding — [Integration] Merge this PR before #2368.** No order is stated in either description or the tracker. #2368's head (`81f98b910`) still has the old line at `CandidateGenerator.scala:756`, while its `:762` declares the previous digest. Its new `fieldsBound` requires every hash from `fieldHashes(inputBlockFields)` to be among the proven leaves, and the proof covers all three input-block keys (`proofForInputBlockData` is `batchProofFor(Extension.InputBlockKeys)`). So at that head `valid` returns false for an input block mined by an unmodified node whenever either transaction set is non-empty. The two sets are disjoint (`filterNot` at `:732`), so the digests coincide only when both are empty. Both heads were read, not run.

Recommended: land #2367 first, then rebase #2368 on `weak-blocks`. To confirm: this PR's new spec, run on #2368's head, should fail at its second equality.

One inline note on the spec, non-blocking.

PR #2380, opened 2026-06-12
Harden blocks API bounds

a-shannon:
Summary:
- Reject `/blocks/chainSlice` requests above the maximum header count by fixing the range-size check.
- Add the same maximum-count guard to bulk `/blocks/headerIds` requests.
- Add route-level regression coverage for both cases.

Tests:
- `sbt "testOnly org.ergoplatform.http.routes.BlocksApiRouteSpec"`


a-shannon:
Closing as superseded: the chainSlice fix landed in #2401, and the remaining /blocks/headerIds bound is covered by the focused #2403, now rebuilt on current master with exact boundary and error-response tests.

PR #2386, opened 2026-06-12
Reject invalid ErgoTree route input

a-shannon:
## Summary
- Catch ErgoTree deserialization failures in the shared REST route parser and reject invalid-but-hex bodies as bad input instead of letting them reach the default 500 handler.
- Apply the same parsing guard to `/utils/ergoTreeToAddress` GET/POST.
- Add route-level regression coverage for valid ErgoTree bodies and invalid hex bodies that are not ErgoTrees.

## Tests
- `sbt "testOnly org.ergoplatform.http.routes.ErgoBaseApiRouteSpec org.ergoplatform.http.routes.UtilsApiRouteSpec"`


a-shannon:
Superseded by the two focused current-master rebuilds: #2399 covers the Utils ErgoTree GET/POST paths, including malformed input and unread suffixes; #2400 covers the shared Blockchain ErgoTree parser. Closing this broader overlapping PR to keep one owner per invariant.

PR #2394, opened 2026-06-12
Reject invalid EC point JSON

a-shannon:
## Summary
- Wrap EC point byte conversion in the JSON decoder so invalid point bytes fail as a normal decoding error.
- Add regression coverage for malformed DHT secret group element bytes.

## Tests
- `sbt "ergoCore/testOnly org.ergoplatform.serialization.JsonSerializationCoreSpec"`
- `sbt "testOnly org.ergoplatform.serialization.JsonSerializationSpec"`


a-shannon:
Consolidated into #2389, which now includes both strict fixed-size byte decoding and exception-safe EC point decoding on current master, with cross-build coverage for Scala 2.11.12, 2.12.20, and 2.13.18.

PR #2379, opened 2026-06-12
Validate NiPoPoW proof parameters

a-shannon:
## Summary
- Reject non-positive or overflowing NiPoPoW proof parameters before proof construction.
- Reuse the validated minimum chain length in both in-memory and database proof builders.
- Keep overflowing API requests as `BadRequest` instead of building a proof with overflowed `m + k`.

## Tests
- `sbt "testOnly org.ergoplatform.http.routes.NipopowApiRoutesSpec"`
- `sbt "testOnly org.ergoplatform.modifiers.history.PoPowAlgosSpec org.ergoplatform.modifiers.history.PoPowAlgosWithDBSpec org.ergoplatform.nodeView.history.PopowProcessorSpecification"`

PR #2396, opened 2026-06-12
Validate NiPoPoW proof header PoW

a-shannon:
## Summary

- Reject NiPoPoW proofs when any header fails Autolykos PoW validation.
- Enforce the check before bootstrap proof headers can be inserted into history.
- Add regression coverage for invalid-PoW proof material.
- Remove a stale `InfoApiRoute` import that prevents the 6.0.5 node and integration-test builds under fatal warnings.

## Relationship to #2379

#2379 validates NiPoPoW `m` and `k` parameter bounds. This PR covers the independent proof-header PoW validation invariant.

## Scope

The NiPoPoW change is intentionally limited to the PoW validation needed for the 6.0.5 candidate. It excludes the `maxLevelOf` arithmetic changes tracked separately in #2332. The build-only import cleanup is kept in a separate commit.

## Tests

- `sbt "testOnly org.ergoplatform.nodeView.history.PopowProcessorSpecification"`
- `sbt "testOnly org.ergoplatform.local.NipopowVerifierSpec org.ergoplatform.http.routes.NipopowApiRoutesSpec org.ergoplatform.modifiers.history.PoPowAlgosSpec org.ergoplatform.modifiers.history.PoPowAlgosWithDBSpec org.ergoplatform.nodeView.history.PopowProcessorSpecification"`
- `sbt -Denv=test clean ++2.12.20 compile`
- Regression verified failing before the fix and passing after it.
- Publication guard on the final commit range.

### DIFF
--- a/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowAlgos.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowAlgos.scala
@@ -75,6 +75,8 @@ class NipopowAlgos(val chainSettings: ChainSettings) {
       Int.MaxValue
     }
 
+  def hasValidPow(header: Header): Boolean = powScheme.validate(header).isSuccess
+
   /**
     * Computes best score of a given chain.
     * The score value depends on number of µ-superblocks in the given chain.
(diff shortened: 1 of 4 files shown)


a-shannon:
@kushti I retargeted this PR to `v6.0.5` and narrowed the NiPoPoW change to the missing Autolykos proof-header validation plus its bootstrap regression test.

#2379 covers the independent `m`/`k` parameter bounds. The previous `maxLevelOf` refactor has been removed from this PR to avoid overlapping #2332.

CI also exposed the stale unused `ContentTypes` import already present on `v6.0.5`; that build fix is kept in a separate one-line commit. All eight CI jobs are now green, and the PR is ready for review for 6.0.5.

PR #2417, opened 2026-07-13
Reject invalid inbound NiPoPoW parameters

a-shannon:
## Summary

- validate received NiPoPoW `m` and `k` through the canonical `PoPowParams` rules before proof admission
- make `bestArg` fail fast for an invalid security parameter
- add binary round-trip and bounded-termination regression coverage

## Why

PR #2379 validates locally constructed and REST-generated proofs, but deserialized proofs are constructed directly. This adds the corresponding consumer-side parameter validation and keeps scoring inside its supported domain.

## Tests

- `NipopowVerifierSpec`
- `PoPowAlgosSpec`
- `PoPowAlgosWithDBSpec`
- `PopowProcessorSpecification`
- `SerializationTests`
- `NipopowApiRoutesSpec`

35 affected tests passed.

## Scope

- no wire-format change
- no change to valid proof selection
- binding a received proof to configured or requested parameters remains separate

### DIFF
--- a/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowProof.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowProof.scala
@@ -72,7 +72,8 @@ case class NipopowProof(popowAlgos: NipopowAlgos,
     * @return true if the proof is valid
     */
   lazy val isValid: Boolean = {
-    this.hasValidConnections &&
+    PoPowParams(m, k, continuous).isSuccess &&
+      this.hasValidConnections &&
       this.hasValidHeights &&
       this.hasValidProofs &&
       this.hasValidDifficultyHeaders &&
--- a/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowProof.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowProof.scala
@@ -72,7 +72,7 @@ case class NipopowProof(popowAlgos: NipopowAlgos,
     * @return true if the proof is valid
     */
   lazy val isValid: Boolean = {
-    PoPowParams(m, k, continuous).isSuccess &&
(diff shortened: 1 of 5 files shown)


ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowProof.scala
```diff
@@ -72,7 +72,8 @@ case class NipopowProof(popowAlgos: NipopowAlgos,
     * @return true if the proof is valid
     */
   lazy val isValid: Boolean = {
-    this.hasValidConnections &&
+    PoPowParams(m, k, continuous).isSuccess &&
```
kushti: better to extract check as a predicate, calling constructor and check result for success is quite dirty

kushti (changes requested):

ergo-core/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowProof.scala
a-shannon: Addressed in 3cc7f3c4c4: extracted PoPowParams.isValid(m, k) and reused it from both PoPowParams.apply and NipopowProof.isValid, so proof validation no longer constructs PoPowParams merely to test validity. Added lower-bound and overflow-boundary coverage. All 35 affected tests pass locally.

a-shannon:
@kushti, the requested predicate extraction is addressed in 3cc7f3c4c4. CI is 7/8 green; the IT job hit two convergence flakes: UtxoStateNodesSyncSpec saw different height-5 tips, and DeepRollBackSpec timed out waiting for convergence at height >= 268. The same DeepRollBackSpec timeout occurred today on the v6.0.5 candidate run 29239361191 (height >= 249), while this PR previous attempt passed the full IT job. The PR changes only NiPoPoW files and these specs run with PoPoWBootstrap=false. I do not have rerun rights (403). Could you please rerun the failed jobs for run 29284853831?

kushti (approved):

PR #2418, opened 2026-07-14
Harden REST defaults and wallet change address updates

a-shannon:
Found while reviewing the v6.0.4 release candidate in #2293.

## Summary

- Bind shipped REST profiles and release-bundle configuration to loopback by default.
- Replace the public default API-key hash with an explicit disabled sentinel, and reject both that sentinel and the legacy public default hash on protected routes.
- Hash installer API keys offline from standard input instead of starting a temporary node. The default flow uses a hidden terminal prompt and restores the original terminal state; the legacy `--api-key` option remains available with a process-list exposure warning.
- Require the wallet to be unlocked and the requested change address to be owned by its active prover before persisting it.
- Ignore a pre-existing persisted unowned change address when resolving change outputs, falling back to the wallet root address.
- Return HTTP 400 for expected change-address validation failures while preserving HTTP 500 for unexpected storage or internal failures.
- Ship fail-closed configuration in generated release bundles and place JVM options before `-jar` in generated launchers.

## Security invariants

1. Protected REST routes remain unavailable while a shipped disabled hash or the legacy public default hash is configured.
2. Exposing the REST API externally requires an explicit operator bind address and API-key configuration.
3. A wallet change address can only be selected from keys controlled by the currently unlocked wallet.
4. Persisted values created before this validation cannot redirect the resolver to an unowned address.
5. The normal installer path does not expose the API key through terminal echo or Java process arguments.

## Operator impact

- The default REST bind changes to loopback: port `9053` for mainnet and `9052` for the other bundled profiles.
- Operators intentionally exposing the API must explicitly configure the external bind and should provide transport protection and firewall controls.
- Configurations still using the legacy public default API key must generate a new key and replace its hash. Protected routes now fail closed with the legacy hash.
- `/wallet/updateChangeAddress` now requires an unlocked wallet and a wallet-owned P2PK address. Invalid requests receive HTTP 400.
- The installer still accepts `--api-key` for compatibility, but the hidden interactive prompt is the recommended path.

## Validation

- Wallet route, service, actor, state, and transaction-resolver suites: 77 tests, 0 failures.
- REST settings, release bundle, hash, and authorization suites: 17 tests, 0 failures.
- `DigestStateSpecification` isolated retry: 7 tests, 0 failures.
- Final fat-JAR assembly and offline API-key helper smoke test.
- Bash and dash installer syntax checks.
- Python release-generator AST and OpenAPI YAML parsing.
- scalafmt 2.3.2 and `git diff --check`.

A repository-wide local run on Windows encountered pre-existing native `leveldbjni` JVM crashes after repeated database lifecycle operations. `CandidateGeneratorSpec` also reports the same forced-cache failure on this branch and on an unmodified worktree at the exact v6.0.4 base. Upstream CI remains the authoritative repository-wide gate.

## Related work

This is complementary to #2408. That PR adds optional HTTPS/TLS transport and deployment guidance; this change addresses fail-closed bind/authentication defaults, installer key handling, and wallet change-address authorization. It does not replace the TLS work.

### DIFF
--- a/ci/ergo.conf
+++ b/ci/ergo.conf
@@ -1,0 +1,7 @@
+# Replace the sentinel with a Blake2b256 API key hash to enable protected routes.
+ergo.node.mining = false
+
+scorex.restApi {
+  bindAddress = "127.0.0.1:9053"
+  apiKeyHash = "0000000000000000000000000000000000000000000000000000000000000000"
+}
--- a/ci/release-binaries.py
+++ b/ci/release-binaries.py
@@ -84,7 +84,7 @@ Write-Host @"
 {SCRIPT_LOGO}
 
 "@ -ForegroundColor Red
-jre/bin/java -jar -Xmx4G {JAR_FILENAME} --mainnet -c ergo.conf 
+jre/bin/java -Xmx4G -jar {JAR_FILENAME} --mainnet -c ergo.conf
     """
     with open(f"{app_dir}/ergo-node.ps1", "w") as f:
        f.write(app_script)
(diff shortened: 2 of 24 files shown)


a-shannon:
Superseded by #2423.

PR #2419, opened 2026-07-14
Fix v6 parameter matching call order

a-shannon:
## Summary

- pass locally calculated parameters as the required subset to
  `matchParameters60`;
- keep the pre-v6 strict parameter rule unchanged;
- cover the real epoch-extension processing path across the v6 activation
  boundary.

## Why

`matchParameters60` is directional: every parameter calculated by the local
node must exist with the same value in the received extension, while the
received extension may contain additional parameters unknown to the local
client.

The production caller supplied those operands in reverse order. As a result, a
v6 node could accept an epoch extension missing a locally required parameter
and reject an extension containing a legitimate new parameter.

The corrected order matches the contract documented in #2239 and exercised by
the helper-level tests in #2240. Neither PR changes the production call site.

## Tests

- `VotingSpecification` on Scala 2.11.12, 2.12.20 and 2.13.16: 11 passed on
  each version.
- `ergoCore/test` on Scala 2.11.12, 2.12.20 and 2.13.16: 105 passed and 2
  pre-existing tests ignored on each version.
- The regression matrix covers:
  - pre-v6 strict rejection of an extra parameter;
  - v6 exact parameters;
  - a v6 received superset;
  - a missing known parameter;
  - the v6-specific subblocks parameter missing;
  - a mismatched known value.

A broader local node run was interrupted by a native `leveldbjni` crash on
Windows/JDK 17 while opening a test database. The interrupted
`DigestStateSpecification` suite then passed 7/7 in isolation. GitHub CI uses
Linux and Java 8.


a-shannon:
Superseded by #2422.

PR #2420, opened 2026-07-14
Remove only the transport peer on rejected handshakes

a-shannon:
## Summary

- remove the observed remote transport endpoint when a handshake is rejected;
- prevent a peer-declared address from selecting a different peer-database
  entry for removal;
- add a wire-level regression test that keeps the existing peer connected while
  closing and removing the rejected transport.

## Rationale

`NetworkController.handleHandshake` uses the address declared in the handshake
to detect duplicate peer identities. That value is supplied by the remote peer.
Reusing it as the operand of `RemovePeer` lets a rejected connection name an
existing peer and remove that peer's database entry.

The controller already has the rejected handler's observed
`connectionId.remoteAddress`. This change uses that transport endpoint only for
destructive cleanup, while leaving duplicate detection, handshake acceptance,
wire format and connection-map behavior unchanged.

## Tests

- `sbt "testOnly scorex.core.network.NetworkControllerSpec"` (12 passed)
- relevant network suites covering the controller, serializers and peer
  filtering (24 passed)

The regression serializes the attacker's handshake and delivers it through the
real `PeerConnectionHandler`. It verifies that the attacker transport is
aborted and removed, the existing peer remains connected, and the closed
transport's connection-map slot is released.

### DIFF
--- a/src/main/scala/scorex/core/network/NetworkController.scala
+++ b/src/main/scala/scorex/core/network/NetworkController.scala
@@ -423,7 +423,7 @@ class NetworkController(ergoSettings: ErgoSettings,
 
       if (shouldDrop) {
         connectedPeer.handlerRef ! CloseConnection
-        peerManagerRef ! RemovePeer(peerAddress)
+        peerManagerRef ! RemovePeer(remoteAddress)
         connections -= connectedPeer.connectionId.remoteAddress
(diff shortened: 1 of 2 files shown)


a-shannon:
Superseded by #2421.

PR #2421, opened 2026-07-14
Remove only the transport peer on rejected handshakes

a-shannon:
Rejecting a handshake must remove its observed transport endpoint from the peer database. An address advertised by the rejected handshake must not select a different peer for removal. The owner correction changes one RemovePeer argument; duplicate detection, acceptance rules, serialization and connection-map cleanup remain unchanged.

Head `59f87747ccf862509f1973225e984bacf500c382` remains targeted at `v6.0.6`, base `8995ff243b140e7fe716bd3204423dd35690c2e4`. Both owner files are preserved. The [two-file handshake review increment](https://github.com/a-shannon/ergo/compare/8995ff243b140e7fe716bd3204423dd35690c2e4...1acc2da974084116c7c935642c115673923374dc) and [September 12 test follow-up](https://github.com/a-shannon/ergo/compare/5868ff722ed3d91802bb2265790e3cbb36d00916...1acc2da974084116c7c935642c115673923374dc) remain the original review scope.

The foreign-network regression sends a serialized handshake with a distinct declared address through the real PeerConnectionHandler, without connecting to that declared address. It checks removal of the transport address and transport abort. Reverting only the cleanup operand makes that assertion fail. Established-peer checks remain invariant guards, and a named bounded silence window checks for additional messages. These fixtures use actor probes and the real handler, without live sockets or a real peer-database instance. For inbound connections the source port is normally ephemeral, so removal will usually find no entry; outbound removal identifies the dialed endpoint. This prevents unrelated advertised-address cleanup, without claiming a standalone eclipse attack or changing discovery policy. Removed peers can be learned again through the normal handshake/gossip path.

The [first prerequisite increment](https://github.com/a-shannon/ergo/compare/1acc2da974084116c7c935642c115673923374dc...bb7889bc6912ee78e9926afe394f145aa07fa49e) reuses exact [D](https://github.com/a-shannon/ergo/commit/71f33c6e45a9394557cf3af1bfb422cbc53d26cf) on shared H, owned by [#2535](https://github.com/ergoplatform/ergo/pull/2535). It requires a settled, fully applied non-mining seed and actual isolation before the existing DeepRollBack convergence test, preserving its final oracle, twenty-minute bound and restored connectivity. The [second increment](https://github.com/a-shannon/ergo/compare/bb7889bc6912ee78e9926afe394f145aa07fa49e...59f87747ccf862509f1973225e984bacf500c382) reuses [G](https://github.com/a-shannon/ergo/commit/d62c945a4d6314c54a7b57f18a94aee8fd258aa7), owned by [#2511](https://github.com/ergoplatform/ergo/pull/2511), with #2535's R fixture prerequisite. G supplies the existing fork-continuation/cache/download correction, including a common genesis anchor in full V2 summaries. Review those exact shared prerequisites with their owners; no X body-cache correction is included.

At previous head `1acc2da974084116c7c935642c115673923374dc`, [CI attempt 2](https://github.com/ergoplatform/ergo/actions/runs/34720534557) passed seven of eight jobs and 21/22 IT cases, with one ignored case. DeepRollBack started from unequal seed heights and later timed out with A257/257 and B69/69, each connected to one peer. D corrects the uncontrolled setup; G addresses an independently reproduced fork-advertisement boundary relevant to that divergence. This does not establish an exclusive cause for the historical timeout.

The actual release composition passed normal compilation and 94/94 focused tests: G consumers 47, NetworkController 14, and observer/isolation contracts 33. All three serialized-handshake cases and the original negative control remain covered; fresh independent source/composition reviews found no blockers. These local gates did not run Docker. New-head native CI remains the behavioral gate.

[#2432](https://github.com/ergoplatform/ergo/pull/2432) owns independent pending-inbound accounting and is not a prerequisite of this correction. The historical [published two-PR composition](https://github.com/a-shannon/ergo/commit/cacb2dcb5d453e06200747c8e1612b3d7715a38b) passed all 18 controller cases; its [increment after #2432](https://github.com/a-shannon/ergo/compare/810843cdf3ff51b0a1ff2eabb6d0b631ac93b43d...cacb2dcb5d453e06200747c8e1612b3d7715a38b) remains integration evidence. Both owner corrections remain available for parallel review. Release order and current CI are tracked in [#2533](https://github.com/ergoplatform/ergo/issues/2533).

### DIFF
--- a/src/main/scala/scorex/core/network/NetworkController.scala
+++ b/src/main/scala/scorex/core/network/NetworkController.scala
@@ -423,7 +423,7 @@ class NetworkController(ergoSettings: ErgoSettings,
 
       if (shouldDrop) {
         connectedPeer.handlerRef ! CloseConnection
-        peerManagerRef ! RemovePeer(peerAddress)
+        peerManagerRef ! RemovePeer(remoteAddress)
         connections -= connectedPeer.connectionId.remoteAddress
       } else {
         val newPeerInfo = peerInfo.copy(lastStoredActivityTime = time())
--- a/src/test/scala/scorex/core/network/NetworkControllerSpec.scala
+++ b/src/test/scala/scorex/core/network/NetworkControllerSpec.scala
@@ -3,6 +3,8 @@ package scorex.core.network
 import akka.actor.ActorRef
 import akka.io.Tcp
 import akka.testkit.{TestActorRef, TestProbe}
+import akka.util.ByteString
+import org.ergoplatform.network.{Handshake, HandshakeSerializer}
 import org.ergoplatform.network.message.MessageConstants.MessageCode
 import org.ergoplatform.network.peer.PeerInfo
 import org.ergoplatform.utils.ErgoCorePropertyTest
@@ -24,6 +26,8 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
     implicit val actorSystem = system
 
     val scorexContext: ScorexContext = ScorexContext(Seq.empty, None, None)
+
+    case class EstablishedConnection(connectionProbe: TestProbe, handlerRef: ActorRef)
 
     def createController(maxConnections: Int): (TestActorRef[NetworkController], TestProbe, TestProbe) = {
       val peerManagerProbe = TestProbe("PeerManager")
@@ -56,6 +60,33 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
       peerManagerProbe: TestProbe,
       remoteAddress: InetSocketAddress
     ): InetSocketAddress = {
+      beginIncomingConnection(controller, peerManagerProbe, remoteAddress)
+      remoteAddress
+    }
+
+    def establishIncomingConnectionWithHandler(
+      controller: TestActorRef[NetworkController],
+      peerManagerProbe: TestProbe,
+      remoteAddress: InetSocketAddress
+    ): EstablishedConnection = {
+      val connectionProbe = beginIncomingConnection(
+        controller,
+        peerManagerProbe,
+        remoteAddress
+      )
+
+      val handlerRef = connectionProbe.expectMsgType[Tcp.Register].handler
+      connectionProbe.expectMsg(Tcp.ResumeReading)
+      connectionProbe.expectMsgType[Tcp.Write]
+
+      EstablishedConnection(connectionProbe, handlerRef)
+    }
+
+    private def beginIncomingConnection(
+      controller: TestActorRef[NetworkController],
+      peerManagerProbe: TestProbe,
+      remoteAddress: InetSocketAddress
+    ): TestProbe = {
       val localAddress = settings.scorexSettings.network.bindAddress
       val connectionProbe = TestProbe("Connection")
 
@@ -66,7 +97,7 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
           controller ! ConnectionConfirmed(ConnectionId(remoteAddress, localAddress, Incoming), handlerRef)
       }
 
-      remoteAddress
+      connectionProbe
     }
 
     def establishOutgoingConnection(
@@ -268,6 +299,49 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
 
       connectionProbe.send(controller, Tcp.Connected(remoteAddress, localAddress))
       connectionProbe.expectMsg(Tcp.Close)
+    }
+  }
+
+  property("rejected handshake should remove only the transport peer") {
+    withFixture { f =>
+      implicit val system = f.system
+      val (controller, peerManagerProbe, _) = f.createController(maxConnections = 30)
+
+      val victimAddress = new InetSocketAddress("192.168.1.1", 9001)
+      val attackerAddress = new InetSocketAddress("192.168.1.2", 9002)
+      f.establishIncomingConnection(controller, peerManagerProbe, victimAddress)
+      val attackerConnection = f.establishIncomingConnectionWithHandler(
+        controller,
+        peerManagerProbe,
+        attackerAddress
+      )
+
+      val attackerHandshake = Handshake(
+        defaultPeerSpec.copy(declaredAddress = Some(victimAddress)),
+        System.currentTimeMillis()
+      )
+      attackerConnection.connectionProbe.send(
+        attackerConnection.handlerRef,
+        Tcp.Received(ByteString(HandshakeSerializer.toBytes(attackerHandshake)))
+      )
+
+      attackerConnection.connectionProbe.expectMsg(Tcp.ResumeReading)
+      peerManagerProbe.expectMsg(RemovePeer(attackerAddress))
+      attackerConnection.connectionProbe.expectMsg(Tcp.Abort)
+      peerManagerProbe.expectNoMessage(200.millis)
+
+      val localAddress = settings.scorexSettings.network.bindAddress
+      val duplicateVictimProbe = TestProbe("DuplicateVictim")
+      duplicateVictimProbe.send(controller, Tcp.Connected(victimAddress, localAddress))
+      duplicateVictimProbe.expectMsg(Tcp.Close)
+
+      val attackerReconnectProbe = TestProbe("AttackerReconnect")
+      attackerReconnectProbe.send(controller, Tcp.Connected(attackerAddress, localAddress))
+      peerManagerProbe.expectMsgPF(1.second) {
+        case ConfirmConnection(connectionId, connectionRef) =>
+          connectionId.remoteAddress shouldBe attackerAddress
+          connectionRef shouldBe attackerReconnectProbe.ref
+      }
     }
   }
 
--- a/src/test/scala/scorex/core/network/NetworkControllerSpec.scala
+++ b/src/test/scala/scorex/core/network/NetworkControllerSpec.scala
@@ -345,6 +345,43 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
     }
   }
 
+  property("rejected handshake should preserve an established peer with a different declared identity") {
+    withFixture { f =>
+      val (controller, peerManagerProbe, _) = f.createController(maxConnections = 30)
+
+      val victimAddress = new InetSocketAddress("192.168.1.1", 9001)
+      val declaredIdentity = new InetSocketAddress("192.168.1.3", 9003)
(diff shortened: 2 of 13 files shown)


src/test/scala/scorex/core/network/NetworkControllerSpec.scala
```diff
@@ -271,6 +302,86 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
...
+      attackerReconnectProbe.send(controller, Tcp.Connected(attackerAddress, localAddress))
+      peerManagerProbe.expectMsgPF(1.second) {
+        case ConfirmConnection(connectionId, connectionRef) =>
+          connectionId.remoteAddress shouldBe attackerAddress
+          connectionRef shouldBe attackerReconnectProbe.ref
+      }
+    }
+  }
+
+  property("rejected handshake should preserve an established peer with a different declared identity") {
```
jozanek, line 353: Both properties here exercise the duplicate-identity clause of `shouldDrop`. Tracing the second
clause: a handshake carrying a mismatched `networkMagic` in `SessionIdPeerFeature` reaches the
same cleanup, and `PeerConnectionHandler` forwards `Handshaked` without validating magic — so
that route needs no prior connection to the victim and no precondition at all. That reads to me
as the more severe of the two cases, and it's the one with no coverage here.

Would you be open to adding a third property — inbound handshake with bad magic plus
`declaredAddress` set to a third, unconnected address, asserting `RemovePeer(attackerAddress)`?

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -423,7 +423,7 @@ class NetworkController(ergoSettings: ErgoSettings,
 
       if (shouldDrop) {
         connectedPeer.handlerRef ! CloseConnection
-        peerManagerRef ! RemovePeer(peerAddress)
+        peerManagerRef ! RemovePeer(remoteAddress)
```
jozanek, line 426: Checking my understanding of this line's reach: peer-database rows are keyed on
`peerSpec.address`, i.e. a declared address with a listening port, while an inbound connection's
`remoteAddress` carries the peer's ephemeral source port. So inbound, this `remove` should match
no key and become a no-op, while outbound it still drops the address we dialed.

If that's right it seems worth a line in the description — it's the strongest case for this exact
shape, since the old argument's only database-effective inbound behaviour was the exploitable
one. No code change needed; I'd just like to know whether you read it the same way.

src/test/scala/scorex/core/network/NetworkControllerSpec.scala
```diff
@@ -271,6 +302,86 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
...
+        attackerConnection.handlerRef,
+        Tcp.Received(ByteString(HandshakeSerializer.toBytes(Handshake(victimSpec, System.currentTimeMillis()))))
+      )
+      attackerConnection.connectionProbe.expectMsg(Tcp.ResumeReading)
+      peerManagerProbe.expectMsg(RemovePeer(attackerAddress))
+      attackerConnection.connectionProbe.expectMsg(Tcp.Abort)
+      peerManagerProbe.expectNoMessage(200.millis)
+
+      val observer = TestProbe("ConnectedPeers")(f.system)
+      observer.send(controller, NetworkController.ReceivableMessages.GetConnectedPeers)
```
jozanek, line 383: On the pre-fix code this property already fails five lines up at
`expectMsg(RemovePeer(attackerAddress))`, and `connections -= connectedPeer.connectionId.remoteAddress`
is untouched by the diff — so `remaining` holds the same value before and after the change.

I read these lines as an invariant guard rather than coverage of the fix, which is a useful thing
to have. Would you add a short comment saying so, so a later reader doesn't count them as
regression coverage?

src/test/scala/scorex/core/network/NetworkControllerSpec.scala
```diff
@@ -271,6 +302,86 @@ class NetworkControllerSpec extends ErgoCorePropertyTest {
...
+      )
+      attackerConnection.connectionProbe.send(
+        attackerConnection.handlerRef,
+        Tcp.Received(ByteString(HandshakeSerializer.toBytes(attackerHandshake)))
+      )
+
+      attackerConnection.connectionProbe.expectMsg(Tcp.ResumeReading)
+      peerManagerProbe.expectMsg(RemovePeer(attackerAddress))
+      attackerConnection.connectionProbe.expectMsg(Tcp.Abort)
+      peerManagerProbe.expectNoMessage(200.millis)
```
jozanek: `docs/review/testing.md` flags wall-clock dependence in specs as a CI-flake source. The handshake
timestamps look inert to me, but the three `expectNoMessage(200.millis)` windows can be exceeded
by a loaded runner, and a flaky network spec costs later contributors a re-run.

Would you be open to either reusing the suite's existing timeout constant, or noting that 200ms
is deliberate here?

jozanek (commented):
I read this as restoring an invariant the rest of the file already keeps: the argument to
`RemovePeer` should be an address the node chose, never one a peer asserted. Before this change
line 426 was the only one of three call sites that let a remote peer name the database row to
delete — `:221` passes `c.remoteAddress` and `:339` passes the address we dialed. If that is the
intent, I think the change is right and the size is proportionate.

Two things I went looking for and did not find, which may save you answering them:

- **No self-purge regression.** I expected the old argument to be load-bearing on the
  self-connection branch. It isn't: `peerAddress` diverges from `remoteAddress` only when a
  declared address is present, and every write path into the peer database filters on `isSelf`
  (`PeerManager.scala:28`, `:53`, `:67`), so that row can't be stored in the first place. In the
  NAT'd case where self *can* be stored, `peerSpec.address` is `None` and both versions emit the
  identical message.
- **The fixture refactor isn't scope drift.** `master` already carries
  `EstablishedConnection`, `establishIncomingConnectionWithHandler` and `beginIncomingConnection`
  byte-for-byte via #2427; `v6.0.6` simply predates it. Back-merging this PR to master costs one
  three-line import hunk, and `NetworkController.scala` auto-merges clean.

On **#2432**, which the description asks the integrator to coordinate with — I test-merged the two
heads, and the interaction looks smaller than that framing suggests. `NetworkController.scala`
auto-merges: this PR's only hunk is at 423-429 and #2432's sit between 63 and 300, so no
production line is shared and there is no behaviour to preserve. The one conflict is in
`NetworkControllerSpec.scala`, four hunks, each mechanical: the imports and both blocks of
`property(...)` additions are keep-both-sides (your two properties and #2432's eight assert
disjoint things), and the fixture divergence resolves by taking #2432's version, which is a strict
superset — `beginPendingIncomingConnection` factored out with `beginIncomingConnection`
delegating, so `establishIncomingConnectionWithHandler` keeps exactly the contract your properties
consume and they need no edits. So either order works and whichever lands second does one
test-file rebase. Would you be willing to say that in both descriptions? "Coordinate the shared
controller" reads as a larger integration risk than what I can find.

What I'd like to resolve before approving is the framing rather than the code. The description
argues the duplicate-identity case, which needs an existing connection to the victim. The second
clause of the same `shouldDrop` looks considerably cheaper to me: a mismatched `networkMagic` in
`SessionIdPeerFeature` plus a forged `declaredAddress`, with `PeerConnectionHandler` forwarding
`Handshaked` without validating magic — no precondition, no prior connection, one inbound
connection per eviction, and the delete reaches LevelDB. Neither new property covers that branch
and the body doesn't mention it, so the strongest reason to take this into a release is the part a
reviewer has to reconstruct. I've left an inline suggesting a third property for it.

On severity, so the claim is on the record rather than implied: an evicted row comes back as a
dial candidate within `getPeersInterval = 2m` via `AddPeerIfEmpty`, because
`RandomPeerExcluding` filters only on exclusion and blacklist. But it returns as
`PeerInfo(peerSpec, 0, None)`, and `SeenPeers` requires `connectionType.isDefined ||
lastHandshake > 0`, so it stays out of our onward gossip until we handshake with it. That reads to
me as an eclipse assist and a gossip-quality degradation rather than a standalone eclipse — real
and worth fixing, not an emergency. Does that match how you'd score it?

Two things I can't settle from here, both for a maintainer rather than for you:

- **Release target.** kushti's own 6.0.6 candidate list, #2473, names #2477, #2493 and #2512 — not
  this PR, which also has no milestone. The line in #2533 about P2P changes being prioritized for
  6.0.6 is your summary, and the nearest maintainer statement I could find (#2464, 7 September)
  names only #2496, says "a next release" rather than 6.0.6, and frames P2P changes as the
  suspected *cause* of bootstrap instability. Could someone confirm whether `v6.0.6` is the
  intended target? For what it's worth the provenance argues the change is safe to take whenever:
  the line is byte-identical inherited Scorex code that upstream Scorex and Sparkz both still
  carry, so there's no prior decision here to overturn.
- **Forward-port.** `master` still has `RemovePeer(peerAddress)` at line 426, and `v6.0.7` exists
  as a diverged branch (7 behind `v6.0.6`, 1 ahead) that also still has it. Targeting a release
  branch matches house convention and the back-merge is structural, so the fix won't be lost — but
  historically the candidate PR takes 2 to 7 months, so master carries the primitive until then.
  A master-side tracking issue might be worth it. Relatedly, `v6.0.6` does not contain the merged
  #2427, which is why #2432 re-adds its two blacklist properties; someone should decide how #2427
  reaches the release branch, since a cherry-pick would duplicate them.

One follow-up worth filing separately, not a request on this PR: `shouldDrop` folds three
conditions with three different correct remediations into one boolean, so the cleanup can't know
which fired. After this fix the bad-magic and self branches apply no effective removal and no
penalty inbound — just `CloseConnection` — so bad-magic handshake churn stays cheap. Making the
verdict a value (`Accept` / `DuplicateIdentity` / `ForeignNetwork` / `SelfConnection`) would let
each arm name its own remediation, with `Penalize` for foreign network. Related and pre-existing:
`PeerSpec.address` is `declaredAddress orElse localAddressOpt` while `getPeerAddress` is
`localAddressOpt orElse upnp orElse declaredAddress` — two names for "the peer's address" that
disagree whenever both are set, affecting `:221` and `:339` identically.

What would move me to approve: the bad-magic property, and one sentence in the description naming
that branch as the severe case. Thanks for the actor-level regressions — driving serialized
handshakes through the real `PeerConnectionHandler` rather than poking the controller directly is
what let me check the claim instead of taking it on faith.

PR #2426, opened 2026-07-14
Bound per-peer outbound buffering

a-shannon:
## Summary

- bound serialized outbound data retained for one backpressured peer by bytes
  and message count
- abort the connection before accepting an entry that would exceed either
  limit
- keep accounting exact across failed-write retries and acknowledgements

## Root cause

After a TCP `CommandFailed(Write(...))`, `PeerConnectionHandler` switched to an
ACK-based buffering mode. Every later outbound message was serialized and kept
in an unbounded `TreeMap` while the peer remained backpressured. A peer that
stopped reading could therefore make one connection retain serialized data
without a fixed ceiling.

This complements #2306: that change bounds individual inbound P2P messages,
while this change bounds aggregate outbound data retained for one peer.

## Limits and behavior

- byte limit: 16,388,621 bytes, equal to one maximum payload plus its framing
- message limit: 64 buffered entries
- overflow: clear retained state, send `Tcp.Abort`, and stop the handler
- no peer penalty, wire-format change, protocol-version change, or consensus
  change

An honestly slow peer may be disconnected and retry elsewhere rather than keep
an unbounded share of node heap. In particular, the transport does not retain
an entire application-level snapshot window after sustained backpressure.

## Tests

- `PeerConnectionHandlerSpecification`: 3/3 passed
- relevant network selection: 17 suites, 73/73 passed
- `DigestStateSpecification`: 7/7 passed in isolation
- mutation control: naive retry byte addition makes the retry property fail on
  an unexpected `Tcp.Abort`; replacement-aware accounting passes

The regression tests cover the byte boundary, the independent message-count
boundary, four maximum valid snapshot frames below the cap and the fifth above
it, a failed retransmission with the same ACK id without double-counting, exact
ACK removal, and ordered progress to the next buffered write.

### DIFF
--- a/src/main/scala/scorex/core/network/PeerConnectionHandler.scala
+++ b/src/main/scala/scorex/core/network/PeerConnectionHandler.scala
@@ -4,14 +4,19 @@ import akka.actor.{Actor, ActorRef, Cancellable, Props, SupervisorStrategy}
 import akka.io.Tcp
 import akka.io.Tcp._
 import akka.util.{ByteString, CompactByteString}
-import org.ergoplatform.network.{Handshake, HandshakeSerializer, PeerSpec, Version}
 import org.ergoplatform.network.Version.Eip37ForkVersion
-import scorex.core.app.ScorexContext
-import scorex.core.network.NetworkController.ReceivableMessages.{Handshaked, PenalizePeer}
-import scorex.core.network.PeerConnectionHandler.ReceivableMessages
+import org.ergoplatform.network.{Handshake, HandshakeSerializer, PeerSpec, Version}
+import org.ergoplatform.network.message.MessageConstants.{
+  ChecksumLength,
+  HeaderLength,
+  MaxMessageSize
+}
 import org.ergoplatform.network.message.MessageSerializer
 import org.ergoplatform.network.peer.{PeerInfo, PenaltyType}
 import org.ergoplatform.settings.ScorexSettings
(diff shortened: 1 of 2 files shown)


src/main/scala/scorex/core/network/PeerConnectionHandler.scala
```diff
@@ -226,7 +237,22 @@ class PeerConnectionHandler(scorexSettings: ScorexSettings,
   }
 
   private def buffer(id: Long, msg: ByteString): Unit = {
-    outMessagesBuffer += id -> msg
+    val previousMessage = outMessagesBuffer.get(id)
```
kushti, line 240: can we have duplicate id  ?

a-shannon: Yes, on a failed retry. `writeFirst()` resends the buffered entry with its existing `Ack(id)`. If that retry fails, `CommandFailed(Write(..., Ack(id)))` calls `buffer(id, msg)` again while the original entry is still present; it is removed only when `Ack(id)` is received. The duplicate therefore represents the same unacknowledged write, not a different message. `previousMessage` prevents that retry from being counted twice in the byte and message limits.

The replacement and accounting behavior for a repeated `Ack` id is covered by the `account retried and acknowledged writes exactly` property in `PeerConnectionHandlerSpecification`.

kushti (approved):

PR #2428, opened 2026-07-14
Reject unknown P2P message codes as malicious

a-shannon:
Classify unregistered P2P message codes as malicious protocol input so the existing permanent-penalty, blacklist and connection-teardown path handles them. Previously MessageSerializer returned Failure(Error); PeerConnectionHandler logged that generic failure, resumed reading and retained the invalid prefix, parsing it again as later data accumulated. Returning MaliciousBehaviorException selects the existing penalty/teardown behavior. Known codes, incomplete frames, payload bounds and the wire format are unchanged.

Head `b2e0c4690a836f124ba7bf831887a32140873776` remains targeted at `v6.0.7`, base `e927ab38e652423adcb79a41e8a052e348545320`. The [original two-file correction](https://github.com/a-shannon/ergo/commit/008167b40dad757888484bbd8e01e078c512af07), including its complete zero-payload unknown-code regression, is unchanged.

The [new eleven-file prerequisite increment](https://github.com/a-shannon/ergo/compare/6fe05788d67cae36b964e7e536419e23be63f6eb...b2e0c4690a836f124ba7bf831887a32140873776) adds settled-seed/isolation [D](https://github.com/a-shannon/ergo/commit/71f33c6e45a9394557cf3af1bfb422cbc53d26cf), owned by [#2535](https://github.com/ergoplatform/ergo/pull/2535), plus [G](https://github.com/a-shannon/ergo/commit/d62c945a4d6314c54a7b57f18a94aee8fd258aa7), owned by [#2511](https://github.com/ergoplatform/ergo/pull/2511), and [X](https://github.com/a-shannon/ergo/commit/a772a5cbf666584db3ed981838dbc11223d502ed), owned by [#2549](https://github.com/ergoplatform/ergo/pull/2549). G corrects full-summary anchors, tip/mode cache identity and retained-common-ancestor download scheduling. X revisits cached sections after their headers arrive. This adds four shared production paths and their tests to the cumulative PR; it does not alter the message-code correction.

Review the exact shared reader prerequisite [R](https://github.com/a-shannon/ergo/commit/76e5ec4bfc377334fdced4bd60a61df06e6b4cb6), owned by #2535, before G, then merge the owning #2511/#2549 corrections before this consumer. Existing [H](https://github.com/a-shannon/ergo/commit/0f9f75c1fc1df3ff3309482b83c0475a05cdd7a2), [U](https://github.com/a-shannon/ergo/commit/bcfcd61166151d8000ae2a05a62fa4f389aa7126) and [T](https://github.com/a-shannon/ergo/commit/0679fc061f13ad2169241e3ea9d96db7bcf5dba7) remain unchanged. Selected-header targets, all convergence assertions and original deadlines are retained. T's separate 2.5-second failure-only collection still rethrows the original timeout. Exact common commits avoid importing unrelated owner branches.

At `6fe05788`, native IT passed 49/50 with one ignored case; DeepRollBack alone timed out at minimum 221. The older [IT job](https://github.com/ergoplatform/ergo/actions/runs/33871364229/job/101017875977) passed 12/13 with one ignored case and failed the flattened stored-header comparison. These observations are historical; G/X are independently demonstrated corrections, not exclusive attribution of either CI failure.

Related scopes remain distinct: [#2320](https://github.com/ergoplatform/ergo/pull/2320) substitutes ErgoException, which still takes the log-only path; [#2366](https://github.com/ergoplatform/ergo/pull/2366) checks peer versions after message identification; declared-length bounds are upstream via [#2306](https://github.com/ergoplatform/ergo/pull/2306), covering closed [#2381](https://github.com/ergoplatform/ergo/pull/2381); and [#2404](https://github.com/ergoplatform/ergo/pull/2404)/[#2405](https://github.com/ergoplatform/ergo/pull/2405) reject trailing payload bytes.

The actual final composition passed normal compilation and 49/49 focused synchronizer/history/cache tests, including UTXO and Digest wake-up cases. Its preceding D composition passed normal IT compilation and H/D 33/33. Unchanged prior H/T 37/37, MessageSerializer 5/5 and network 71/71 evidence remains historical and input-qualified; the original negative control failed with java.lang.Error before the correction. Independent source review found no blockers. Local checks did not run Docker or establish full CI success. New-head native CI remains pending in [#2533](https://github.com/ergoplatform/ergo/issues/2533).

### DIFF
--- a/src/it/scala/org/ergoplatform/it/DeepRollBackIsolationSpec.scala
+++ b/src/it/scala/org/ergoplatform/it/DeepRollBackIsolationSpec.scala
@@ -1,0 +1,58 @@
+package org.ergoplatform.it
+
+import akka.actor.{ActorRef, ActorSystem}
+import akka.io.Tcp
+import akka.testkit.{ExplicitlyTriggeredScheduler, TestActorRef, TestProbe}
+import com.typesafe.config.ConfigFactory
+import org.ergoplatform.network.message.MessageConstants.MessageCode
+import org.ergoplatform.network.peer.PeerManager.ReceivableMessages.RandomPeerExcluding
+import org.ergoplatform.settings.ScorexSettings
+import org.ergoplatform.utils.ErgoNodeTestConstants.settings
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import scorex.core.app.ScorexContext
+import scorex.core.network.NetworkController
+
+import java.net.InetSocketAddress
+import scala.concurrent.{Await, ExecutionContext}
+import scala.concurrent.duration._
+
+class DeepRollBackIsolationSpec extends AnyFlatSpec with Matchers {
+  "Isolated rollback mining" should "disable automatic peer selection and reject incoming connections" in {
+    val ordinaryConfig = ConfigFactory.load()
+    val isolatedNetwork = ScorexSettings.fromConfig(
+      DeepRollBackSpec.isolatedMiningConfig.withFallback(ordinaryConfig).resolve()).network
+    ScorexSettings.fromConfig(ordinaryConfig).network.maxConnections should be > 0
+
+    implicit val system: ActorSystem = ActorSystem("RollbackIsolation", ConfigFactory.parseString(
+      "akka.scheduler.implementation = akka.testkit.ExplicitlyTriggeredScheduler"))
+    implicit val ec: ExecutionContext = system.dispatcher
+    try {
+      def controller(maxConnections: Int): (TestActorRef[NetworkController], TestProbe) = {
+        val peers = TestProbe()
+        val tcp = TestProbe()
+        val controllerSettings = settings.copy(scorexSettings = settings.scorexSettings.copy(
+          network = settings.scorexSettings.network.copy(maxConnections = maxConnections)))
+        val ref = TestActorRef(new NetworkController(controllerSettings, peers.ref,
+          ScorexContext(Seq.empty, None, None), tcp.ref, _ => Map.empty[MessageCode, ActorRef]))
+        tcp.expectMsgType[Tcp.Bind]
+        ref ! Tcp.Bound(controllerSettings.scorexSettings.network.bindAddress)
+        (ref, peers)
+      }
+
+      val (isolated, isolatedPeers) = controller(isolatedNetwork.maxConnections)
+      val (_, ordinaryPeers) = controller(maxConnections = 1)
+      system.scheduler.asInstanceOf[ExplicitlyTriggeredScheduler].timePasses(5.seconds)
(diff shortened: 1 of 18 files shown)


jozanek (commented):
I read this as making an unregistered code cost one frame instead of the whole connection. The skip looks framing-safe to me: it runs only after magic, length and checksum pass, and the length matches `MessageBase.messageLength`.

Two questions. As far as I can tell, senders already gate new codes by peer version (`NetworkController` filter, plus call-site checks on `weak-blocks`), so which senders did you have in mind? And the move from penalty to skip cites a maintainer clarification I couldn't find on GitHub; could you link or quote it, so the policy is on record?

One follow-up I'd keep separate: `SendToPeer`/`SendToPeers` bypass the central version filter (`SendingStrategy.scala:23-29`). With the inline points settled, I'd be comfortable approving.

src/main/scala/scorex/core/network/PeerConnectionHandler.scala
```diff
@@ -202,6 +203,10 @@ class PeerConnectionHandler(scorexSettings: ScorexSettings,
             chunksBuffer = chunksBuffer.drop(message.messageLength)
             process()
           case Success(None) =>
+          case Failure(UnknownMessageCodeException(code, messageLength)) =>
+            log.debug(s"Skipping unsupported message code $code from $connectionId")
+            chunksBuffer = chunksBuffer.drop(messageLength)
+            process()
           case Failure(e) =>
```
jozanek, line 216: The non-malicious `case _` arm below (:218–219) still logs and leaves the head frame in `chunksBuffer`. As far as I can see nothing reaches it today, but a future throw in `deserialize` would bring back the stuck connection this PR removes. Would you be open to closing the connection there, without a penalty?

src/main/scala/scorex/core/network/PeerConnectionHandler.scala
```diff
@@ -202,6 +203,10 @@ class PeerConnectionHandler(scorexSettings: ScorexSettings,
             chunksBuffer = chunksBuffer.drop(message.messageLength)
             process()
           case Success(None) =>
+          case Failure(UnknownMessageCodeException(code, messageLength)) =>
+            log.debug(s"Skipping unsupported message code $code from $connectionId")
```
jozanek: On the base this case surfaced at INFO ("Corrupted data"); now it is DEBUG only, and skipped frames don't refresh peer activity, so operators see nothing by default. An unknown code from a peer is also the earliest sign of a sender-side gating slip. Could this log once per connection and code at INFO, with the peer's version?

jozanek (approved):
Both inline points are addressed in `de10210a4`: unexpected parser failures now close without a penalty, and unsupported codes log once per connection and code at INFO with the peer version. The quoted policy and the sender-scope note answer my questions. The `SendToPeer`/`SendToPeers` gap stays a separate follow-up.

PR #2429, opened 2026-07-14
Reject invalid first mining solutions without restart

a-shannon:
## Summary

- Reject an invalid solution for the first cached mining candidate without dereferencing an absent previous candidate.
- Return a normal `StatusReply.Error` while preserving the current candidate template for later valid work.
- Add actor-level regression coverage for the exact error and cached-object identity.

## Root cause

When a solution failed validation against the current candidate, `CandidateGenerator` unconditionally retried it against `cachedPreviousCandidate.get`. The first generated candidate has no previous candidate, so an invalid solution threw `NoSuchElementException`, restarted the actor and returned no normal error to the external miner.

The fallback is now optional. If no previous candidate exists, the actor returns an error and keeps the valid candidate template. Current valid solutions, previous-candidate fallback, final PoW validation, block publication and solved-state transitions are unchanged.

## Related work

PR #2411 removes duplicate PoW validation on `master`, but still dereferences `cachedPreviousCandidate.get`; this change covers only the missing-previous branch on `v6.0.7`.

## Tests

- RED on `v6.0.7` (`e927ab38`): `NoSuchElementException: None.get`, followed by the expected-reply timeout.
- `CandidateGeneratorSpec`: 20/20 passed.
- `CandidateGeneratorPropSpec` + `MiningApiRouteSpec`: 23/23 passed.
- Total focused mining/API closure: 43/43 passed.
- Independent exact-diff review: no Critical or Important findings.
- `git diff --check` and the publication guard passed.

### DIFF
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -251,28 +251,41 @@ class CandidateGenerator(
         } else {
           preSolution
         }
-      val result: StatusReply[Unit] = {
-        val newBlock = state.cachedCandidate
-          .map(candidate => completeBlock(candidate.candidateBlock, solution))
-          .filter(block => ergoSettings.chainSettings.powScheme.validate(block.header).isSuccess)
-          .getOrElse {
-            log.info(s"Using previous candidate as a solution: " + state.cachedPreviousCandidate)
-            completeBlock(state.cachedPreviousCandidate.get.candidateBlock, solution)
-          }
-        log.info(s"New block mined, header: ${newBlock.header}")
-        ergoSettings.chainSettings.powScheme.validate(newBlock.header) match {
-          case Success(_) =>
-            sendToNodeView(newBlock)
-            context.become(initialized(state.copy(solvedBlock = Some(newBlock))))
-            StatusReply.success(())
-          case Failure(exception) =>
-            log.warn(s"Removing candidates due to invalid block", exception)
-            context.become(initialized(state.copy(cachedCandidate = None, cachedPreviousCandidate = None)))
-            StatusReply.error(
-              new Exception(s"Invalid block mined: ${exception.getMessage}", exception)
-            )
+      val newBlockOpt = state.cachedCandidate
+        .map(candidate => completeBlock(candidate.candidateBlock, solution))
+        .filter(block => ergoSettings.chainSettings.powScheme.validate(block.header).isSuccess)
+        .orElse {
+          log.info(
+            s"Using previous candidate as a solution: ${state.cachedPreviousCandidate}"
+          )
+          state.cachedPreviousCandidate.map(candidate =>
+            completeBlock(candidate.candidateBlock, solution)
+          )
         }
-      }
+      val result: StatusReply[Unit] = newBlockOpt match {
+        case Some(newBlock) =>
+          log.info(s"New block mined, header: ${newBlock.header}")
+          ergoSettings.chainSettings.powScheme.validate(newBlock.header) match {
+            case Success(_) =>
+              sendToNodeView(newBlock)
+              context.become(initialized(state.copy(solvedBlock = Some(newBlock))))
+              StatusReply.success(())
+            case Failure(exception) =>
+              log.warn(s"Removing candidates due to invalid block", exception)
+              context.become(
+                initialized(
+                  state.copy(cachedCandidate = None, cachedPreviousCandidate = None)
+                )
+              )
+              StatusReply.error(
+                new Exception(s"Invalid block mined: ${exception.getMessage}", exception)
+              )
+          }
+        case None =>
+          StatusReply.error(
+            "Invalid solution for current candidate and no previous candidate available"
+          )
+        }
       log.info(s"Processed solution $solution with the result $result")
       sender() ! result
 
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -215,27 +215,42 @@ class CandidateGenerator(
         } else {
           preSolution
         }
-      val result: StatusReply[Unit] = {
-        val newBlock = state.cachedCandidate
-          .map(candidate => completeBlock(candidate.candidateBlock, solution))
-          .filter(block => ergoSettings.chainSettings.powScheme.validate(block.header).isSuccess)
-          .getOrElse {
-            log.info(s"Using previous candidate as a solution: " + state.cachedPreviousCandidate)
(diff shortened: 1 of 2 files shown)


cafebedouin (commented):
Review carried out by Claude (Anthropic, Claude Fable 5.1) for cafebedouin, with independent passes by Grok and Gemini. Taken from the #2533 review table.

**Commit:** `d04240bc7` (base `v6.0.7`).

**Scope:** both files in the diff; `CandidateGenerator.scala:245-295` at this head; the same lines in #2411 (base `master`) at `1fa1076e9`.

**Result:** the missing-previous-candidate fix is correct as read. With `cachedPreviousCandidate` empty, the old `.get` threw inside `receive`; the new `orElse { … .map(…) }` yields `None`, and the `None` arm replies with an error and leaves `state` alone. The spec pins both halves: the error text, and `theSameInstanceAs` on the next `GenerateCandidate`, which would fail after a restart because a restarted actor has no cached candidate.

**Tests run:** none locally. Read from the PR's CI: all 8 checks green at `d04240bc7`.

**[Integration] #2411 and this PR edit the same lines from the same blob (`080c64409b`, `:254-266`) on different bases.** They will collide when `v6.0.7` reaches `master`, and as your description says #2411 keeps `state.cachedPreviousCandidate.get`.

Recommended: give #2411 the same optional fallback and `None` arm now, so that merge resolves to one shape instead of reintroducing the `.get`.

**[Nit]** `CandidateGenerator.scala:258`: the "Using previous candidate as a solution" line still prints when there is none (`…: None`), as it did before the throw. Moving it inside the `.map` keeps the log truthful.

**[Nit]** `:284`: the `None` arm replies with a fixed string, while the other failure arm passes on `exception.getMessage`. `.filter(_.isSuccess)` has already discarded the PoW failure by then; keeping the `Try` and including its message gives a pool operator the same reason on both paths.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -251,28 +251,41 @@ class CandidateGenerator(
...
         }
-      val result: StatusReply[Unit] = {
-        val newBlock = state.cachedCandidate
-          .map(candidate => completeBlock(candidate.candidateBlock, solution))
-          .filter(block => ergoSettings.chainSettings.powScheme.validate(block.header).isSuccess)
-          .getOrElse {
-            log.info(s"Using previous candidate as a solution: " + state.cachedPreviousCandidate)
-            completeBlock(state.cachedPreviousCandidate.get.candidateBlock, solution)
+      val newBlockOpt = state.cachedCandidate
+        .map(candidate => completeBlock(candidate.candidateBlock, solution))
```
jozanek: A nonce that isn't 8 bytes (say `"n": "00"`) still throws on this line: `completeBlock` needs `header.id`, and serializing the header hits `require(obj.n.length == 8)` (`AutolykosSolution.scala:98`) outside any `Try`. `ErgoMiner` forwards solutions unchecked and the route needs no API key, so the actor restarts as before. Would you be open to rejecting a wrong-length `n` up front with `StatusReply.error`, plus a one-byte-nonce test? I found this by reading, not by running it.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -251,28 +251,41 @@ class CandidateGenerator(
...
+        .filter(block => ergoSettings.chainSettings.powScheme.validate(block.header).isSuccess)
+        .orElse {
+          log.info(
```
jozanek: In the case this PR handles, this logs `Using previous candidate as a solution: None`. When a previous candidate exists, it logs the whole candidate block as JSON (`CandidateBlock.scala:21`) at INFO on every rejected POST. I agree with moving it inside the `.map`, as cafebedouin suggested. Would logging only the candidate's header id there work, to keep logs readable?

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -63,6 +64,15 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
 
   private val defaultSettings60 = defaultSettings.copy(networkType = DevNet60, directory = defaultSettings.directory + "60")
 
+  private class ToggleRejectingPowScheme(k: Int, n: Int)
+      extends DefaultFakePowScheme(k, n) {
+    @volatile var rejectSolutions: Boolean = false
+
+    override def validate(header: Header): Try[Unit] =
```
jozanek, line 72: This rejects every header, and test PoW is `fake` (`validate` always succeeds, `DefaultFakePowScheme.scala:15`), so I think no test reaches the reshaped fallback: current candidate rejects, previous accepts. Would you be open to rejecting only the current candidate's header and adding that case, plus asserting that no block reaches the node view on rejection?

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -87,6 +97,60 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
...
+      .get
+      .header
+      .powSolution
+    powScheme.rejectSolutions = true
+
+    candidateGenerator.tell(solution, testProbe.ref)
+
+    testProbe.expectMsgPF(blockValidationDelay) {
+      case StatusReply.Error(error) =>
+        error.getMessage shouldBe
```
jozanek: #2411 answers the same case with this text plus the PoW cause, and asserts with `should include`. With `shouldBe` here, whichever message wins at integration breaks one of the two tests. Would `include` work here, so that merge stays mechanical?

jozanek (commented):
What I understood this PR to be for: stopping an invalid `/mining/solution` from restarting `CandidateGenerator` when no previous candidate is cached. From reading, that window is wider than the first candidate: `FullBlockApplied` clears both slots (`CandidateGenerator.scala:178-183`), so it reopens after every block until the first regeneration (60s or more), and a stale share right after a network block lands in it. That makes the fix worth more than the description says. Would you be open to noting it there?

A question I'm holding: kushti's `weak-blocks` merge (`c216c5b6e`) handles the same `None` case by clearing both caches, while this PR keeps them. I lean towards keeping, since a failed PoW says nothing about the candidate, but a line on why would help the branches converge.

What would move me to approve: the malformed-nonce path noted inline. I ran no builds.

jozanek (approved):
Thanks, this covers everything I raised: the nonce guard (`:246`) with 0/1/9-byte tests, the fallback now exercised in both directions via `HeaderRejectingPowScheme`, the header-id-only log, and the error text aligned with #2411. The description now explains the keep-cache choice and the per-block window, and your correction on the 60s setting is fair.

PR #2432, opened 2026-07-16
Count pending inbound connections toward the limit

a-shannon:
Count raw inbound sockets awaiting `PeerManager` confirmation toward `incomingLimit`. Previously, a burst of `Tcp.Connected` events could all pass admission before the confirmation replies created established handlers.

## Focused release increment

The branch is rebuilt on `v6.0.6` at `8995ff243b140e7fe716bd3204423dd35690c2e4`. Review the [two-file correction](https://github.com/a-shannon/ergo/compare/8995ff243b140e7fe716bd3204423dd35690c2e4...810843cdf3ff51b0a1ff2eabb6d0b631ac93b43d).

The previous head `6924665a2ca45c2a90c3c7c808e221fd131add5e` remains the historical review reference. Its unrelated #2427 blacklist backport and pending-blacklist expansion have been removed. This PR no longer carries the shared blacklist changes also encountered in [#2449](https://github.com/ergoplatform/ergo/pull/2449) and [#2461](https://github.com/ergoplatform/ergo/pull/2461).

## September 12 review corrections

- Pending reservations are keyed by the raw connection `ActorRef`. Admission counts established incoming handlers plus pending sockets; confirmation transfers a reservation, so no second numeric limit check is needed.
- Denial or raw actor termination releases its reservation. A late confirmation must still own a pending reservation. Only actually pending actors are unwatched.
- Two sockets sharing a remote endpoint cannot overwrite the first established handler when their confirmations arrive. The listener can accept connections on different local interfaces, so the same remote endpoint does not imply the same TCP connection. Established handler termination retains its existing disconnect notification.
- `ConnectionClosed` keeps the release's logging behavior. Pending cleanup uses the raw actor lifecycle; the tests stop that actor instead of injecting a synthetic close event as proof of termination.
- Admission logging reports established count, pending count and limit. Blacklist admission remains with the existing `PeerManager` path.

## Validation

The controller suite passes 15 tests. The bounded network, peer and peers-API closure passes 25 tests across seven suites. The admission regression fails on the original release implementation. Separate single-fault mutations of the pending-membership and endpoint-uniqueness guards each fail their targeted race regression; restoring both guards passes all 15 controller tests. Independent source review is complete.

Local execution used JDK 8 and Scala 2.12.20. Current GitHub checks bind head `810843cdf3ff51b0a1ff2eabb6d0b631ac93b43d`; the [checks page](https://github.com/ergoplatform/ergo/pull/2432/checks) reports the live result.

## Integration

No prerequisite on #2449 or #2461 remains. [#2421](https://github.com/ergoplatform/ergo/pull/2421) independently corrects rejected-handshake cleanup in the same controller. Preserve both corrections and the union of their actor fixtures when integrating. The current merge/review plan is maintained in [#2533](https://github.com/ergoplatform/ergo/issues/2533).

All eight [CI jobs passed](https://github.com/ergoplatform/ergo/actions/runs/34720760986) on the current head. The [published composition with #2421](https://github.com/a-shannon/ergo/commit/cacb2dcb5d453e06200747c8e1612b3d7715a38b) passes all 18 controller cases; its [increment after this PR](https://github.com/a-shannon/ergo/compare/810843cdf3ff51b0a1ff2eabb6d0b631ac93b43d...cacb2dcb5d453e06200747c8e1612b3d7715a38b) preserves both corrections and their owned tests.

### DIFF
--- a/src/main/scala/scorex/core/network/NetworkController.scala
+++ b/src/main/scala/scorex/core/network/NetworkController.scala
@@ -63,6 +63,7 @@ class NetworkController(ergoSettings: ErgoSettings,
 
   private var connections = Map.empty[InetSocketAddress, ConnectedPeer]
   private var unconfirmedConnections = Set.empty[InetSocketAddress]
+  private var unconfirmedIncomingConnections = Map.empty[InetSocketAddress, ActorRef]
 
   private val mySessionIdFeature = SessionIdPeerFeature(networkSettings.magicBytes)
   /**
@@ -168,6 +169,17 @@ class NetworkController(ergoSettings: ErgoSettings,
           Option(peer.connectionId.remoteAddress.getAddress).contains(blacklistedIp)
         }.toSeq
         peersToClose.foreach(_.handlerRef ! CloseConnection)
+
+        val pendingToClose = unconfirmedIncomingConnections.iterator.collect {
+          case (address, connectionRef)
+            if Option(address.getAddress).contains(blacklistedIp) =>
+            address -> connectionRef
+        }.toSeq
+        pendingToClose.foreach { case (address, connectionRef) =>
+          unconfirmedIncomingConnections -= address
+          context.unwatch(connectionRef)
+          connectionRef ! Close
+        }
       }
   }
 
@@ -179,13 +191,20 @@ class NetworkController(ergoSettings: ErgoSettings,
       log.info(s"Unconfirmed connection: ($remoteAddress, $localAddress) => $connectionId")
       if (connectionDirection.isOutgoing) {
         createPeerConnectionHandler(connectionId, sender())
+      } else if (unconfirmedIncomingConnections.contains(remoteAddress)) {
+        log.warn(s"Connection to peer $remoteAddress is already pending")
+        sender() ! Close
       } else {
-        val incomingCount = connections.values.count(_.connectionId.direction.isIncoming)
+        val incomingCount = connections.values.count(_.connectionId.direction.isIncoming) +
+          unconfirmedIncomingConnections.size
         if (incomingCount >= incomingLimit) {
           log.info(s"Incoming connection from $remoteAddress denied: too many incoming connections ($incomingCount)")
           sender() ! Close
         } else {
-          peerManagerRef ! ConfirmConnection(connectionId, sender())
+          val connectionRef = sender()
+          unconfirmedIncomingConnections += remoteAddress -> connectionRef
+          context.watch(connectionRef)
+          peerManagerRef ! ConfirmConnection(connectionId, connectionRef)
         }
       }
 
@@ -194,11 +213,39 @@ class NetworkController(ergoSettings: ErgoSettings,
       sender() ! Close
 
     case ConnectionConfirmed(connectionId, handlerRef) =>
-      log.info(s"Connection confirmed to $connectionId")
-      createPeerConnectionHandler(connectionId, handlerRef)
+      if (connectionId.direction.isIncoming) {
+        val remoteAddress = connectionId.remoteAddress
+        if (unconfirmedIncomingConnections.get(remoteAddress).contains(handlerRef)) {
+          unconfirmedIncomingConnections -= remoteAddress
+          context.unwatch(handlerRef)
+          val incomingCount = connections.values.count(_.connectionId.direction.isIncoming)
+          if (incomingCount >= incomingLimit) {
+            log.info(s"Incoming connection from $remoteAddress denied: too many incoming connections ($incomingCount)")
+            handlerRef ! Close
+          } else {
+            log.info(s"Connection confirmed to $connectionId")
+            createPeerConnectionHandler(connectionId, handlerRef)
+          }
+        } else {
+          log.info(s"Ignoring stale incoming connection confirmation from $remoteAddress")
+          handlerRef ! Close
+        }
+      } else {
+        log.info(s"Connection confirmed to $connectionId")
+        createPeerConnectionHandler(connectionId, handlerRef)
+      }
 
     case ConnectionDenied(connectionId, handlerRef) =>
       log.info(s"Incoming connection from ${connectionId.remoteAddress} denied")
+      if (connectionId.direction.isIncoming) {
+        val remoteAddress = connectionId.remoteAddress
+        if (unconfirmedIncomingConnections.get(remoteAddress).contains(handlerRef)) {
+          unconfirmedIncomingConnections -= remoteAddress
+          context.unwatch(handlerRef)
+        } else {
+          log.info(s"Ignoring stale incoming connection denial from $remoteAddress")
+        }
+      }
       handlerRef ! Close
 
     case Handshaked(connectedPeer) =>
@@ -222,6 +269,14 @@ class NetworkController(ergoSettings: ErgoSettings,
       }
 
     case Terminated(ref) =>
+      val pendingAddress = unconfirmedIncomingConnections.collectFirst {
+        case (address, connectionRef) if connectionRef == ref => address
+      }
+      pendingAddress.foreach { address =>
+        log.info(s"Pending incoming connection from $address terminated")
+        unconfirmedIncomingConnections -= address
+      }
+
       connectionForHandler(ref) match {
         case Some(connectedPeer) =>
           log.info(s"Terminating connection to $connectedPeer")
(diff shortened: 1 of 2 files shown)


src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -63,6 +63,7 @@ class NetworkController(ergoSettings: ErgoSettings,
 
   private var connections = Map.empty[InetSocketAddress, ConnectedPeer]
   private var unconfirmedConnections = Set.empty[InetSocketAddress]
+  private var unconfirmedIncomingConnections = Map.empty[InetSocketAddress, ActorRef]
```
jozanek: This tracks pending sockets by `InetSocketAddress`, though each one already carries a stronger
identity: its connection `ActorRef`, which is one-per-connection by construction.

Endpoint keying introduces an ambiguity TCP itself does not have, and I count four branches added
to absorb it — the duplicate check at :194, the ref guards at :218 and :242, and the sweep at
:293. Held as `Set[ActorRef]`, removal is idempotent, so a stale reply is a no-op and none of
those four is needed; the accounting lands in roughly eight lines and the spec shrinks with it.

Would you be open to trying that? If endpoint keying carries a requirement I have not spotted,
I would like to understand it.

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -194,11 +213,39 @@ class NetworkController(ergoSettings: ErgoSettings,
...
 
     case ConnectionConfirmed(connectionId, handlerRef) =>
-      log.info(s"Connection confirmed to $connectionId")
-      createPeerConnectionHandler(connectionId, handlerRef)
+      if (connectionId.direction.isIncoming) {
+        val remoteAddress = connectionId.remoteAddress
+        if (unconfirmedIncomingConnections.get(remoteAddress).contains(handlerRef)) {
+          unconfirmedIncomingConnections -= remoteAddress
+          context.unwatch(handlerRef)
+          val incomingCount = connections.values.count(_.connectionId.direction.isIncoming)
```
jozanek: I do not think this second limit check can fire. Admission at :198 maintains
`established + pending <= incomingLimit`, and the pending entry is removed at :219 before this
count is taken, so `established` is at most `incomingLimit - 1` by the time we get here.

If that reading is right, the branch is unreachable and the suite asserts on it. What interests
me more than the dead code is what it suggests: the invariant is real but is not named anywhere,
so it is hard to trust by reading. Would you be open to dropping the check, or to keeping it with
a comment that states the invariant?

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -163,11 +164,22 @@ class NetworkController(ergoSettings: ErgoSettings,
       peerManagerRef ! PeerManager.ReceivableMessages.Penalize(peerAddress, penaltyType)
 
     case Blacklisted(peerAddress) =>
-      connections.get(peerAddress).foreach { peer =>
-        connections = connections.filterNot { case (address, _) => // clear all connections related to banned peer ip
-          Option(peer.connectionId.remoteAddress.getAddress).exists(Option(address.getAddress).contains(_))
+      Option(peerAddress.getAddress).foreach { blacklistedIp =>
```
jozanek: This hunk is byte-identical to what is already on `master` via #2427, and to the same hunk in
open PRs #2449 and #2461 — so three PRs currently carry it into `v6.0.6`. I checked all three
diffs. It also brings ~122 test lines that already exist on `master` verbatim.

The cost lands on whoever merges second, and #2449 is kushti's own. Would you be open to dropping
this commit and rebasing once one of the others lands? That would also bring the PR back to the
single concern its title names.

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -222,18 +269,31 @@ class NetworkController(ergoSettings: ErgoSettings,
...
+        case None if pendingAddress.isEmpty =>
           log.warn(s"No connection found for $ref during termination")
+        case None => ()
       }
 
     case _: ConnectionClosed =>
+      val connectionRef = sender()
+      unconfirmedIncomingConnections =
+        unconfirmedIncomingConnections.filterNot { case (_, ref) => ref == connectionRef }
+      context.unwatch(connectionRef)
```
jozanek: This unwatches `sender()` on every `ConnectionClosed`, including refs that were never in the
pending map.

It looks harmless today, since `PeerConnectionHandler` consumes close events itself and never
forwards them to the controller. The controller does also watch its handlers, though, so if one
ever forwarded, that handler would be silently unwatched, its `connections` entry would leak, and
`DisconnectedPeer` would stop firing for it.

Would you be open to moving the unwatch inside the branch where the ref was actually found?

jozanek (commented):
Thanks for bringing this back as a standalone change. Having traced the history, I think that was
the right call, and I want to put the evidence in the thread so the next reader does not have to
re-derive it.

**What I understood this to be for.** The node documents an inbound limit of 22 and does not
enforce it: admission counts established handlers only, so sockets in flight to `PeerManager` are
invisible to it. The gap arrived with #2306 — kushti's own `incomingLimit` commits of 25 May — and
the fix was proposed the next day inside #2336. When #2336 was closed as "incorporated into 6.0.3
canidate via #2306", this hunk was not in fact incorporated: #2306 took the Docker and IT changes,
the serializer tests and a narrower `PeerManager` guard, and wrote its own controller spec whose
limit tests drive each socket fully through confirmation, which is exactly why they never see the
burst. So this reads as overlooked rather than declined. Saying that in the PR body would save the
next reviewer the dig.

**I would make a stronger case than the body does.** It describes "raw inbound sockets", which
sounds cheap. But there is no limit check on `ConnectionConfirmed` today, so each over-admitted
socket becomes a full `PeerConnectionHandler` holding a slot for up to `handshakeTimeout = 30s`,
and `connections.size` also gates outbound dialing — so an inbound burst suppresses the node's own
peer discovery. The controller's mailbox is unbounded (`bounded-mailbox` binds via
`RequiresMessageQueue`, which this actor does not declare), so the overshoot is bounded by
scheduling latency rather than any configured cap. Under a stall that is not a small number.

**Where I get stuck is the shape, not the goal.** My main question is the inline one at the new
field: tracking pending connections by endpoint rather than by `ActorRef` seems to create the
ambiguity that four of the five new branches then resolve, one of which I believe is unreachable.
I may be missing a requirement that forces the endpoint key.

**What I could not verify.** I ran no build tooling for this review, so all of it is a read.
- Three of the new tests release a pending slot by sending `Tcp.PeerClosed` to the controller from
  the connection probe. My reading of akka 2.6.10 is that an un-`Register`ed connection actor is
  not reading, and close events go to the registered handler, which here is never the controller.
  If that is right, those three defend a path production does not produce. What am I missing?
- `v6.0.7` exists, and `v6.0.6` is 7 ahead and 1 behind it while sitting 55 behind `master`. Which
  branch should this land on, and who forward-ports? That also decides whether the blacklist
  backport is needed here at all.

**What would move me to approve.** Either the `ActorRef` keying or your reasoning for why the
endpoint key is necessary; the blacklist commit dropped so the PR matches its title; and the
unwatch narrowed to pending refs. None of that is a correctness objection to what is here — the
invariant you are enforcing is one the node already claims to have, and nobody else is enforcing
it.

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -179,13 +191,20 @@ class NetworkController(ergoSettings: ErgoSettings,
       log.info(s"Unconfirmed connection: ($remoteAddress, $localAddress) => $connectionId")
       if (connectionDirection.isOutgoing) {
         createPeerConnectionHandler(connectionId, sender())
+      } else if (unconfirmedIncomingConnections.contains(remoteAddress)) {
+        log.warn(s"Connection to peer $remoteAddress is already pending")
+        sender() ! Close
       } else {
-        val incomingCount = connections.values.count(_.connectionId.direction.isIncoming)
+        val incomingCount = connections.values.count(_.connectionId.direction.isIncoming) +
+          unconfirmedIncomingConnections.size
```
jozanek: After this change the count in the denial log just below blends two different populations.

An operator reading "too many incoming connections (22)" cannot tell 22 settled peers from a few
peers plus a backlog of pending entries. Neither `/peers/connected` nor `peersCount` in `/info`
exposes the pending set, and there is no metrics system in the node, so this line is the only
signal available for the mechanism being added here.

Would you be open to logging the two numbers separately?

jozanek (commented):
Thanks for the rework — this addresses everything I raised, at about half the production size.

One correction I owe you: I overstated the `ActorRef` case. Three of the four branches collapsed, not four. You were right to keep the staleness guard on confirmation, since idempotent removal protects the accounting but does not stop the controller building a handler for a socket already reaped.

I also went looking for a reason to doubt your comment about shared endpoints on different local addresses, on the assumption that one `Bind` meant one local address. It does not — `bindAddress` defaults to `0.0.0.0:9052`, so on a multi-homed host those 4-tuples really are distinct. Your guard is right and I was wrong to question it. Worth a line in the PR body, since it is the least obvious part of the change.

The one thing still open is not code: this targets `v6.0.6`, which is 1 behind `v6.0.7` and 55 behind `master`. Which branch should it land on, and who forward-ports? Separately, `v6.0.6` still has the old blacklist handler. Dropping it from here was right, but the fix now rides on #2449 or #2461, both of which carry it incidentally alongside unrelated work — worth making sure one of them reaches `v6.0.6` before 6.0.6 ships.

jozanek (approved):

kushti (approved):

PR #2440, opened 2026-07-20
Fix duplicate transaction ids in OrderedTxPool.updateFamily

arkadianet:
## Problem
`OrderedTxPool` holds `transactionsRegistry` (id → `WeightedTxId`) and `orderedTransactions` (a `TreeMap` keyed and ordered by `(-weight, id)`) that must agree on each id's key. `WeightedTxId` equality is id-only, so a same-id key at a different weight is a *distinct* `TreeMap` entry.

`updateFamily` captures each parent's `WeightedTxId` once from the pre-fold snapshot, then re-keys with `orderedTransactions - staleWtx + (newWtx -> ut)`. On a reconvergent DAG — a transaction spending a parent `P` **plus** ≥2 of `P`'s other children — earlier fold iterations re-key `P` to a higher weight, so the later `- staleWtx` removes nothing and `+ newWtx` inserts a **second ordered entry for `P`'s id**. The registry then names one key while `orderedTransactions` holds two.

This is the *producer* of the duplicate ids observed after 6.0.3: `a21ece610` hardened only the removal-side (`remove`/`invalidate`) cleanup, which does not fire here because the registry key is present.

## Fix
Re-resolve each parent's current key from `transactionsRegistry` inside the fold, so the registry and the ordered map cannot diverge. This is a producer-only change; the 6.0.3 removal-side cleanup is left intact as a complement.

## Test
`ErgoMemPoolSpec` — *"not produce duplicate ids for a multi-diamond family put through the pool"*: starts from `OrderedTxPool.empty`, `put()`s `P`, `C1`, `C2` (each child spending `P`), then `put(X)` where `X` spends `C1`, `C2` and `P` with inputs ordered children-first so the fold reaches `P` last. It asserts no duplicate id in `orderedTransactions`, registry/ordered-map consistency, and `get(id)` resolution — over two distinct source boxes (two id sets), so it is not id-order-dependent. Reverting only the source change fails it deterministically.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/mempool/OrderedTxPool.scala
@@ -208,20 +208,28 @@ class OrderedTxPool(val orderedTransactions: TreeMap[WeightedTxId, UnconfirmedTr
       this
     } else {
 
-      val uniqueTxIds: Set[WeightedTxId] = tx.inputs.flatMap(input => this.outputs.get(input.boxId)).toSet
-      val parentTxs = uniqueTxIds.flatMap(wtx => this.orderedTransactions.get(wtx).map(ut => wtx -> ut))
+      val parentIds: Set[ModifierId] =
+        tx.inputs.flatMap(input => this.outputs.get(input.boxId).map(_.id)).toSet
 
-      parentTxs.foldLeft(this) { case (pool, (wtx, ut)) =>
-        val parent = ut.transaction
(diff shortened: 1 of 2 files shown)


arkadianet:
Independently arrived at the same root cause and reconvergent-DAG regression (P/C1/C2 then `put(X)`) while investigating the empty-block reports — #2439 covers it more completely (same producer fix plus the removal-side cleanup and a broader test suite), so closing in favor of it. Good to have the diagnosis confirmed from two independent angles. 👍

PR #2442, opened 2026-07-20
Preserve extra index consistency across chain switches

### LINKED ISSUE #2441: Extra indexer : not found input box
10:22:11.254 WARN  [ref-indexer-dispatcher-34] o.e.n.h.extra.ExtraIndexer - Unknown box used as input: e6dfd4938e9b19861ccbf0386305d0a661f2c17eacd312658174773cf2796433
10:22:11.254 WARN  [ref-indexer-dispatcher-34] o.e.n.h.extra.ExtraIndexer - Not found input box: e6dfd4938e9b19861ccbf0386305d0a661f2c17eacd312658174773cf2796433
13:21:38.093 WARN  [ef-indexer-dispatcher-397] o.e.n.h.extra.ExtraIndexer - Unknown box used as input: 7ecad68f9df9812665f568775e66e2e5a2f3d6ee35e317c11629d52e2891c9bd
13:21:38.093 WARN  [ef-indexer-dispatcher-397] o.e.n.h.extra.ExtraIndexer - Not found input box: 7ecad68f9df9812665f568775e66e2e5a2f3d6ee35e317c11629d52e2891c9bd

a-shannon:
## Summary

- Track the exact header represented by the running extra indexer, not only its height.
- Index an applied block directly only when it is the next height, its `parentId` matches the indexed tip, and it is the current best header at that height.
- Skip duplicate applications without changing actor behavior. If history has switched away from the indexed tip, wait for the normal `Rollback`, then catch up from the branch point.
- Add focused regressions for duplicate delivery and parent continuity across a delayed rollback.

## Why

During a chain switch, replacement `FullBlockApplied` events can reach `ExtraIndexer` before `Rollback`. Height alone cannot distinguish a duplicate from a replacement branch. Indexing a successor whose parent is not the header represented by the extra index can leave numeric input indexes pointing to the wrong boxes.

The previous revision introduced a dedicated waiting behavior. A genuine duplicate could enter that behavior without a later rollback and block subsequent applied blocks. This revision removes that behavior and checks chain identity explicitly.

This affects only the optional extra index and its derived queries. It does not change UTXO validation or consensus state.

## Scope

The change is local to `ExtraIndexer` and its in-memory state. It does not change `NodeViewHolder` event ordering, database schemas, APIs, or consensus rules.

PR #2345 touches the same rollback event for telemetry, but does not address this ordering invariant.

## Tests

- `sbt "testOnly org.ergoplatform.nodeView.history.extra.ExtraIndexerSpecification"`
- 12 tests passed.
- Duplicate `H+1` delivery followed by `H+2` proves liveness and idempotence.
- A successor whose parent is replacement `H`, not the indexed original `H`, is deferred until rollback; all numeric input indexes then resolve to the exact spent boxes.

Closes #2441


src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -491,26 +511,20 @@ trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
       } else if (header.height > state.indexedHeight + 1) { // applied block is ahead of indexer
         context.become(receive.orElse(loaded(state.copy(caughtUp = false))))
         self ! Index()
-      } else // applied block has already been indexed, skipping duplicate
-        log.warn(s"Skipping block ${header.id} applied at height ${header.height}, indexed height is ${state.indexedHeight}")
+      } else { // a chain switch is applied before its rollback event is published
+        log.info(s"Waiting for rollback before indexing block ${header.id} at height ${header.height}")
+        stash()
+        context.become(receive.orElse(waitingForRollback(state)))
```
kushti: why is waitingForRollback handler is needed at all ? what if somehow double application attempt happened like removed handling assumed?  I think it is better to skip waitingForRollback, and verify not just height but also  parentId in "header.height > state.indexedHeight + 1"

kushti (changes requested):

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
a-shannon: Addressed in `1172a8f23`.

I removed `waitingForRollback`. `IndexerState` now tracks the exact indexed header ID. A `FullBlockApplied` event is indexed directly only when it is the next height, its `parentId` matches the indexed tip, and it is still the best header at that height. Exact duplicate delivery is skipped without changing actor behavior. If history has moved away from the indexed tip, applied events are ignored until the normal `Rollback`, then the indexer catches up from the branch point.

I added two focused regressions:

- Deliver `H+1` twice, then `H+2`, and verify final height plus exact transaction and box counts. This covers the duplicate case and proves the actor does not block or double-index.
- Make `H+1.parentId` point to replacement `H`, not the indexed original `H`, and deliver the successor first so the parent check is causal. After the delayed rollback, every numeric input index must resolve to the exact spent box.

The full `ExtraIndexerSpecification` suite passes, 12/12.

kushti (approved):

PR #2448, opened 2026-07-22
Block candidate generation improvements

kushti:
In this PR: 
* transaction ids from the block applied are used in candidate generation to filter out quickly transactions which can be in the mempool yet 
* extra check that blockchain is synced after candidate generation (and discarding candidate if blockchain is not synced, then a new candidate will be generated quickly on new FullBlockApplied signal)

### DIFF
--- a/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
+++ b/src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
@@ -105,7 +105,8 @@ class CandidateGenerator(
             h,
             s,
             m,
-            avgGenTime = 1000.millis
+            avgGenTime = 1000.millis,
+            lastAppliedBlockTxs = None
           )
         )
       )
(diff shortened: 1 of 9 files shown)


odiseusme:
Went through the diff. I did not find a production-code issue in the two mechanisms themselves. The `bestFullBlockIdOpt.contains(appliedHeaderId)` guard in `excludeAppliedTxs` was the detail I checked first; filtering only when the recorded applied block is still the best full block looks right to me, including across a reorg.

A few observations, mostly about test reach, since I just went through a similar issue on #2447:

The new tests exercise the fallback behavior and the helpers, but not the production wiring. The two stale-transaction tests call `collectTxs` directly, which proves the existing invalidation path still works; `excludeAppliedTxs` and `isChainSynced` are tested as pure functions. As written, the `FullBlockApplied(header, txIds)` → actor state → `generateCandidate` → filtered mempool snapshot path, and the post-assembly candidate discard, could both be removed while the added tests remained green.

On #2447, two reviewers caught me on essentially this shape: my test asserted on a returned pool while production discarded it. The useful regression there was an actor-level test through the real consumer. I'm happy to attempt the equivalent here: one test driving a real `FullBlockApplied` while the corresponding IDs are still visible in the mempool reader, then asserting they neither enter the candidate nor appear in `EliminateTransactions`; and one controllable-history test where the best-block id changes during assembly and the completed candidate is discarded.

Two smaller observations:

- The `takeRight(10)` / `takeRight(5)` fixtures do not assert that the selected inputs are non-empty, so an unlucky `boxesHolderGen.sample` could make the empty-vs-empty assertions pass without exercising a transaction.
- On the new post-assembly discard path, the `EliminateTransactions` produced by `createCandidate` is discarded together with the candidate. That may be intentional because the next assembly rediscovers those transactions, but it would be useful to confirm.

One coordination question rather than a finding: this targets `master`, while #2443/#2444/#2447 target `v6.0.4`, and #2443 also touches `CandidateGenerator`. How would you prefer those to be sequenced?

odiseusme:
Correction to my fixture note, via review discussion: `validFullBlock(None, us0, mempoolTxs)` rejects an empty transaction sequence, so the `takeRight(10)` case cannot pass vacuously — only the zero-fee `takeRight(5)` case needs an explicit non-empty assertion.

PR #2449, opened 2026-07-24
Cap on peers database size

kushti:
In this pull request, max cap on peer database size is introduced. Also, when the node is sending known peers list to a peer, it is doing a random slice from all the database and then working with it (shuffling and slicing a smaller set), instead of doing costly shuffling with all the database


odiseusme:
Reviewed the full diff. The direction is right: cap the peer database, validate persisted records, protect connected peers, age out stale entries, and bound work in peer propagation and the /peers/all route. I found several correctness issues that I think should be addressed before this goes into 6.0.5.

**1. Startup trimming neither keeps the newest peers nor bounds startup memory**

`validPeers.splitAt(maxKnownPeers)` runs on the list produced from persistent-store iteration, reversed by prepending, with no ordering by recency. `LoadedPeer.lastHandshake` is captured but never used.

Commit c69c75e sorted by `lastHandshake` descending before the split; d2f2ee4 removed that sort. As a result, on the first 6.0.5 startup with an oversized database, retention is determined by store iteration order rather than recency. Recently handshaked peers can be selected for deletion while stale peers survive.

The test "should load only newest peers when persisted set exceeds cap" does not independently vary address/insertion ordering from timestamp ordering, so it can pass without proving that timestamps drive selection. Please permute the timestamps independently and verify the retained and physically deleted records after reopening.

The same path also calls `persistentStore.getAll.toVector` and then constructs additional collections containing the valid peers and rejected keys. Startup memory therefore remains O(total persisted database size), despite the method comment stating that an oversized or malformed database cannot OOM the node during startup.

Suggest using bounded newest-peer selection while iterating, together with bounded/batched removal of rejected and excess keys.

**2. Connected-peer protection does not reliably cover inbound peers**

`PeerManager` populates `connectedPeerAddresses` with `remote.connectionId.remoteAddress`, while the peer database is keyed by the address derived from `peerInfo.peerSpec.address`.

For an inbound connection, the transport endpoint generally contains the peer's ephemeral source port, which will not equal its advertised listening address. Consequently, the database identity of an actively connected inbound peer may not be present in `connectedPeerAddresses`, allowing it to be selected by either `makeRoomForPeer` or `removeOldPeers`.

The current test covers only an outgoing connection where the two addresses coincide. Suggest tracking the advertised/database address once it becomes available after handshake, potentially alongside the transport endpoint, and adding an inbound test where the socket and advertised ports differ.

**3. Peers with `lastHandshake == 0` are treated as immediately stale**

Peers learned through `AddPeerIfEmpty` are created with `lastHandshake = 0`; configured seed peers also begin without a successful handshake.

Two behaviors follow:

- When the database is full, a zero-handshake candidate cannot displace any existing peer because `makeRoomForPeer` requires `candidateHandshake > oldest.lastHandshake`. This may be an intentional anti-spam policy, but it should be explicit.
- More importantly, `removeOldPeers` removes every peer whose `lastHandshake` is below the 60-day cutoff. Zero therefore qualifies immediately, so the first cleanup after approximately 24 hours removes all untried, disconnected peers regardless of when they were discovered.

Configured peers are seeded only during `PeerManager` construction when the database is empty; they are not automatically restored after cleanup while the process is running.

Suggest either exempting `lastHandshake == 0` from handshake-age cleanup or storing a separate discovery/insertion timestamp and defining an explicit expiry policy for untried peers. Please add tests for both an untried discovered peer and an unavailable configured seed.

**4. Malformed records are skipped but not removed from persistent storage**

Oversized records are added to the removal collection, but the deserialization-failure branch only logs and skips the record. Its key remains in the database.

These records are consequently reparsed on every startup, and the physical store can remain above the nominal cap indefinitely.

Suggest adding all unparseable keys to the removal path and adding a close/reopen test proving that the record was physically deleted.

**5. /peers/all introduces a silent compatibility change but does not enforce a maximum response size**

Previously, GET /peers/all returned the complete known-peer collection. It now silently defaults to 50 records, with no total count or continuation metadata indicating that the result was truncated. That is a semantic change for monitoring and crawler clients, particularly in a patch release.

At the same time, `limit` has no upper bound, so a caller can still request the complete capped database. `offset + limit` can also overflow before being passed to `slice`.

Please make the API contract explicit. Possible approaches are a separate paginated endpoint, a versioned response, or a documented compatibility break with pagination metadata. Whichever contract is selected, enforce an explicit maximum page size and prefer `drop(offset).take(limit)` over `slice(offset, offset + limit)`.

**Non-blocking observation**

`SeenPeers.choose` filters only after selecting its bounded scan window. It can therefore return no peers when that window contains no eligible records, even though eligible peers exist elsewhere in the database. That may be an acceptable bounded-work tradeoff, but it is a behavior change from the previous full scan and should be intentional and tested with sparse eligible peers.

The insertion-time cap, eviction preference for older handshakes, per-record parse isolation, bounded peer-response work, and the added regression coverage are all useful improvements. The issues above are mostly localized, but the startup-selection, connected-peer identity, zero-handshake cleanup, and malformed-record persistence cases should be resolved before release.

PR #2462, opened 2026-08-11
Break fee inputs into chunks in collectRewards (#2185)

### LINKED ISSUE #2185: Put mining fee inputs into multiple transactions when too many of them
In the block candidate generation logic (`collectRewards` function in `CandidateGenerator` class) all the fee inputs are collected with one transaction only just, however, if there are too many inputs, the transaction may be invalid. Break fee inputs up in chunks. Write a test for that.

Ergologica:
Closes #2185.

## Problem

`CandidateGenerator.collectRewards` collects every fee box of a block into one fee transaction:

```scala
val inputs = feeBoxes.map(b => new Input(b.id, ProverResult.empty))
Some(ErgoTransaction(inputs.toIndexedSeq, IndexedSeq(), IndexedSeq(minerBox)))
```

With enough fee-paying transactions in a block, that transaction can grow past the block limits and be invalid, which invalidates the candidate.

## Fix

Fee boxes are grouped into chunks of at most `MaxFeeBoxesPerTransaction` (100), one fee-collecting transaction per chunk. Each chunk gets its own miner box and its own token set truncated at `MaxAssetsPerBox`, as before.

Chunking `collectRewards` alone would have had no effect where it counts: `collectFees` ended in `.headOption`, and the assembly loop in `collectTxs` carried the fee transaction as a single `Option[CostedTransaction]`, so every chunk after the first would have been dropped silently. `collectFees` now returns `Seq[ErgoTransaction]` and the loop threads `Seq[CostedTransaction]`. Each chunk is validated on its own with `statefulValidity`; one failing chunk aborts the step exactly as one failing fee transaction did before, and the cost/size check sees all of them together.

Behaviour is unchanged for any block with 100 fee boxes or fewer, which is every block on mainnet today.

## Tests

- new: `fee boxes are collected in chunks of at most MaxFeeBoxesPerTransaction` - with more fee boxes than one chunk holds, the transaction count is `ceil(n / 100)`, every fee box is spent exactly once across the chunks, and the miner's total is unchanged.
- updated: `collect reward from both emission box and fees` asserted exactly 2 transactions, which stops holding once the generator produces more than 100 fee boxes. It now checks the invariants that survive a split: the emission transaction, per-chunk input bounds, and the total paid to the miner.

`sbt "testOnly org.ergoplatform.mining.CandidateGeneratorPropSpec"` - 9/9 green.

## On the chunk size

100 is a deliberate conservative constant rather than a computed limit. The alternative is to size chunks from the block cost and size parameters in the upcoming context, which would put block-limit arithmetic inside reward collection. Happy to change the number, or to derive it, if you would rather have that.


cafebedouin:
Review carried out by Claude (Anthropic, Claude Opus 5.5) for cafebedouin, using forkbench; executed `CandidateGeneratorPropSpec` with the change, reverted and with chunking disabled, and a two-node payment run of the 6.0.5 merge base and of this branch.

**Chunking is correct and harmless at 100 fee boxes or fewer, and the changed properties guard it in `collectRewards`; the rationale in the code and description does not match what the base code did.**

Executed (sbt, JDK 8): the spec at `80e4663` passes 14/14; with the production file reverted it does not compile (it uses `MaxFeeBoxesPerTransaction`); with only `.grouped(MaxFeeBoxesPerTransaction)` replaced by one group of all fee boxes, 2 of 14 fail (`1 was not equal to 3` at `CandidateGeneratorPropSpec.scala:315`; `191 was not less than or equal to 100`). Payments on two nodes: 10/10 confirmed, 17 transactions in blocks, on base and branch alike. Not run: a block with more than 100 fee boxes on a network; other platforms. A mixed-version mining run was inconclusive in both directions (including with the base jar on both nodes, one run) and is not cited.

**[Maintainability] Correct the rationale at `CandidateGenerator.scala:802-803` and `:919-920`**

Observed: at `5528ef569`, `collectTxs` validated the single fee transaction and stopped at the last accepted step when it failed (`CandidateGenerator.scala:991-1010`), so an oversized fee transaction shortened the block; it did not invalidate the candidate. Its cost cap is `maxBlockCost` (`ErgoTransaction.scala:377-395`), while assembly is bounded by `maxBlockCost - safeGap` (`CandidateGenerator.scala:682-692`), so the block check binds; the per-transaction input bound, `Short.MaxValue` (`ErgoTransaction.scala:95`), is looser. Each extra chunk adds `interpreterInitCost` (10000) and one output.

Wrong if a per-transaction consensus rule tighter than the block limits applies to a fee transaction.

Recommended: say in the scaladoc and description that chunking is defensive, not a validity fix. If a specific limit motivated #2185, naming it would settle the constant. (read)

**[Maintainability] Add a `collectTxs` property past one chunk**

Observed: both changed properties call `collectRewards`; the `Seq` threading through `collectTxs` (`CandidateGenerator.scala:1007-1045`), has no test above 100 fee boxes.

Recommended: one `collectTxs` property with more than 100 fee-paying transactions, asserting every chunk is in the result. (read)

**[Integration] #2504 carries the same batching**

Observed: draft #2504 (on `weak-blocks`) adds the same `MaxFeeBoxesPerTransaction = 100` and says it follows this PR; #2479, #2520 and #2357 rewrite the same `lastFeeTx` loop. It merges cleanly onto `master` (`23aabead8`).

Recommended: keep the constant and its comment identical across the two. (read)

```
revert: CandidateGeneratorPropSpec | with: pass | without: compile error
chunking disabled: Tests: succeeded 12, failed 2
payments, base jar 6.0.5 (sha256 2a7e2978…1588): accepted 10, confirmed 10, txs_in_blocks 17
payments, this branch (6.0.5-1-d10a9b3c, sha256 40071221…3a83): accepted 10, confirmed 10, txs_in_blocks 17
```

</details>

PR #2465, opened 2026-08-12
Make ExtraIndexer checkpoints reorg-safe

a-shannon:
## Summary

- Persist the exact full-chain header represented by each ExtraIndexer checkpoint and rebuild legacy, malformed, or interrupted checkpoints under schema 7.
- Write forward index rows and checkpoint metadata in one LevelDB batch, bind cached transaction sections to their selected header, and serialize extra-cache misses with writes.
- Retry transient catch-up gaps, reconcile restart/reorg ordering against the valid full-block chain, and fail closed when a multi-step rollback cannot be persisted safely.

## Why

#2442 made the running indexer track an exact header, but restart still reconstructed that identity from the current best header at the stored height. During a fork, the persisted rows could represent branch A while the reloaded state claimed branch B. Replacement block events could then be applied over the wrong rows before the delayed rollback notification.

This change persists provenance instead of inferring it, validates checkpoint shape and terminal mappings on startup, and drives recovery from the current valid full-block chain. Missing block sections and transient header/full-block disagreement are deferred and retried rather than being recorded as caught up.

## Validation

- `ExtraIndexerSpecification`: 23/23
- `HistoryStorageSpec`: 4/4
- `LDBKVStoreSpec`: 3/3

The regressions cover restart onto a competing branch, replacement events before rollback, buffered catch-up followed by reorg, missing transaction sections, malformed/interrupted checkpoints, rollback write failure, cache write failures and retries, an in-flight cache-miss/write race, and production actor startup through the event stream.

## Operational notes

- Existing non-empty schema-6 extra indexes rebuild once under schema 7.
- Forward rows and checkpoint metadata are committed atomically. Rollback remains multi-batch: a durable rollback marker is written first, and a later write failure stops only the ExtraIndexer actor. The node keeps running, and startup validation rebuilds the optional index on the next node restart.
- This affects the optional derived extra index only. It does not change consensus validation, UTXO state, P2P serialization, or API schemas.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
@@ -17,7 +17,6 @@ import org.ergoplatform.nodeView.history.storage.HistoryStorage
 import org.ergoplatform.nodeView.history.storage.modifierprocessors.FullBlockProcessor
 import org.ergoplatform.settings.{Algos, CacheSettings, ChainSettings}
 import scorex.util.{ModifierId, ScorexLogging, bytesToId}
-import scorex.db.ByteArrayWrapper
 import sigma.ast.ErgoTree
 import sigma.Extensions._
 import sigma.interpreter.ProverResult
@@ -75,9 +74,7 @@ trait ExtraIndexerBase extends Actor with Stash with Timers with ScorexLogging {
   protected def fullChainHeaderAtHeight(height: Int): Option[Header] = {
     _history.headerIdsAtHeight(height)
       .find { id =>
-        historyStorage.getIndex(FullBlockProcessor.chainStatusKey(id))
-          .map(ByteArrayWrapper.apply)
-          .contains(ByteArrayWrapper(FullBlockProcessor.BestChainMarker)) &&
+        FullBlockProcessor.isInBestFullChain(historyStorage, id) &&
           _history.isSemanticallyValid(id) == ModifierSemanticValidity.Valid
       }
       .flatMap(id => _history.typedModifierById[Header](id))
@@ -145,6 +142,13 @@ trait ExtraIndexerBase extends Actor with Stash with Timers with ScorexLogging {
     * Holds upcoming blocks to be indexed, and when empty, it is filled back from multiple threads
     */
   private val blockCache: concurrent.Map[Int, BlockTransactions] = new ConcurrentHashMap[Int, BlockTransactions]().asScala
+
+  private[extra] final def putBlockTransactionsInCache(
+    height: Int,
+    transactions: BlockTransactions
+  ): Unit =
+    blockCache.put(height, transactions)
+
   private var readingUpTo: Int = 0
 
   /**
@@ -168,7 +172,9 @@ trait ExtraIndexerBase extends Actor with Stash with Timers with ScorexLogging {
           blockNums.zip(blockNums.tail).map { range => // ranges of 250 blocks for each thread to read
             Future {
               (range._1 until range._2).foreach { blockNum =>
-                fullChainHeaderAtHeight(blockNum).flatMap(blockTransactionsForHeader).map(blockCache.put(blockNum, _))
+                fullChainHeaderAtHeight(blockNum)
+                  .flatMap(blockTransactionsForHeader)
+                  .map(putBlockTransactionsInCache(blockNum, _))
               }
             }
           }
@@ -176,7 +182,9 @@ trait ExtraIndexerBase extends Actor with Stash with Timers with ScorexLogging {
           val blockNums = height + 1 to readingUpTo
           Future {
             blockNums.foreach { blockNum =>
-              fullChainHeaderAtHeight(blockNum).flatMap(blockTransactionsForHeader).map(blockCache.put(blockNum, _))
+              fullChainHeaderAtHeight(blockNum)
+                .flatMap(blockTransactionsForHeader)
+                .map(putBlockTransactionsInCache(blockNum, _))
             }
           }
         }
--- a/src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
@@ -1,7 +1,8 @@
 package org.ergoplatform.nodeView.history.extra
 
-import akka.actor.{Actor, ActorRef, ActorSystem, Props, Stash}
-import org.ergoplatform.{ErgoAddress, ErgoAddressEncoder, GlobalConstants, Pay2SAddress}
+import akka.actor.{Actor, ActorRef, ActorSystem, Props, Stash, Timers}
+import org.ergoplatform.{ErgoAddress, ErgoAddressEncoder, ErgoApp, GlobalConstants, Pay2SAddress}
+import org.ergoplatform.consensus.ModifierSemanticValidity
 import org.ergoplatform.modifiers.history.BlockTransactions
 import org.ergoplatform.modifiers.history.header.Header
 import org.ergoplatform.modifiers.mempool.ErgoTransaction
@@ -13,8 +14,10 @@ import org.ergoplatform.nodeView.history.extra.IndexedContractTemplateSerializer
 import org.ergoplatform.nodeView.history.extra.IndexedErgoAddressSerializer.hashErgoTree
 import org.ergoplatform.nodeView.history.extra.IndexedTokenSerializer.uniqueId
 import org.ergoplatform.nodeView.history.storage.HistoryStorage
+import org.ergoplatform.nodeView.history.storage.modifierprocessors.FullBlockProcessor
 import org.ergoplatform.settings.{Algos, CacheSettings, ChainSettings}
 import scorex.util.{ModifierId, ScorexLogging, bytesToId}
+import scorex.db.ByteArrayWrapper
 import sigma.ast.ErgoTree
 import sigma.Extensions._
 import sigma.interpreter.ProverResult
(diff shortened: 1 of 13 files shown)


src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -62,16 +72,60 @@ trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
 
   protected def historyStorage: HistoryStorage = _history.historyStorage
 
+  protected def fullChainHeaderAtHeight(height: Int): Option[Header] = {
```
jozanek, line 74: **MINOR** `fullChainHeaderAtHeight` re-derives `chainStatusKey` + `BestChainMarker` inline even though `FullBlockProcessor.isInBestFullChain` already encapsulates this check, and the `ErgoHistory` startup block similarly hand-rolls the `validity`-key hash that `HeadersProcessor.validityKey` derives. That's three copies of history's storage layout that can silently drift apart if the encoding ever changes. Suggestion: expose static key-derivation/accessor helpers (e.g. lift `isInBestFullChain` so the indexer and the pre-construction startup check can both reuse it).

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -678,13 +827,14 @@ object ExtraIndexer {
   /**
     * Current newest database schema version. Used to force extra database resync.
     */
-  val NewestVersion: Int = 6
+  val NewestVersion: Int = 7
```
jozanek, line 836: **MINOR** Two deliberate operational impacts (both documented in the PR body) should reach the operator-facing release notes: every `extraIndex = true` node performs a one-time full extra-index rebuild on upgrade to schema 7 (hours on mainnet), and a rollback write failure now shuts down the whole node fail-closed, with a rebuild on restart. Explorer/wallet operators will want to plan for both.

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
```diff
@@ -150,16 +165,46 @@ class HistoryStorage(indexStore: LDBKVStore, objectsStore: LDBKVStore, extraStor
 
   def insertExtra(indexesToInsert: Array[(Array[Byte], Array[Byte])],
```
jozanek, line 166: **NIT** The log-and-swallow `insertExtra`/`removeExtra` wrappers now have zero production callers — everything migrated to the `Try` variants. Consider deprecating or removing them so new call sites can't silently lose write errors.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -27,19 +30,26 @@ import java.util.concurrent.ConcurrentHashMap
...
 import scala.concurrent.{ExecutionContextExecutor, Future}
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.jdk.CollectionConverters._
+import scala.util.{Failure, Success, Try}
 
 /**
   * Base trait for extra indexer actor and its test.
   */
-trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
+trait ExtraIndexerBase extends Actor with Stash with Timers with ScorexLogging {
```
jozanek, line 39: **NIT** `ExtraIndexer.scala` is now 853 lines, over the repo's 800-line guideline. The rollback/reconcile logic added here is fairly self-contained — a candidate to split into its own file.

src/test/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexerTestActor.scala
```diff
@@ -98,4 +169,54 @@ class ExtraIndexerTestActor(test: ExtraIndexerSpecification) extends ExtraIndexe
     test.lock.unlock()
   }
 
+  private def cacheBlockTransactions(height: Int, transactions: BlockTransactions): Unit = {
```
jozanek, line 173: **NIT** `cacheBlockTransactions` reaches the private `blockCache` via reflection on the scalac-mangled name (`$$blockCache`) — this breaks on a rename or a Scala-version mangling change. Making the cache `protected` (or adding an explicit test seam like the other overrides in this actor) would be sturdier.

jozanek (commented):
Verified this fixes the catch-up-stall finding from #2293/#2416 — root cause removed (header selection now follows the best full chain), plus generation-guarded retries and self-driven reconciliation, with the stall scenarios regression-tested. Findings: **2 MINOR, 3 NIT** inline; no protocol/consensus concerns.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
a-shannon: Addressed the best-full-chain half in 71630c375: FullBlockProcessor now owns a package-scoped storage helper, the existing instance method delegates to it, and ExtraIndexer reuses it. I am leaving the pre-construction validity-key lift for a focused cleanup, since it requires changing the HeadersProcessor companion surface rather than this indexer fix.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
a-shannon: Agreed. @kushti_ru, please carry these two points into the v6.0.4 release notes: existing non-empty schema-6 extra indexes are reset and rebuilt once on schema 7; and a rollback persistence failure now shuts the node down fail-closed, with startup validation/rebuild on restart. Rebuild duration depends on database size and hardware; this PR does not benchmark it.

src/main/scala/org/ergoplatform/nodeView/history/storage/HistoryStorage.scala
a-shannon: Confirmed: there are no in-tree callers left. I am retaining the public wrappers in this maintenance PR to avoid an unannounced source/binary compatibility change. Deprecation can be handled in a cleanup follow-up, with removal reserved for an appropriate breaking release.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
a-shannon: Agreed on the guideline breach. I am deferring the split to a dedicated cleanup PR: rollback/reconciliation is coupled to actor state, timers, stash, and self-messages, so extracting it here would widen access boundaries and the review surface of the correctness fix.

src/test/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexerTestActor.scala
a-shannon: Fixed in 71630c375. The cache remains private; a private[extra] final write seam is used by production prefetch and the same-package test actor, so the reflection and scalac-mangled-name dependency are gone.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -62,16 +71,58 @@ trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
 
   protected def historyStorage: HistoryStorage = _history.historyStorage
 
+  protected def fullChainHeaderAtHeight(height: Int): Option[Header] = {
+    _history.headerIdsAtHeight(height)
+      .find { id =>
+        FullBlockProcessor.isInBestFullChain(historyStorage, id) &&
+          _history.isSemanticallyValid(id) == ModifierSemanticValidity.Valid
```
kushti, line 78: the second check looks excessive (if block section with identifier `id` in the best chain , it is semantically valid)

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -460,102 +525,194 @@ trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
...
+      removeAfter(state, targetHeader) match {
+        case Success(newState) =>
+          context.become(receive.orElse(loaded(newState)))
+          if (resume && !newState.caughtUp) self ! Index()
+          caughtUpHook()
+          log.info(s"Successfully rolled back indexes to ${targetHeader.height}")
+          unstashAll()
+        case Failure(error) =>
+          log.error(s"Failed to roll back extra indexes to ${targetHeader.height}; shutting down so startup can rebuild", error)
+          requestShutdown()
```
kushti: EXtra indexer should not kill the node in any case

kushti (changes requested):

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
a-shannon: I traced this before removing it and found a pre-validation window. VerifyADHistorySpecification shows that bestFullBlockOpt can already select a newly stored block while its header and sections are still Unknown, before reportModifierIsValid. typedModifierById filters only Invalid, and ExtraIndexer catch-up can run from startup/retry independently of FullBlockApplied. I therefore kept the explicit Valid check to avoid indexing an Unknown best-chain block.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
a-shannon: Fixed in 98547b569. A rollback persistence failure now stops only the ExtraIndexer actor via context.stop(self); it no longer shuts down the ActorSystem or node. The durable rollback marker remains for startup rebuild. The regression watches a dedicated actor, proves that actor terminates, proves the ActorSystem stays alive, and preserves the interrupted-marker assertion. Focused suite: 23/23 green.

src/main/scala/org/ergoplatform/nodeView/history/extra/ExtraIndexer.scala
```diff
@@ -62,16 +71,56 @@ trait ExtraIndexerBase extends Actor with Stash with ScorexLogging {
...
+    _history.headerIdsAtHeight(height)
+      .find { id =>
+        FullBlockProcessor.isInBestFullChain(historyStorage, id) &&
+          _history.isSemanticallyValid(id) == ModifierSemanticValidity.Valid
+      }
+      .flatMap(id => _history.typedModifierById[Header](id))
+  }
+
+  protected def blockTransactionsForHeader(header: Header): Option[BlockTransactions] = {
+    history.typedModifierById[BlockTransactions](header.transactionsId).filter(_.headerId == header.id)
```
kushti: .filter is excessive

a-shannon:
Updated in a40c2064. I removed the extra `.filter`: `header.transactionsId` already identifies the corresponding transaction section, so the typed lookup is sufficient. I kept the separate cache/header check because a cached entry may belong to a different branch. `ExtraIndexerSpecification` remains green (23/23).

Issue #1180, opened 2020-08-25
Rework offchain registry


kushti:
Currently offchain registry being refreshed only when a transaction is entering the memory pool or when a new block arrives.

It is needed to refresh it also when a transaction is leaving the pool, and when rollback happens.

kushti:
Duplicate of #1228

PR #2466, opened 2026-08-13
Persist unconfirmed wallet transactions and restore them at start-up

### LINKED ISSUE #1154: Unconfirmed wallet transactions during node restart
I think there's a problem that can occur when a node is restarted, which is that the mempool is empty after restart and doesn't synchronise with other nodes very quickly.

The main issue here is that unconfirmed wallet transactions are not persisted (AFAIK), since all unconfirmed wallet transactions are sent directly to the mempool.  After restart, it then becomes easy for the wallet to create a new tx which double-spends any outputs which were previously spent before the restart, but which have not been confirmed yet.

I think the best solution would be to persist unconfirmed wallet transactions, and add them to the mempool on startup.

Ergologica:
Closes #1154.

## The problem

The memory pool is not persisted: `ErgoNodeViewHolder` builds it with `ErgoMemPool.empty(settings)` on both the genesis and the restore path, so a restart drops every transaction waiting in it. Peers may re-gossip them, but nothing guarantees that they will, or when.

What turns this from a nuisance into a correctness problem is the wallet side. `ErgoWalletState.walletFilter` decides whether an on-chain box may be spent by asking the off-chain registry:

```scala
val preStatus = if (trackedBox.chainStatus.onChain) {
  offChainRegistry.onChainBalances.exists(_.id == trackedBox.boxId)
} else { true }
```

and `OffChainRegistry.init` rebuilds `onChainBalances` from `walletRegistry.unspentBoxes(PaymentsScanId)` with `offChainBoxes = ArraySeq.empty`. The wallet registry still lists those boxes as unspent - the spending transaction never got onto the blockchain - so after a restart every box spent by a still-unconfirmed transaction is offered for spending again, and the wallet will build a conflicting transaction spending it. The memory pool cross-check in the same filter cannot save it, because the pool is empty too.

## The change

Wallet-related unconfirmed transactions are kept in `WalletStorage` and replayed at start-up.

**Where they are stored.** `WalletStorage`, not `WalletRegistry`: it is documented as holding *"version-agnostic wallet actor's mutable state (which is not a subject to rollbacks in case of forks)"*, which is what off-chain data is. New secondary prefix byte `3`, key `[0, 3] ++ txId`, value `height ++ txBytes`.

**Which ones.** `scanOffChainUpdate` now also reports whether the transaction is of any interest to the wallet: it either pays the wallet (`extractWalletOutputs` found something) or spends one of its boxes (`registry.getBox` on the inputs). Nothing else is stored, so the store does not turn into a copy of the memory pool.

**Start-up, wallet side.** `ErgoWalletService.restoreOffChainState` replays the stored transactions through the same off-chain scan, rebuilding `offChainRegistry` so the spent boxes stay spent. `updateOnTransaction` is idempotent, so a replayed transaction being scanned again later does no harm.

**Start-up, memory pool side.** `ErgoNodeViewHolder.preStart` asks the wallet for them and feeds them back through `txModify` as a new `RestoredTransaction` message. Accepted ones land in the pool and, via `SuccessfulTransaction`, get gossiped again. Anything the pool refuses - it got onto the blockchain while the node was down, or a conflicting transaction did - is dropped from the store, so the wallet self-heals instead of retrying forever.

**Ordering.** Unconfirmed transactions form chains, and both consumers need the parent first: the off-chain registry only nets a spending out if it has already seen the box being spent, and the pool refuses a transaction whose inputs it does not know. `ErgoWalletActor.orderByDependency` does that.

**Bounding the store.** Entries go away when their transaction appears in an applied block, and are given up on after `UnconfirmedTxLifetimeInBlocks = 1440` blocks (about two days). Re-storing a transaction keeps the height it was first seen at - otherwise the replay at every restart would push the expiry back forever. I kept the lifetime a constant rather than a config key to keep the diff small; happy to make it a setting if you prefer.

## Tests

- `WalletStorageSpec` - round trip through the database, the new bucket not colliding with the public keys and scans buckets, re-storing keeping the original height, forgetting an unknown id being a no-op.
- `ErgoWalletServiceSpec` - a box spent off-chain is filtered out; **after a simulated restart it is spendable again, and `restoreOffChainState` makes it unspendable** (that property is this issue, start to finish); chained transactions replayed parent first; `orderByDependency` leaving independent transactions alone.
- `ErgoWalletSpec` - end to end through the actor: a generated payment is kept after `scanOffchain`, comes back from `unconfirmedTransactionsToRestore`, and is forgotten once the block carrying it is applied.

Full suite green (`sbt test`, 751 tests, 0 failed).

## Notes for review

- `ErgoNodeViewHolder.preStart` uses an `ask` on the wallet actor. The wallet stashes messages until `ReadWallet` completes, so the reply arrives once it is loaded; a failure is logged and start-up carries on.
- Nothing changes for a node whose wallet has no unconfirmed transactions: `readUnconfirmedTransactions()` comes back empty and `restoreOffChainState` returns the state untouched.
- The rollback path still has the pre-existing gap of #1180 (the off-chain registry is not refreshed on rollback). This change does not make it worse, and once unconfirmed transactions are kept, refreshing on rollback could reuse the same replay.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -29,6 +29,7 @@ import org.ergoplatform.modifiers.history.extension.Extension
 
 import scala.annotation.tailrec
 import scala.collection.mutable
+import scala.concurrent.ExecutionContext
 import scala.util.{Failure, Success, Try}
 
 /**
@@ -87,10 +88,33 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
         Escalate
     }
 
+  override def preStart(): Unit = {
+    super.preStart()
+    restoreUnconfirmedWalletTransactions()
+  }
+
   override def postStop(): Unit = {
     log.warn("Stopping ErgoNodeViewHolder")
     history().closeStorage()
     minimalState().closeStorage()
+  }
+
+  /**
+    * The memory pool is not persisted, so a restart drops every transaction which was waiting in it.
+    * The wallet does keep its own unconfirmed transactions, so ask it for them and put them back,
+    * instead of waiting for peers to gossip them again - which they may never do.
+    */
+  private def restoreUnconfirmedWalletTransactions(): Unit = {
+    implicit val ec: ExecutionContext = context.dispatcher
+    vault().unconfirmedTransactionsToRestore.onComplete {
+      case Success(txs) if txs.nonEmpty =>
+        log.info(s"Putting ${txs.size} unconfirmed wallet transaction(s) back into the memory pool")
+        txs.foreach(tx => self ! RestoredTransaction(UnconfirmedTransaction(tx, None)))
+      case Success(_) =>
+        log.debug("No unconfirmed wallet transactions to restore")
+      case Failure(t) =>
+        log.warn("Could not read unconfirmed wallet transactions to restore: ", t)
+    }
   }
 
   /**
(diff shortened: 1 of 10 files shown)


Ergologica:
On the two red checks - neither is caused by this change, and I checked rather than assumed.

**`Run it node tests`** fails on `master` itself: the CI run for commit 4a7dba0, which is the base of this branch, is red on exactly that job, and so is every other open pull request I looked at (e.g. #2462). The cause is `UtxoStateNodesSyncSpec`, which #2452 fixes.

**`Run node tests`** failed on `CandidateGeneratorSpec`, "accept solution for previous candidate after regeneration" (line 318, the reward-script assertion). That suite is flaky on master, independently of this branch. I stashed every change here, went back to a pristine 4a7dba0 and ran the suite twice:

- run 1: `should ignore cached candidate when forced = true` *** FAILED ***
- run 2: `should ignore cached candidate when forced = true` and `should preserve previous candidate when forced regeneration occurs` *** FAILED ***

A different test each time, none of them the one CI hit. The suite runs with `blockCandidateGenerationInterval = 1.millis` and `blockInterval = 1.seconds` and fishes for messages, so it loses races on a loaded machine; the failure mode is the test reading `bestFullBlockOpt` before the block it mined has been applied, which is why the assertion sees a plain P2PK script instead of the delayed miner reward script.

For the record, the full suite on this branch was green when I ran it before opening the PR (`sbt test`, 751 tests, 0 failed), and every wallet suite passed in the CI run above - the only failing suite was the mining one.

Happy to open a separate PR to make `CandidateGeneratorSpec` deterministic if that is useful; it seems worth doing regardless of this change.

PR #2467, opened 2026-08-13
Sign and verify arbitrary messages with wallet keys

### LINKED ISSUE #1392: Sign a custom message [Feature request]

Ergologica:
Closes #1392.

Two endpoints, so that the holder of an address can be proven without moving funds:

- `POST /wallet/signMessage` — needs an unlocked wallet, optionally takes an address of that wallet. Returns the address, the byte string actually signed, and the signature.
- `POST /utils/verifyMessage` — needs neither wallet nor blockchain state, so anyone can check a signature. Returns whether it is valid and the message it attests to.

## Schnorr, not ECDSA

The issue asks for ECDSA. I went with the sigma protocol the node already uses, and would rather argue the point than quietly do the other thing.

`ProverInterpreter.signMessage` and `Interpreter.verifySignature` are already in sigma-state, which the node depends on, and they are the same primitive as a transaction input proof. Adding ECDSA would put a second signature scheme in the node for one feature, and nothing else in the ecosystem - sigma-rust, appkit, wallets - would be able to check the result. On the "k" reuse the issue worries about, and RFC 6979: each sigma proof draws a fresh commitment from a secure source, so the failure mode does not arise.

## What is signed, and why it is not the message

This is the part most worth a second pair of eyes.

A sigma proof over an arbitrary byte string **is** the proof which spends a box - both are a Fiat-Shamir transcript over some bytes. If the node signed the caller's bytes verbatim, a caller could hand it the `messageToSign` of a transaction spending the wallet's own boxes and receive a proof that makes that transaction valid. So the node signs

```
"Ergo signed message:\n" || salt || message
```

with 32 fresh random bytes of salt, and returns the byte string it signed. The prefix says what the transcript is for; the salt makes the signed string unpredictable to whoever supplied the message, so it cannot be steered onto a chosen byte string. This is what EIP-0028 (ErgoAuth) already prescribes for wallet applications - the wallet adds its own bytes and reports back what it signed - so a node doing the same keeps the two consistent.

Verification insists on the wrapping too. Accepting an unwrapped byte string would give the separation away, since a transaction input proof would then pass as a message signature.

The convention lives in `ergo-wallet` (`org.ergoplatform.wallet.crypto.MessageSigning`) rather than in the node's API layer, because signer and verifier have to agree on it and `ergo-wallet` is the published library other tools use.

## Tests

- `ErgoWalletServiceSpec` — a signature verifies for its address and for nothing else (other key, altered byte string, other proof all rejected); signing twice gives two different byte strings, both valid; signing needs an unlocked wallet holding the address asked for.
- `ErgoWalletServiceSpec` — **the attack, defeated**: a real transaction is built and signed, its `messageToSign` is then handed to `signMessage` dressed up as a message, and the resulting proof does not verify against the transaction. The other direction is covered too: the transaction's own input proof does not pass as a message signature.
- `UtilsApiRouteSpec` — the verify route: valid, wrong address, unwrapped byte string, and a request which does not parse.
- `WalletApiRouteSpec` — the sign route: default address, chosen address, and a non-P2PK address refused. The stub wallet actor signs for real, so the route test checks a genuine signature.

## Notes

- The endpoint placement is a proposal, not a conviction — happy to move verification to `/script` or elsewhere if you prefer.
- `signMessage` takes text and encodes it as UTF-8, matching `/utils/hash/blake2b` which also takes a plain string. If a hex-bytes variant is wanted for binary payloads it is a small addition.
- No consensus code is touched. The new file is in `ergo-wallet`, which cross-compiles on 2.11, 2.12 and 2.13; nothing there uses anything newer.


Ergologica:
Full suite on this branch: `sbt test`, **753 tests, 0 failed**.

Two notes on CI, in case it comes up red here as it did on my other PR:

- `Run it node tests` fails on `master` itself - the CI run for 4a7dba0, which is the base of this branch - and on every other open pull request. #2452 fixes the cause, `UtxoStateNodesSyncSpec`.
- `CandidateGeneratorSpec` is flaky regardless of any change. On a pristine 4a7dba0, with everything stashed, it failed two runs out of two with a different test each time. It happened to pass in the full run above.

src/main/scala/org/ergoplatform/http/api/WalletApiRoute.scala
```diff
@@ -486,6 +487,42 @@ case class WalletApiRoute(readersHolder: ActorRef,
...
+              case Success(other) => reject(ValidationRejection(s"Not a P2PK address: $other"))
+              case Failure(t) => reject(ValidationRejection(s"Can not parse the address: ${t.getMessage}"))
+            }
+          case Left(t) => reject(ValidationRejection(s"Can not read the address: ${t.getMessage}"))
+        }
+    }
+  }
+
+  def signMessageR: Route = (path("signMessage") & post & messageSigningRequest) { case (message, addressOpt) =>
+    withWalletOp(_.signMessage(message.getBytes(org.ergoplatform.settings.Constants.StringEncoding), addressOpt)) {
```
jozanek: Inlining the fully-qualified `org.ergoplatform.settings.Constants.StringEncoding` here reads awkwardly — it's forced by the existing `org.ergoplatform.wallet.Constants` import, so an alias would clean it up:

```scala
import org.ergoplatform.settings.{Constants => NodeConstants}
```

Also, a few of the new lines here (and in `ErgoUtilsApiRoute.verifyMessageR`) run past scalafmt's `maxColumn = 90`. Not CI-enforced and the surrounding code already violates it, but since the hunks are new, worth a formatting pass.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/crypto/MessageSigning.scala
```diff
@@ -0,0 +1,79 @@
...
+  /** Fresh salt for one signature */
+  def freshSalt(): Array[Byte] = {
+    val salt = new Array[Byte](SaltLength)
+    secureRandom.nextBytes(salt)
+    salt
+  }
+
+  /** The byte string actually signed when `message` is signed with `salt` */
+  def wrap(message: Array[Byte], salt: Array[Byte]): Array[Byte] = {
+    require(salt.length == SaltLength, s"Salt must be $SaltLength bytes long, got ${salt.length}")
```
jozanek: `wrap` throws via `require` from published library code, which goes against the repo's "avoid throwing in library code" guideline. It's a programmer-error precondition (the node always passes `freshSalt()`), so I don't think it needs to become a `Try` — but since ergo-wallet is consumed externally, an explicit `@throws[IllegalArgumentException]` scaladoc note would make the contract visible to library users.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletService.scala
```diff
@@ -655,6 +672,32 @@ class ErgoWalletServiceImpl(override val ergoSettings: ErgoSettings) extends Erg
...
+            val signedMessage = MessageSigning.wrap(message, MessageSigning.freshSalt())
+            prover.signMessage(pubKey, signedMessage, HintsBag.empty).map { proof =>
+              SignedMessage(P2PKAddress(pubKey)(ergoSettings.addressEncoder), signedMessage, proof)
+            }
+          case None =>
+            val what = addressOpt.map(a => s"address $a").getOrElse("any address")
+            Failure(new Exception(s"The wallet has no secret for $what"))
+        }
+      case None =>
+        Failure(new Exception("Wallet is locked"))
```
jozanek: This branch reports "Wallet is locked", but `proverOpt` is also `None` when the wallet was never initialized, so an uninitialized-wallet user gets a misleading 400. Suggest: `"Wallet is locked or not initialized"`.

src/main/resources/api/openapi.yaml
```diff
@@ -596,6 +596,78 @@ components:
...
+        signedMessage:
+          description:
+            Base16-encoded byte string actually signed, that is the message with the prefix and the salt
+            prepended. Pass it verbatim to /utils/verifyMessage.
+          type: string
+          example: '4572676f207369676e6564206d6573736167653a0a'
+        proof:
+          description: Base16-encoded signature
+          type: string
+          example: 'cd07f00d80cd0c9a0d16f9dbba1b4ff5d0e0aa5cb2b1e0ff1e6c0b3f9e4a1a4c'
```
jozanek: The `proof` example is 32 bytes of hex, but a real single-`ProveDlog` signature is 56 bytes. Client authors sometimes size fields from examples — a realistic 56-byte example (e.g. one generated by an actual `signMessage` call) would prevent that.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
```diff
@@ -132,6 +134,37 @@ class ErgoUtilsApiRoute(val ergoSettings: ErgoSettings)(
...
+        case other => Left(s"address: not a P2PK address, $other")
+      }
+    } yield (p2pk, signedMessage, proof)
+
+    parsed.fold(
+      e => BadRequest(e),
+      { case (p2pk, signedMessage, proof) =>
+        val isValid = MessageSigning.verify(p2pk.pubkey, signedMessage, proof)
+        val message = MessageSigning.unwrap(signedMessage)
+          .map(bytes => new String(bytes, Constants.StringEncoding).asJson)
```
jozanek, line 171: `unwrap`-ed payload bytes are decoded as UTF-8 unconditionally here. For node-produced signatures that's always correct, but a signature produced by an external tool over non-UTF-8 payload bytes would return `isValid: true` with replacement characters in `message`. Harmless, just worth a sentence in the OpenAPI description: `message` is the UTF-8 decoding of the signed payload, and may be lossy if the payload wasn't UTF-8 text.

jozanek (commented):
A few minor polish items — nothing blocking. Five inline comments below, plus one that spans two test files:

**Route-test gaps (one-liners each):**
- `WalletApiRouteSpec`: no test for a request missing the `message` field (the `messageSigningRequest` rejection path);
- `UtilsApiRouteSpec`: no test for a non-P2PK address on the verify side (the `Left("address: not a P2PK address, ...")` branch is untested).

src/main/scala/org/ergoplatform/http/api/WalletApiRoute.scala
Ergologica: Done in 82cce94. I split it into two import lines rather than one, because the single-line form is itself 92 columns:

```scala
import org.ergoplatform.settings.{Constants => NodeConstants}
import org.ergoplatform.settings.{ErgoSettings, RESTApiSettings}
```

Formatting pass done on this hunk and on `verifyMessageR`. In `verifyMessageR` the shortest way under 90 turned out to be pulling the two repeated `.left.map(...)` shapes into local `field` / `base16` helpers rather than rewrapping each line.

Code lines in the new hunks are now all under 90. A few scaladoc prose lines sit at 91-93: scalafmt 2.3.2 does not reflow comments, and there is no scalafmt plugin in the build so there is no `scalafmtCheck` task to satisfy either. Happy to hard-wrap those too if you would rather have them uniform.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/crypto/MessageSigning.scala
Ergologica: Done in c08ef13 — kept as a `require`, since it is a programmer-error precondition as you say, and made the contract visible:

```scala
/**
  * ...
  * The salt is a precondition rather than something to recover from: callers are expected
  * to get it from [[freshSalt]], and a salt of the wrong length would silently change what
  * [[unwrap]] reads back.
  *
  * @throws IllegalArgumentException if `salt` is not [[SaltLength]] bytes long
  */
@throws[IllegalArgumentException]
def wrap(message: Array[Byte], salt: Array[Byte]): Array[Byte] = {
```

While in the file I also rewrapped it to 90 columns, since all of it is new.

src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletService.scala
Ergologica: Good catch, done in 7554d52:

```scala
case None =>
  // no prover means either a locked wallet or one which was never initialized
  Failure(new Exception("Wallet is locked or not initialized"))
```

src/main/resources/api/openapi.yaml
Ergologica: Done in 472c524. The `signedMessage` example above it was wrong in the same way — it was only the 21-byte prefix, with neither salt nor payload — so rather than hand-write either, I took a matched pair from an actual `signMessage` call:

- `signedMessage` is 83 bytes: `4572676f207369676e6564206d6573736167653a0a` ("Ergo signed message:\n") ++ 32 bytes of salt ++ the 30 bytes of the example message
- `proof` is the 56-byte signature over exactly those bytes

So the two examples round-trip through `/utils/verifyMessage` instead of just being the right length. I put the same pair on `MessageVerificationRequest`, which had no examples at all, and noted the 56 bytes in the `proof` description.

src/main/scala/org/ergoplatform/http/api/ErgoUtilsApiRoute.scala
Ergologica: Agreed, documented in 472c524. The `message` description now reads:

> The message the signed byte string carries, or null when the byte string is not wrapped for message signing at all. It is the UTF-8 decoding of the signed payload, so a payload which is not UTF-8 text - which a signer following this convention does not produce, but an external one might - comes back lossily decoded. Check `signedMessage` itself if the exact bytes matter.

I left a two-line comment at the decoding site too, so the next reader does not have to go to the OpenAPI file to find out why it is unconditional.

`isValid` is unaffected either way: it is computed over the raw bytes, not over the decoded string.

Ergologica:
Thanks — all six addressed, replies inline on each thread. The two route-test gaps are covered in b0464e3:

- **`WalletApiRouteSpec`**, "should refuse to sign a request which carries no message": asserts the exact `ValidationRejection("A message to sign is required")`, and separately that the sealed route answers 400. My first attempt asserted on the sealed response body instead and failed — `FailFastCirceSupport` unmarshals `responseAs[String]` as JSON while a rejection body is `text/plain` — and pinning the rejection is the better assertion anyway.
- **`UtilsApiRouteSpec`**, "should refuse to verify a message against an address which is not a P2PK one": asserts the 400 carries `not a P2PK address`, so it cannot pass on some unrelated parse failure.

I checked both actually bite rather than trusting that they do: making the missing-message case `provide` instead of `reject`, and changing the non-P2PK error string, each fails its own test and nothing else.

`WalletApiRouteSpec`, `UtilsApiRouteSpec` and `ErgoWalletServiceSpec` are green — 63 tests, 0 failures — and `compile` is clean with `-Xfatal-warnings` on.

cafebedouin:
Review carried out by Claude (Anthropic, Claude Opus 5.5) for cafebedouin, using forkbench; executed the PR's wallet and route specs unchanged and with each guard in `MessageSigning` removed in turn.

**The separation between message signatures and transaction proofs holds: each of the three guards `MessageSigning` names has a test that fails without it. No blocking issue.**

Executed: `sbt testOnly ErgoWalletServiceSpec UtilsApiRouteSpec` on JDK 8 at `e4a71c26`, unchanged and with each guard in `MessageSigning.scala` removed in turn. Unchanged: 32 of 32 pass. Signing the message verbatim (`:62` → `message`): 3 fail. At `ErgoWalletServiceSpec.scala:596` the wallet's message signature then verifies over the transaction's `messageToSign`, so the case the PR guards against is real. `verify` without the wrapping check (`:89`): 2 fail, including `:602`, where the transaction's own input proof passes as a message signature. A constant salt (`:45`): 1 fails, at `:553`. Not run: `src/it`, other JDKs or Scala versions, a live call to either endpoint.

**[Maintainability] Add one test that reaches `SignMessage` through the real wallet actor**

Observed: the route specs get their `SignMessage` answer from the test stub (`Stubs.scala:200-209`), and the service spec calls `ErgoWalletServiceImpl.signMessage` directly. With `ErgoWalletActor.scala:287-288` deleted, `WalletApiRouteSpec`, `UtilsApiRouteSpec` and `ErgoWalletServiceSpec` still pass 65 of 65.

The wiring is correct today; losing that case would go unnoticed. A search of `src/test` and `src/it` for `SignMessage`/`signMessage` finds no other test on that path.

Recommended: call `signMessage` through `ErgoWalletReader` on a real wallet, as `WalletTestOps` does for `publicKeys`, and check the result with `MessageSigning.verify`. (executed)

**[Integration] Link the other open pull requests for #1392**

Observed: #2286 and #1484 are open pull requests for #1392; this thread mentions neither.

Recommended: cross-link #2286 and #1484 here, so that whichever is not taken can be closed. (read)

**[Observability] State in the schema that `message` is returned even when `isValid` is false**

Observed: `ErgoUtilsApiRoute.scala:170-173` sets `message` from `unwrap(signedMessage)` whether or not the proof verified. `openapi.yaml:665-671` gives only "not wrapped" as the null case.

A client that displays `message` without checking `isValid` shows text that no key signed. This is intended if the field exists to debug failed checks.

Recommended: add that sentence to the `message` description, or return `null` unless `isValid` is true. (read)

```
== unchanged: sbt exit 0      Tests: succeeded 32, failed 0
== wrap signs verbatim (MessageSigning.scala:62 -> `message`): sbt exit 1
   - a signed message verifies for the address it was signed with, and for nothing else *** FAILED ***   (ErgoWalletServiceSpec.scala:547)
   - a message signature can not be turned into a transaction proof *** FAILED ***   true was not equal to false (ErgoWalletServiceSpec.scala:596)
   - should verify a signed message *** FAILED ***   Right(false) did not equal Right(true) (UtilsApiRouteSpec.scala:179)
== verify skips the wrapping check (MessageSigning.scala:89 -> `true &&`): sbt exit 1
   - a message signature can not be turned into a transaction proof *** FAILED ***   true was not equal to false (ErgoWalletServiceSpec.scala:602)
   - should not verify a message which was not wrapped for signing *** FAILED ***   Right(true) did not equal Right(false) (UtilsApiRouteSpec.scala:207)
== constant salt (MessageSigning.scala:45, salt left all zero): sbt exit 1
   - a signed message verifies for the address it was signed with, and for nothing else *** FAILED ***   (ErgoWalletServiceSpec.scala:553)
== wallet actor case removed (ErgoWalletActor.scala:287-288), + WalletApiRouteSpec: sbt exit 0
   Tests: succeeded 65, failed 0
```

</details>

PR #2469, opened 2026-08-14
Fix transaction proofs with duplicate witness ids

### LINKED ISSUE #2463: Incorrect txProof when providing a prioritized transaction
# `/mining/candidateWithTxsAndPk`: `txProofs` names a mempool transaction instead of the supplied one

**Node version: v6.0.3RC1. Network: testnet.**

## Summary

We POST **one** transaction to `/mining/candidateWithTxsAndPk`. The node's own log confirms that
transaction **was included** in the resulting candidate. The `proof.txProofs` returned with that same
candidate contains **one** leaf, and it is a **different** transaction — one submitted to the mempool
via `/transactions` about three seconds earlier.

Reproduced four times in the same second, on four separate requests differing in **both** the supplied
transaction and the requested `pk`. All four responses carried the **identical** leaf.

## Why it matters

`proof.txProofs` is the only mechanism an external miner has to verify that its mandatory transaction
reached the candidate. Here it reports the opposite of what happened: the transaction was included,
and the proof says it was not.

A miner that checks the proof — as it must, if the candidate is only valuable when its transaction is
present — will discard candidates it should mine. In our case the transaction pays a third party's
key, so mining without it is a loss; we therefore reject on a proof miss and lose the block.

## Evidence

The two logs can be tied together by the candidate `msg`, which appears in both.

### Node — the supplied transaction is included

```
13:08:05.849  ErgoMemPool - Processing mempool transaction: ErgoTransaction(id: 709306ae887fafa5f98f320a5e9910b18ff30fdd9e43f0a080f377c7ed7d9ceb, …
                                                                                ^ arrives via /transactions

13:08:09.491  CandidateGenerator$ - Assembling a block candidate for block #480553 from 4 transactions available
13:08:09.492  CandidateGenerator$ - No fee proposition found in txs List(9259a02f4ffc41fe86045c00867ab94ad500740ab43c40aa39ecd9db29c9ec7d)
13:08:09.493  CandidateGenerator$ - No fee proposition found in txs List(9259a02f4ffc41fe86045c00867ab94ad500740ab43c40aa39ecd9db29c9ec7d,
                                                                        fea3d4f57b42073799fa59dfc648e337f23865ec836700e1c2523aee2abbbce3)
                                                                        ^ the transaction we POSTed
13:08:09.501  CandidateGenerator$ - Got candidate block at height 480553 with 5 transactions,
                                    msg 0ec47685c073266ce3469ce893d9d1867d7df36b8ce89e82e6f11fcbcc7209e8
13:08:09.501  CandidateGenerator  - Generated new candidate in 12 ms
```

The same pattern repeats three more times, once per request:

| Request | Supplied tx (in the assembly log) | Candidate `msg` |
|---|---|---|
| 1 | `fea3d4f5…` | `0ec47685c073266c…` |
| 2 | `69bc4547…` | `58715d2ae6f0cdca…` |
| 3 | `bf5908ec…` | `2dd41be3a293acef…` |
| 4 | `3f3b0587…` | `ee0b3fb5ca3b2537…` |

### Client — the proof returned with those candidates

```
13:08:09  Requested pk 032306cb…, got a candidate for pk 032306cb…, which match.
          Supplied tx: fea3d4f57b42073799fa59dfc648e337f23865ec836700e1c2523aee2abbbce3
          Raw txProofs: [{"leaf":"709306ae887fafa5f98f320a5e9910b18ff30fdd9e43f0a080f377c7ed7d9ceb",
          "levels":["01687ba6ae3d4d2812ae48c6cc62687b158b49684a886a9505e65d04c393047a17",
                    "01093473a1f2ef32b41d763ae0298e84df5afd1332e3b3d5374be2e27c783a01f9",
                    "0032ab6ca1c8a52af6bf30ad3ef976306c3084b1dfdd08342205741a26ddc697f8",
                    "0013986b8365e33ca188bc0c7c05d0f39327949528cb3404ca347fdd05f7d259f8"]}]

13:08:09  Supplied tx: 69bc454758ffefd09445b06c5f29dbec0644e5a2fb702475bf972deed3b520c3
          Raw txProofs: [{"leaf":"709306ae887fafa5f98f320a5e9910b18ff30fdd9e43f0a080f377c7ed7d9ceb",
          "levels":["01687ba6ae3d…","012dabe188ca…","0022b59f3d03…","0013986b8365…"]}]

13:08:09  Supplied tx: bf5908ec87f06d4f82f29f922248ed7514beec553f6a732bbab4758fa1f6c8be
          Raw txProofs: [{"leaf":"709306ae887fafa5f98f320a5e9910b18ff30fdd9e43f0a080f377c7ed7d9ceb",
          "levels":["01687ba6ae3d…","01c0a21b173d…","005d5228cbd5…","0013986b8365…"]}]

13:08:09  Supplied tx: 3f3b0587e6202e3416c2e12ee02397e16f6ff0d9cb1b8be1d0305efc7e590cc2
          Raw txProofs: [{"leaf":"709306ae887fafa5f98f320a5e9910b18ff30fdd9e43f0a080f377c7ed7d9ceb",
          "levels":["01687ba6ae3d…","01479f9e63e5…","00f02fdcff19…","0013986b8365…"]}]
```

The client then broadcast jobs carrying `msg` values `0ec47685c073266c`, `58715d2ae6f0cdca`,
`2dd41be3a293acef`, `ee0b3fb5ca3b2537` — matching the four candidates in the node log above. The
proofs quoted are the ones attached to those candidates.

### Points that narrow it

1. **The supplied transaction is in the block.** The assembly log lists it; the candidate has 5
   transactions where the previous internal candidate had 4.
2. **Four different supplied transactions, four different `pk` values, one identical leaf.**
3. **The `levels` differ between the four responses**, so these are four distinct candidates, not one
   cached response returned four times.
4. `709306ae…` was **never** sent to `/mining/candidateWithTxsAndPk`. It reached the node only through
   `/transactions`, at `13:08:05.849`.
5. The multi-transaction path in our client is disabled and was off for this run; `txs` contains
   exactly one element per request.

## Expected behaviour

Every leaf in `proof.txProofs` should be the id of a transaction supplied in `txs`. Tracing v6.0.3RC1:

```
MiningApiRoute.candidateWithTxsAndPkR
  → GenerateCandidate(txsAndPk.txs, reply = true, forced = false, Some(pk))

CandidateGenerator.generateCandidate
  val stateWithMandatoryTxs = s.withTransactions(txsToInclude)
  lazy val unspentTxsToInclude = txsToInclude.filter { tx => inputsNotSpent(tx, stateWithMandatoryTxs) }
  → createCandidate(..., prioritizedTransactions = unspentTxsToInclude, ...)

CandidateGenerator.createCandidate
  def deriveWorkMessage(block: CandidateBlock) =
    powScheme.deriveExternalCandidate(block, minerPk, prioritizedTransactions.map(_.id))

AutolykosPowScheme.deriveExternalCandidate
  val ps = mandatoryTxIds.flatMap { txId => bt.proofFor(txId).map(mp => TransactionMembershipProof(txId, mp)) }

TransactionMembershipProof
  merkleProofEncoder:      "leaf" -> Algos.encode(mp.leafData)
  txMembershipProofEncoder: tmp.proof.asJson          // txId is not emitted

BlockTransactions.merkleTree
  Algos.merkleTree(LeafData @@ txIds)                     // v1
  Algos.merkleTree(LeafData @@ (txIds ++ witnessIds))     // v2+
```

`mandatoryTxIds ⊆ txsToInclude` because `filter` can only remove elements, and the tree's leaves are
transaction ids, so every emitted `leaf` should be a member of `txsToInclude`. The observed leaf is
not.

One detail that may be relevant: `txMembershipProofEncoder` discards `TransactionMembershipProof.txId`
and serialises only the `MerkleProof`, so the emitted `leaf` comes from `mp.leafData` rather than from
the id the proof was requested for. If `leafData` can be populated from a position in the tree rather
than from the requested id, that would produce exactly this symptom.

## Reproduction conditions

Not reproduced on demand, but the correlation is consistent across several sessions:

- **Occurs** when the node has recently accepted the reporter's own transactions into its mempool via
  `/transactions`. Every occurrence follows such a submission within a few seconds.
- **Does not occur** on a freshly restarted client with no transactions in flight.

Suggested reproduction: submit a transaction via `/transactions`, then within a few seconds POST a
different, unrelated transaction to `/mining/candidateWithTxsAndPk` and inspect `proof.txProofs`.

## Secondary observation — `cachedFor` ignores the requested `pk`

Separable from the above; delete if out of scope. It is included because it was observed in the same
four requests.

```scala
def cachedFor(
  candidateOpt: Option[Candidate],
  txs: Seq[ErgoTransaction]
): Boolean = {
  candidateOpt.isDefined && candidateOpt.exists { c =>
    txs.isEmpty || (txs.size == c.txsToInclude.size && txs.forall(c.txsToInclude.contains))
  }
}
```

`GET /mining/candidate` issues `GenerateCandidate(Seq.empty, …)`, so `txs.isEmpty` is always true and
the cached candidate is returned **whatever `pk` it was built for**.

Observed here: after each POST above, the client called `GET /mining/candidate` and received the same
candidate it had just been given for the lender's `pk` — confirmed by the `msg` values matching. A
client using `GET /mining/candidate` as its "mine under my own key" path can therefore be handed a
candidate that pays a different key, with nothing in the response marking it as such.

Question: is `pk` intentionally excluded from the cache key?

## Questions for maintainers

1. Is `proof.txProofs` intended to contain only ids supplied in `txs`?
2. Given `txMembershipProofEncoder` drops `txId` and emits `mp.leafData`, is `leaf` guaranteed to
   equal the requested transaction id?
3. If a supplied transaction is filtered by `inputsNotSpent` or dropped by `collectTxs`, is there an
   intended way for the caller to learn that from the response? At present that case and this one are
   indistinguishable to a miner.
4. Should `cachedFor` include the requested `pk`?

a-shannon:
## Summary

- return transaction membership proofs for the requested transaction when witness ids repeat
- traverse the existing Merkle tree with its real ordered leaf count instead of scrypto's deduplicated map length
- cover duplicate witnesses, padding boundaries, absent ids, and the mandatory-mining consumer

## Root cause

For block versions after v1, `BlockTransactions` commits to `txIds ++ witnessIds`.
The scrypto tree keeps every leaf in `topNode`, but its hash-to-index map collapses
duplicate leaf hashes. `proofByElement` then traverses the real tree using the map's
smaller size, so it can return a proof for another leaf that still validates against
the transaction root.

This change locates the requested transaction in `txIds` and traverses the existing
tree with `n` leaves for v1 or `2n` leaves for later versions. It also fails closed if
the traversal does not terminate at the requested leaf.

## Compatibility

The ordered leaves, Merkle-tree construction, transaction root, block serialization,
JSON schema, and consensus bytes are unchanged. Only membership-proof selection is
corrected. The transaction lookup is intentionally tx-only and uses a bounded linear
scan followed by logarithmic tree traversal.

The generic duplicate-leaf fix is already merged in scrypto #5. This `v6.0.6` branch
still resolves scrypto 3.0.0, so the node-local fix remains necessary here.

## Validation

- regression reproduced on the current `v6.0.6` base: the test-only `BlockTransactionsSpec` run passed 1 test and failed the 2 new proof-identity cases
- `sbt "ergoCore/Test/testOnly org.ergoplatform.modifiers.history.BlockTransactionsSpec"` — 3 passed
- `sbt "ergoCore/Test/testOnly org.ergoplatform.serialization.SerializationCoreTests -- -z BlockTransactions"` — 1 passed
- `sbt "Test/testOnly org.ergoplatform.mining.ErgoMinerSpec"` — 5 passed
- `sbt "Test/testOnly org.ergoplatform.http.routes.BlocksApiRouteSpec"` — 10 passed
- GitHub CI — 8/8 jobs passed on `202f762cb38b6ac56c6cc9afdb9bf51aa5208dca`
- independent review of the exact 3-file patch — no findings

## Related work

Fixes #2463.

- #2373 fixes candidate caching by miner public key; it does not change transaction-proof lookup.
- #2348 touches `BlockTransactions` serializer parsing only; it does not change membership-proof selection.
- ergoplatform/scrypto#5 is the merged generic-library fix; this PR covers the dependency version used by the release branch.

### DIFF
--- a/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
@@ -12,8 +12,8 @@ import org.ergoplatform.nodeView.mempool.TransactionMembershipProof
 import org.ergoplatform.settings.{Algos, Constants}
 import org.ergoplatform.modifiers.history.header.Header.Version
 import org.ergoplatform.serialization.ErgoSerializer
-import scorex.crypto.authds.LeafData
-import scorex.crypto.authds.merkle.{Leaf, MerkleProof, MerkleTree}
+import scorex.crypto.authds.{LeafData, Side}
+import scorex.crypto.authds.merkle.{InternalNode, Leaf, MerkleProof, MerkleTree, Node}
 import scorex.crypto.hash.Digest32
 import scorex.util.serialization.{Reader, Writer}
 import scorex.util.{ModifierId, bytesToId, idToBytes}
@@ -74,7 +74,18 @@ case class BlockTransactions(headerId: ModifierId,
     * @return Some(proof) or None (if transaction with given id is not in the block)
     */
   def proofFor(txId: Array[Byte]): Option[MerkleProof[Digest32]] =
-    merkleTree.proofByElement(Leaf[Digest32](LeafData @@ txId)(Algos.hash))
+    txIds.indexWhere(_.sameElements(txId)) match {
+      case -1 => None
+      case index =>
+        val leafCount = if (blockVersion == Header.InitialVersion) {
+          txIds.size.toLong
+        } else {
+          txIds.size.toLong + witnessIds.size.toLong
+        }
+        BlockTransactions
+          .proofByIndex(merkleTree, index, leafCount)
+          .filter(_.leafData.sameElements(txId))
+    }
 
   def proofFor(txId: ModifierId): Option[MerkleProof[Digest32]] = proofFor(scorex.util.idToBytes(txId))
 
@@ -99,6 +110,43 @@ case class BlockTransactions(headerId: ModifierId,
 object BlockTransactions extends ApiCodecs {
 
   val modifierTypeId: NetworkObjectTypeId.Value = BlockTransactionsTypeId.value
+
+  private def proofByIndex(tree: MerkleTree[Digest32],
+                           index: Int,
+                           leafCount: Long): Option[MerkleProof[Digest32]] = {
+    def paddedLength(length: Long, target: Long): Long =
+      if (length >= target) length else paddedLength(length * 2, target)
+
+    def loop(node: Node[Digest32],
+             leafIndex: Long,
+             subtreeLength: Long,
+             levels: List[(Digest32, Side)]): Option[MerkleProof[Digest32]] = node match {
+      case internal: InternalNode[Digest32] =>
+        val halfLength = subtreeLength / 2
+        if (leafIndex < halfLength) {
+          loop(
+            internal.left,
+            leafIndex,
+            halfLength,
+            (internal.right.hash, MerkleProof.LeftSide) :: levels
+          )
+        } else {
+          loop(
+            internal.right,
+            leafIndex - halfLength,
+            halfLength,
+            (internal.left.hash, MerkleProof.RightSide) :: levels
+          )
+        }
+      case leaf: Leaf[Digest32] if leafIndex == 0 && subtreeLength == 1 =>
+        Some(MerkleProof[Digest32](leaf.data, levels)(leaf.hf))
+      case _ =>
+        None
+    }
+
+    if (index < 0 || index.toLong >= leafCount) None
+    else loop(tree.topNode, index.toLong, paddedLength(2L, leafCount), Nil)
+  }
 
   // Used in the miner when a BlockTransaction instance is not generated yet (because a header is not known)
   def transactionsRoot(txs: Seq[ErgoTransaction], blockVersion: Version): Digest32 = {
--- a/ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
+++ b/ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
@@ -58,5 +58,17 @@ class BlockTransactionsSpec extends ErgoCorePropertyTest {
       }
     }
   }
+  property("Merkle proof returns the requested transaction when witness ids repeat") {
+    val txs = fixtureTransactions.take(3)
+    txs.map(_.id).distinct should have size 3
+    txs.map(_.witnessSerializedId.toSeq).distinct should have size 1
+
+    val blockTransactions = BlockTransactions(headerId, Header.HardeningVersion, txs)
+    val requestedTx = txs.head
+    val proof = blockTransactions.proofFor(requestedTx.id).get
+
+    BlockTransactions.proofValid(blockTransactions.digest, proof) shouldBe true
+    proof.leafData.sameElements(requestedTx.serializedId) shouldBe true
+  }
 
 }
--- a/ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
+++ b/ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
@@ -1,10 +1,27 @@
 package org.ergoplatform.modifiers.history
 
+import org.ergoplatform.modifiers.history.header.Header
+import org.ergoplatform.modifiers.mempool.ErgoTransaction
+import org.ergoplatform.settings.Constants.TrueTree
 import org.ergoplatform.utils.ErgoCorePropertyTest
+import org.ergoplatform.{ErgoBoxCandidate, Input}
+import scorex.crypto.authds.ADKey
+import scorex.util.bytesToId
 
 class BlockTransactionsSpec extends ErgoCorePropertyTest {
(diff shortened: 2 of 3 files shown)


a-shannon:
Added an explicit deterministic regression in 30d9c02c. It creates three distinct transaction IDs with the same witness ID, first verifies that the proof is valid against the block root, and then verifies that the leaf is the requested transaction ID. On the parent commit the test fails only on the leaf-identity assertion; on this branch it passes. `BlockTransactionsSpec` is green (3/3).

a-shannon:
@kushti #2469 is ready for maintainer review. The current head (202f762c) is one commit on top of v6.0.6, mergeable, and CI is 8/8 green. It fixes #2463 on the release branch that still resolves scrypto 3.0.0; the ordered leaves, Merkle root, serialization, and consensus bytes are unchanged.

PR #2471, opened 2026-08-18
Fix for lastHeaders in upcoming state context , miner not stuck on divergence in block pre-validation and validation

kushti:
This PR fixes two mining-related issues. 

First, it prevents the node from stalling after a locally mined block is semantically or syntactically rejected by clearing `solvedBlock` and cached candidates, allowing the miner to resume work instead of rejecting new solutions with “Block already solved.”

 Second, it resolves a divergence between candidate construction and full-block validation: `UpcomingStateContext` (used by `createCandidate`, `/transactions/check`, and mempool validation) was exposing 10 previous headers while full-block validation only exposed 9, so a script reading `CONTEXT.headers(9)` could pass candidate validation and then fail block validation with `ArrayIndexOutOfBoundsException`. The fix trims `lastHeaders` to `Constants.LastHeadersInContext - 1` in `upcoming()` and `simplifiedUpcoming()`, adds a defensive size check in `UpcomingStateContext`, and includes equivalence tests verifying that both upcoming contexts expose the same headers as the post-`appendFullBlock` validation context.

### DIFF
--- a/ergo-core/src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/nodeView/state/ErgoStateContext.scala
@@ -41,6 +41,11 @@ case class UpcomingStateContext(override val lastHeaders: Seq[Header],
   extends ErgoStateContext(lastHeaders, lastExtensionOpt, genesisStateDigest, currentParameters,
                             validationSettings, votingData)(chainSettings) {
 
+  if (lastHeaders.size > Constants.LastHeadersInContext - 1) {
+    log.error(s"UpcomingStateContext created with ${lastHeaders.size} last headers, " +
+              s"expected at most ${Constants.LastHeadersInContext - 1}")
+  }
+
   override def sigmaPreHeader: sigma.PreHeader = PreHeader.toSigma(predictedHeader)
 
   override def sigmaLastHeaders: Coll[sigma.Header] = {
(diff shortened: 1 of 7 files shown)


odiseusme:
Reviewed against today's incident log (6.1.3 node, 17 blocks rejected for b44970ed between 21:44 and 09:08 UTC, all txScriptValidation).

Context fix: take(LastHeadersInContext - 1) on both upcoming() and simplifiedUpcoming() is the right shape (newest-first, parent stays CONTEXT.headers(0)) and the tests compare ids, not just sizes. Good.

Eviction path: today's rejection was txScriptValidation, which ErgoTransaction.scala:157 tags with the tx id and ErgoTransaction typeId, so extractFailedTxId would have caught b44970ed on the first rejected block instead of hours later. Verified pmodModify re-reads memoryPool() after updateState, so the invalidate inside applyState survives. When the tx is still in the pool, invalidate records its id in invalidatedTxIds and ErgoNodeViewSynchronizer refuses peer inventory for it (mp.isInvalidated); OrderedTxPool.canAccept itself does not consult that cache, so direct/local resubmission is not blocked. Suggest one more t20-style case using a script failure rather than a negative output, since that's the incident shape.

Scope: eviction also fires on remote invalid blocks, so a miner willing to forfeit a block can make every upgraded node that holds the tx evict it (peer inventory stays suppressed while invalidatedTxIds keeps the id). After this context fix, HEIGHT alone no longer produces the mismatch on the current tip; it needs a reorg or dependence on another pre-header field, so the surface is small. Note applyState/updateState's `local` is propagated from the triggering pmodModify call across all modToApply, so it isn't per-block provenance; if the intent is only to unwedge locally solved blocks, the condition should identify the failing locally solved block; if remote suppression is intended, worth stating that policy explicitly.

onSolvedBlockFailed: today the SemanticallyFailedModification carried the full-block typeId (-127) and the header id; block.toSeq includes the header so the filter matches. CandidateGeneratorSpec only exercises the BlockTransactions id; adding the header-id case would pin that.

PR #2473, opened 2026-08-18
Candidate for 6.0.6 release

kushti:
* #2432 
* #2477
* #2493
* #2512
* #2547
* #2549 
* #2555

### DIFF
--- a/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
+++ b/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
@@ -3,12 +3,14 @@ package org.ergoplatform.it
 import java.io.File
 import java.util.concurrent.TimeoutException
 import com.typesafe.config.Config
-import org.ergoplatform.it.api.NodeApi.NodeInfo
+import io.circe.Json
+import org.ergoplatform.it.api.NodeApi.{NodeInfo, nodeInfoDecoder}
 import org.ergoplatform.it.container.{IntegrationSuite, Node}
 import org.ergoplatform.nodeView.history.ErgoHistoryUtils
+import org.ergoplatform.it.util.ConvergenceObservations
 import org.scalatest.freespec.AnyFreeSpec
 import scala.async.Async
-import scala.concurrent.{Await, Future, blocking}
+import scala.concurrent.{Await, Future}
 import scala.concurrent.duration._
 
 class DeepRollBackSpec extends AnyFreeSpec with IntegrationSuite {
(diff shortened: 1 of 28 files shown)


odiseusme (commented):
Reviewed the candidate at 2046cfc2e. Two blockers, both in candidate generation; I found no additional blockers in the production diff, subject to the assessment limits below.

**Blocker 1 - the candidate does not compile.** `CandidateGenerator.scala:762:93`: `EliminateTransactions(retryEliminate)` where `retryEliminate` is already an `EliminateTransactions` (found `EliminateTransactions`, required `Seq[ModifierId]`). Reproduced locally with `sbt compile`. The retry-success path at :751 passes it correctly.

**Blocker 2 - the change contradicts its own comment.** The emission-only fallback is reached only when both proof attempts failed, so the second collection's classifications were never validated against a state that produced a usable candidate. At `f002151d2` this branch read `EliminateTransactions(Seq.empty)` under the comment "Both collections produced failed proofs; their rejections may be stale." At the head the comment is unchanged but the code attempts to propagate `retryEliminate`; removing only the extra wrapper would retain potentially stale rejections. I'd keep the empty list on the second-failure fallback, and keep the retry's classifications on retry success, as in #2559. @a-shannon reports a rollback regression failing here once the type error is fixed locally; that matches, though I have not run it myself.

Non-blocking, from reading the rest of the diff:
1. Mempool JSON (`TransactionsApiRoute.scala:133,321`): `cost` is new, and the digest-state ErgoTree query now serializes with resolved inputs so digest mode matches the declared `ErgoTransactionWithInputBoxes` schema. Both are visible to API consumers and deserve a release note.
2. Wallet change addresses (`ErgoWalletService.scala:402`, `ErgoWalletState.scala:85-98`): a change address must now be tracked and signable, and updates require an unlocked wallet. Previously persisted external change addresses stop being used and change goes to the root address. That is a behaviour change for existing operators and needs a release note beyond the OpenAPI text.
3. NiPoPoW (`ErgoNodeViewHolder.scala:320-330`): applying a proof now calls `setHeadersChainSynced()` when `utxoBootstrap` is on. This starts snapshot discovery earlier; it does not itself establish a change in the proof-validation trust boundary.
4. Inbound accounting (`NetworkController.scala:185-198,245-267`) looks correct to me: the pending slot covers only `Connected` to `ConfirmConnection`, that decision is local, dead sockets are caught by `Terminated`, and post-confirmation handlers are bounded by `handshakeTimeout`.

Not assessed: crash-consistency ordering between snapshot state and history writes, digest replay from the recovered checkpoint, and whether cached or in-flight body sections remain correctly gated during snapshot bootstrap.

odiseusme:
Both blockers are resolved at `ac6f9d943`. The emission-only fallback reads `EliminateTransactions(Seq.empty)` again, which fixes the compile error and the stale-rejection propagation together, and the code now matches its comment.

Local on `ac6f9d943`: `testOnly *CandidateRetryReorgSpec *CandidateGeneratorSpec`, 28/28 passed, including the second-collection rollback regression.

The non-blocking items in my earlier comment still stand, in particular release notes for the mempool JSON `cost` field, the digest-mode response shape, and the wallet change-address behaviour change.

PR #2489, opened 2026-08-21
Post 6.0.4 additional tests

kushti:
…ock solvedBlock failure

### DIFF
--- a/src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
+++ b/src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
@@ -151,6 +151,70 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
     system.terminate()
   }
 
+  it should "recover when locally mined block is invalidated by full block id" in new TestKit(
+    ActorSystem()
+  ) {
+    val replyProbe = new TestProbe(system)
+    // fake node view holder: solved block is never applied, so solvedBlock stays set
+    val viewHolderProbe = new TestProbe(system)
+
+    // real readers holder over real node view holder, needed for candidate generation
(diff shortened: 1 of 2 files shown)


odiseusme (approved):
Reviewed against merge-base c313356. Tests-only, +104/-2, both tests verified against the production code they target.

CandidateGeneratorSpec "recover when locally mined block is invalidated by full block id": exercises onSolvedBlockFailed correctly. The handler matches by modId only against the solved block's section ids (toSeq = header +: blockSections, and ErgoFullBlock.id = header.id), so publishing block.id clears the stall; the event's typeId is ignored by the handler. The probe expectations (header, then mandatoryBlockSections in order) match sendToNodeView exactly for a Utxo-state config. On the ordering between the eventStream publish and the subsequent GenerateCandidate tell: Akka's local EventStream currently enqueues to subscriber mailboxes on the publishing thread, so with the default FIFO mailbox the invalidation is processed first in practice - this is implementation behavior rather than a documented ordering contract, but it makes the test reliable as written. The test brackets the behavior properly: stall confirmed while solvedBlock is set, recovery confirmed after invalidation.

t21 "txScriptFailure carries failing transaction id": complements t20 (preservation failure) with the script-failure path. Verified the helper behavior the comment relies on - validTransactionFromBoxes falls back to emptyProverResult inputs when signing returns Failure, so the FalseTree spend produces the intended unprovable tx. The assertion checks the end result, extractFailedTxId == Some(invalidTx.id); it does not pin which branch of extractFailedTxId resolved it. Optional, non-blocking: asserting the cause is a MalformedModifierError with ErgoTransaction.modifierTypeId would pin the script-failure path specifically and exclude e.g. a TooHighCostError resolution. Registration in verifyingTxCases only, and the inner verifyTransactions guard, both match the file's convention (t20).

Ran both suites locally on the PR head: 117/117 passed.

Approving. The one suggestion above is optional.

Issue #1614, opened 2022-02-02
Make a test that the UTXO set tree is not modified as a result of proofsForTransactions call


kushti:
While the current 
```
property("concurrent applyModifier() and proofsForTransactions()")
```

test ensures that ```proofsForTransactions()``` and ```applyModifier()``` can work in parallel , it is not checking that tree is not modified in some edge cases. Actually, it could be the case that it happens in some edge cases.

Make a test which is generating e.g. 500 blocks of 500 txs each as testing samples. Call proofsForTransactions first, check that digest from its output corresponds to header's stateRoot, and header.ADProofsRoot corresponds to ```ADProofs.proofDigest(proofBytes)```. Then call applyModifier and check it is finishing successfully

dimension-drifter:
@kushti i have raised a PR for this issue. Please Review it

Ergologica:
Is the gap here the scale rather than the invariant itself? On current master `UtxoStateSpecification` already carries two properties that assert it:

- `proofsForTransactions() does not change state digest` (line 471) - takes the root digest, calls `proofsForTransactions` twice, and asserts the digest is unchanged after each call;
- `proofsForTransactions() to be deterministic` (line 289) - same transactions in, same proof digest and state digest out.

Both run against a state freshly built from a `boxesHolderGen` sample. What they do not cover is the shape this issue sizes it at: the invariant holding on a tree that has already been mutated many times by `applyModifier`, rather than on a newly created one - which is where an in-place mutation would actually show up.

I have that version written and green: root digest and version checked around every `proofsForTransactions` call along a chain, with repeated and interleaved calls at each height, and a check that applying the block moves the root exactly to the digest the proof announced. Happy to open it as a PR if that is the missing piece.

Asking first rather than sending a PR because this is still listed as an open 200 SigUSD bounty in the ErgoDevs registry, and if you consider the invariant already covered by the two properties above then the issue should be closed instead - I would rather not claim a bounty that is already satisfied.

PR #2493, opened 2026-08-29
Test that proofsForTransactions() does not modify the UTXO set tree along a growing chain

Ergologica:
Addresses #1614.

**What was already covered, and what was missing.** `UtxoStateSpecification` already checks that a `proofsForTransactions()` call does not change the state digest on a fresh state ("proofsForTransactions() does not change state digest") and that it is deterministic there ("proofsForTransactions() to be deterministic"). No test checked the same invariants while a chain grows - a tree with real history, with proofs generated between block applications, which is how the node actually uses it. I asked about this reading of the scope in an issue comment on 2026-08-11 and heard no objection.

**The new property**, `proofsForTransactions() does not modify the UTXO set tree along a growing chain`, builds a chain of valid full blocks the same way the neighbouring concurrency test does, and at every height asserts:

1. the tree root is unchanged after generating proofs (`rootDigest` before == after);
2. a second call returns the identical proof digest and state digest, and still leaves the root untouched;
3. the block assembled from those proofs (`stateRoot = adDigest`, `ADProofsRoot = ADProofs.proofDigest(...)`) is accepted by `applyModifier`, and the state lands exactly on the promised digest.

**On the numbers in the issue**: the literal 500 blocks x 500 transactions would take hours in CI, so the chain here is 20 blocks with the standard generator transactions - the invariants are the same at any scale, and the length is a one-line constant. Happy to raise it, or to add an `ignored` long-run variant with the literal numbers, if preferred.

**Verified the test bites**: injecting a mutation into `proofsForTransactions` (staging an extra `Insert` on the persistent prover and committing it) makes the property fail on the root check; on clean master it is green, and the full `UtxoStateSpecification` suite passes locally (22/22).

Note: the issue shows a years-old assignee with no linked work; if that work is still planned I am happy to defer.

### DIFF
--- a/src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
+++ b/src/test/scala/org/ergoplatform/nodeView/state/UtxoStateSpecification.scala
@@ -439,6 +439,49 @@ class UtxoStateSpecification extends ErgoCorePropertyTest with OptionValues {
     }
   }
 
+  property("proofsForTransactions() does not modify the UTXO set tree along a growing chain") {
+    // Chain-scale variant of the invariants above, as asked in #1614: at every height of a
+    // growing chain, generating proofs must leave the tree untouched and be repeatable, and
+    // the block built from those proofs must be applicable, landing exactly on the promised digest.
+    // The issue literally asks for 500 blocks of 500 transactions; that would take hours in CI,
+    // so the chain is kept short here - the invariants checked are the same at any scale.
+    val blocksInChain = 20
+
+    var bh = BoxHolder(Seq(genesisEmissionBox))
+    var us = createUtxoState(bh, parameters)
+    var height: Int = GenesisHeight
+
+    (0 until blocksInChain) foreach { _ =>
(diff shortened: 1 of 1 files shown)


Ergologica:
CI note: the red job is `Run it node tests`. Both failures are Docker multi-node convergence timeouts, in suites this PR does not touch:

- `DeepRollBackSpec` ("Deep rollback handling") - `TimeoutException: Nodes did not converge to the same best full block at height >= 255`
- `ForkResolutionSpec` ("Fork resolution after isolated mining") - `TimeoutException: Futures timed out after [15 minutes]`

This PR only adds one property to `UtxoStateSpecification`, which runs in `Run node tests` - green on this run, as are the wallet and ergo-core jobs; 11 of 13 integration tests passed too, including `UtxoStateNodesSyncSpec`. `DeepRollBackSpec` timeouts have hit other PRs as well (see #2465 / #2469, "Retrigger CI after DeepRollBackSpec timeout"). Happy to rebase to retrigger if a green re-run is preferred.

kushti (approved):

Issue #1995, opened 2023-05-09
Remaining checks and todos for bootstrapping with UTXO set snapshots


kushti:
What to check / todo
* prune manifests and UTXO set snapshot chunks after download is complete and 
* check that there are no issues with small UTXO set
* check if nextModifiersToDownload() can ask for snapshots too many times , fix if so
* more tests

PR #2496, opened 2026-09-02
Fixes for nipopows + utxo set snapshot bootstrapping

### LINKED ISSUE #2464: NiPoPoW/UTXO bootstrap stuck at negotiation stage — no progress, "Extension is empty while comparison is younger" warning (v6.1.3)
**Node version:** 6.1.3
**RAM/CPU:** Intel MacBook Pro intel i5 16GB RAM (2018 model)
**OS:** macOS 15.7.9
**Storage location:** SSD (internal)

I am a beginner with this sort of project, relying heavily on Claude and Grok build.

### Summary

Running a pruned node with both `utxoBootstrap` and `nipopowBootstrap` enabled. Sync has been stuck indefinitely with no forward progress. Node is connected to 60 peers and exchanging normal P2P traffic (Sync messages, inventory), but `headersHeight` is frozen and neither bootstrap method appears to be pulling actual payload data.

### Config (`ergo.conf`, node section)

```
node {
  mining = false
  stateType = "utxo"
  blocksToKeep = 1440
  utxo {
    utxoBootstrap = true
    storingUtxoSnapshots = 0
    p2pUtxoSnapshots = 1
  }
  nipopow {
    nipopowBootstrap = true
    p2pNipopows = 2
  }
  ...
}
```

### Observed state (stable/unchanged over an extended period)

- `headersHeight`: 1827762 (frozen, not advancing)
- `fullHeight`: null
- `maxPeerHeight`: ~1849430 (climbing normally as network advances)
- `peersCount`: 60
- `stateVersion`: genesis
- `.ergo/snapshots` directory: contains only metadata, no actual snapshot chunk data ever downloaded

### Log analysis

- `grep -ci 'nipopow'` on the full current log returns **0** — no "requesting NiPoPoW proof" activity, no interlink-related log lines at all, despite `nipopowBootstrap = true`.
- No evidence of a fallback to full/normal header sync either — no "applying header" / "new header" volume, and `headersHeight` genuinely isn't moving.
- Dominant log activity is routine P2P chatter: thousands of `Processing 400/399 non-t invs (of type 101)`, repeated `Send message MessageSpec(65: Sync)`.
- The one recurring, non-routine signal is this WARN, appearing multiple times:
  ```
  WARN [...] o.e.n.ErgoNodeViewSynchronizer - Extension is empty while comparison is younger
  ```
- Occasional `No peers available in requestDownload`, but this doesn't correlate with the stall since peer count stays healthy at ~60 throughout.

### Interpretation / question for maintainers

Both bootstrap paths (NiPoPoW and UTXO snapshot) appear to be stuck at a negotiation/metadata stage rather than actively transferring data or falling back to normal sync. Given `p2pNipopows = 2` and `p2pUtxoSnapshots = 1` require some threshold of peers to agree/provide matching proofs or snapshots, is it possible the node is deadlocking when it can't currently find enough peers simultaneously serving valid, matching data for the configured threshold — rather than falling back or retrying with a relaxed threshold?

Is the `Extension is empty while comparison is younger` warning expected/benign during this stage, or is it symptomatic of the stall (e.g. a peer's sync-status extension, which should carry NiPoPoW interlink data, coming back empty during comparison)?

### What I've tried / plan to try

- Confirmed config matches documented recommendations for racing both bootstrap methods on a pruned node.
- Considering lowering `p2pNipopows` to 1 to see if the threshold is the blocker.
- Considering disabling `nipopowBootstrap` to isolate whether UTXO snapshot bootstrap alone makes progress.
- As a fallback, planning to disable both bootstrap flags, do a plain full sync (with `blocksToKeep` already set so it self-prunes as it goes), and once fully synced, set `storingUtxoSnapshots` to a non-zero value to help serve future bootstrapping peers.
- Willing to enable DEBUG logging on network/bootstrap classes and provide a fuller log excerpt if useful.

Happy to provide the full log or run additional diagnostics on request.

UPDATE 1: proof gets fetched and tentatively applied (confirmed via stdout log showing height 1827829), but reverts — /info headersHeight remains frozen at 1827762 hours later, with Extension is empty while comparison is younger recurring throughout.

UPDATE 2 — Reproduced from clean state; new finding on peer connection drops

Did a full clean start to rule out corrupted local state as the cause: stopped the node, deleted ~/.ergo/state and ~/.ergo/history (config/peer db untouched), restarted with nipopowBootstrap re-enabled specifically to test reproducibility on a fresh genesis state.

Result: identical failure pattern reproduces from clean state.

- headersHeight frozen at 1827327 across two checks ~10 min apart (no movement). stateVersion still all zeros.
- Log confirms the same apply-then-revert behavior seen pre-wipe: a NiPoPoW proof for height 1827829 was fetched and logged as applied ("Nipopow proof applied, best header now is ... height: 1827829"), but /info headersHeight never reflects this — it stayed at 1827327, lower than the previous stall point, confirming this is an independent, freshly-reproduced stall rather than a residual artifact from the prior session.
- "Extension is empty while comparison is younger" (WARN, ErgoNodeViewSynchronizer) continued recurring throughout.

New finding — peer count collapse during the stall window:

- peersCount dropped from 52 to 11 within roughly 10 minutes.
- Log shows a wave of "Connection closed ... reason: error: Connection reset" and "error: Operation timed out" entries clustered in the same window, affecting many distinct peer IPs.
- This looks like more than normal churn and may point to the actual mechanism behind the stall: if the proof apply/revert cycle causes malformed or repeated requests, peers could be actively dropping or rate-limiting the connection — which would explain why the node can never sustain the quorum (2 proofs from independent peers) long enough to finalize a NiPoPoW bootstrap, rather than a simple unavailability of proof-serving peers.

Given this reproduces cleanly and doesn't self-resolve with time, falling back to nipopowBootstrap=false / utxoBootstrap=false (plain full sync) for this node going forward. Flagging here in case the peer-drop pattern is useful to anyone investigating the NiPoPoW bootstrap path in 6.1.3.

kushti:
Close #2464 : nipopows + utxo set snapshot bootstrapping fixes for p2p layer done in this PR

Also contains #2468

### DIFF
--- a/.github/workflows/bootstrap-mainnet.yml
+++ b/.github/workflows/bootstrap-mainnet.yml
@@ -1,0 +1,42 @@
+name: Mainnet UTXO snapshot bootstrap test
+
+# Boots a mainnet node with nipopowBootstrap + utxoBootstrap from an empty data dir and
+# waits for a full sync. Validates the fast-bootstrap path end-to-end against the real
+# network. Kept separate from release.yml so the multi-hour run cannot block publishing.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  bootstrap_mainnet:
+    name: NiPoPoW + UTXO snapshot bootstrap on mainnet
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout current branch (full)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Java and Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.8
+
+      - name: Cache sbt
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.sbt
+            ~/.ivy2/cache
+            ~/.coursier/cache/v1
+            ~/.cache/coursier/v1
+            ~/AppData/Local/Coursier/Cache/v1
+            ~/Library/Caches/Coursier/Cache/v1
+          key: ${{ runner.os }}-sbt-cache-v4-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
+
+      - name: Run mainnet bootstrap test
+        run: |
+          mkdir tmp
+          TMPDIR=$(pwd)/tmp sbt -Denv=test clean ++2.12.20 docker "it2:testOnly *TestUtxoSnapshotBootstrapOnMainNetSpec"
--- a/src/it2/scala/org/ergoplatform/it2/TestUtxoSnapshotBootstrapOnMainNetSpec.scala
+++ b/src/it2/scala/org/ergoplatform/it2/TestUtxoSnapshotBootstrapOnMainNetSpec.scala
@@ -1,0 +1,60 @@
+package org.ergoplatform.it2
+
+import com.typesafe.config.{Config, ConfigFactory}
+import org.ergoplatform.it.api.NodeApi.NodeInfo
+import org.ergoplatform.it.container.{IntegrationSuite, Node}
+import org.scalatest.OptionValues
+import org.scalatest.flatspec.AnyFlatSpec
+
+import scala.async.Async
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+class TestUtxoSnapshotBootstrapOnMainNetSpec
+  extends AnyFlatSpec
+    with IntegrationSuite
+    with OptionValues {
+
+  // Unlike TestOnMainNetSpec, no host volume is mounted: node data lives in the container's
+  // anonymous volume (/home/ergo/.ergo) and is discarded with the container, so every run
+  // performs a real bootstrap from an empty data dir.
+
+  val bootstrapConfig: Config = ConfigFactory.parseString(
+    s"""
+       |ergo.node.utxo.utxoBootstrap = true
+       |ergo.node.nipopow.nipopowBootstrap = true
+    """.stripMargin
+  )
+
+  val nodeConfig: Config = bootstrapConfig
+    .withFallback(nodeSeedConfigs.head)
+    .withFallback(nonGeneratingPeerConfig)
+  val node: Node = docker.startMainNetNodeYesImSure(nodeConfig).get
+
+  it should "Bootstrap from a UTXO set snapshot via NiPoPoW proof on mainnet and fully sync" in {
+    // Phase 1: headers appear, proving the trusted NiPoPoW proof was applied
+    val headersResult = Async.async {
+      Async.await(node.waitFor[NodeInfo](
+        _.info,
+        nodeInfo => nodeInfo.bestHeaderHeightOpt.exists(_ > 1000),
+        1.minute
+      ))
+    }
+    val nodeInfoAfterHeaders = Await.result(headersResult, 1.hour)
+    log.info(s"Headers appeared, best header height: ${nodeInfoAfterHeaders.bestHeaderHeightOpt}")
+
+    // Phase 2: wait for a full sync (snapshot applied + full blocks downloaded)
+    val syncResult = Async.async {
+      Async.await(node.waitFor[NodeInfo](
+        _.info,
+        nodeInfo => nodeInfo.bestBlockHeightOpt.exists(nodeInfo.bestHeaderHeightOpt.contains),
+        1.minute
+      ))
+    }
+    val syncedInfo = Await.result(syncResult, 5.hours)
+
+    // guard against a degenerate "synced at genesis" pass
+    syncedInfo.bestHeaderHeightOpt.value should be > 1000000
+  }
+
+}
--- a/src/it2/scala/org/ergoplatform/it2/TestUtxoSnapshotBootstrapOnMainNetSpec.scala
+++ b/src/it2/scala/org/ergoplatform/it2/TestUtxoSnapshotBootstrapOnMainNetSpec.scala
@@ -23,6 +23,9 @@ class TestUtxoSnapshotBootstrapOnMainNetSpec
     s"""
        |ergo.node.utxo.utxoBootstrap = true
        |ergo.node.nipopow.nipopowBootstrap = true
+       |# genesisId of mainnet.conf, needed for ErgoSettings validation on the test host,
+       |# where the network config file is not loaded (see Docker.buildErgoSettings)
+       |ergo.chain.genesisId = "b0244dfc267baca974a4caee06120321562784303a8a688976ae56170e4d175b"
     """.stripMargin
   )
 
--- a/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
+++ b/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
@@ -1147,8 +1147,22 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
           Seq.empty
         }
       case _ =>
-        log.info(s"Processing ${invData.ids.length} non-tx invs (of type $modifierTypeId) from $peer")
-        invData.ids.filter(mid => deliveryTracker.status(mid, modifierTypeId, Seq(hr)) == ModifiersStatus.Unknown)
+        // During UTXO set snapshot bootstrap, ignore block-section invs (extension, transactions, ADProofs)
+        // until the snapshot is applied. Headers and snapshot-related types are still processed.
+        val utxoBootstrapInProgress =
+          settings.nodeSettings.utxoSettings.utxoBootstrap &&
+            !hr.isUtxoSnapshotApplied &&
+            modifierTypeId != Header.modifierTypeId &&
+            modifierTypeId != ManifestTypeId.value &&
+            modifierTypeId != UtxoSnapshotChunkTypeId.value
+
+        if (utxoBootstrapInProgress) {
+          log.debug(s"Ignoring ${invData.ids.length} non-tx invs (of type $modifierTypeId) from $peer: UTXO snapshot bootstrap in progress")
+          Seq.empty
+        } else {
+          log.info(s"Processing ${invData.ids.length} non-tx invs (of type $modifierTypeId) from $peer")
+          invData.ids.filter(mid => deliveryTracker.status(mid, modifierTypeId, Seq(hr)) == ModifiersStatus.Unknown)
+        }
     }
 
     if (newModifierIds.nonEmpty) {
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -316,6 +316,12 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
       if (history().isEmpty) {
         history().applyPopowProof(proof)
         if (!history().isEmpty) {
+          // When UTXO set snapshot bootstrap is enabled, mark headers chain as synced right after
+          // a trusted NiPoPoW proof is applied. This allows the node to start requesting UTXO set
+          // snapshots immediately, instead of waiting for normal header sync to reach the tip.
+          if (settings.nodeSettings.utxoSettings.utxoBootstrap) {
+            history().setHeadersChainSynced()
+          }
           updateNodeView(updatedHistory = Some(history()))
         }
       }
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -583,6 +583,18 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
       )
   }
 
+  private def isPreparedUtxoSnapshotState(state: State, history: ErgoHistory): Boolean =
+    ErgoNodeViewHolder.isPreparedUtxoSnapshotState(
+      state.isInstanceOf[UtxoState],
+      settings.nodeSettings.utxoSettings.utxoBootstrap,
+      history.isUtxoSnapshotApplied,
+      state.version,
+      state.rootDigest,
+      {
+        val snapshotHeight = history.minimalFullBlockHeight - 1
+        history.bestHeaderAtHeight(snapshotHeight)
+      })
+
   private def restoreConsistentState(stateIn: State, history: ErgoHistory): Try[State] = {
     (stateIn.version, history.bestFullBlockOpt, stateIn) match {
       case (ErgoState.genesisStateVersion, None, _) =>
@@ -591,8 +603,12 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
       case (stateId, Some(block), _) if stateId == block.id =>
         log.info(s"State and history have the same version ${encoder.encode(stateId)}, no recovery needed.")
         Success(stateIn)
+      case (_, None, _) if isPreparedUtxoSnapshotState(stateIn, history) =>
+        log.info(s"Prepared UTXO snapshot state ${encoder.encode(stateIn.version)} restored before the first full block")
+        Success(stateIn)
       case (_, None, _) =>
         log.info("State and history are inconsistent. History is empty on startup, rollback state to genesis.")
+        stateIn.closeStorage()
         Success(recreatedState())
       case (_, Some(bestFullBlock), _: DigestState) =>
         log.info(s"State and history are inconsistent. Going to switch state to version ${bestFullBlock.encodedId}")
@@ -721,6 +737,25 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
 
 
 object ErgoNodeViewHolder {
+
+  private[nodeView] def isPreparedUtxoSnapshotState(
+      stateIsUtxo: Boolean,
+      utxoBootstrap: => Boolean,
+      snapshotApplied: => Boolean,
+      stateVersion: VersionTag,
+      stateRoot: Array[Byte],
+      snapshotHeaderOpt: => Option[Header]): Boolean =
+    stateIsUtxo &&
+      utxoBootstrap &&
+      snapshotApplied &&
+      snapshotHeaderOpt.exists(matchesPreparedUtxoSnapshotHeader(stateVersion, stateRoot, _))
+
+  private def matchesPreparedUtxoSnapshotHeader(
+      stateVersion: VersionTag,
+      stateRoot: Array[Byte],
+      header: Header): Boolean =
+    stateVersion == idToVersion(header.id) &&
+      java.util.Arrays.equals(stateRoot, header.stateRoot)
 
   object ReceivableMessages {
     // Tracking last modifier and header & block heights in time, being periodically checked for possible stuck
--- a/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
@@ -294,6 +294,18 @@ object ErgoHistory extends ScorexLogging {
 
     repairIfNeeded(history)
 
+    // If the node is restarted in the middle of UTXO set snapshot bootstrapping (NiPoPoW proof
+    // headers are already in the database, but the snapshot was not applied yet), the headers
+    // chain is already complete, so mark it synced. Otherwise the synchronizer never asks for
+    // the snapshot manifest / block sections (ErgoNodeViewSynchronizer.requestMoreModifiers) and
+    // bootstrapping can never resume.
+    if (nodeSettings.utxoSettings.utxoBootstrap &&
+        nodeSettings.nipopowSettings.nipopowBootstrap &&
+        history.bestHeaderOpt.isDefined &&
+        !history.isUtxoSnapshotApplied) {
+      history.setHeadersChainSynced()
+    }
+
     log.info("History database read")
     if(ergoSettings.nodeSettings.extraIndex) // start extra indexer, if enabled
       context.system.eventStream.publish(StartExtraIndexer(history))
--- a/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
@@ -111,6 +111,11 @@ trait ToDownloadProcessor
     if (!nodeSettings.verifyTransactions) {
       // A regime that do not download and verify transaction
       Nil
+    } else if (nodeSettings.utxoSettings.utxoBootstrap && !isUtxoSnapshotApplied) {
+      // While bootstrapping from a UTXO set snapshot, do not download full block sections
+      // until the snapshot has been applied. Block sections downloaded before the snapshot
+      // would be stored as non-best and never applied to the freshly recreated state.
+      Nil
     } else if (shouldDownloadBlockAtHeight(header.height)) {
       // Already synced and header is not too far back. Download required modifiers.
       requiredModifiersForHeader(header)
--- a/src/test/scala/org/ergoplatform/nodeView/history/UtxoBootstrapRestartSpecification.scala
+++ b/src/test/scala/org/ergoplatform/nodeView/history/UtxoBootstrapRestartSpecification.scala
@@ -1,0 +1,100 @@
+package org.ergoplatform.nodeView.history
+
+import org.ergoplatform.nodeView.mempool.ErgoMemPoolUtils.SortingOption
+import org.ergoplatform.nodeView.state.StateType
+import org.ergoplatform.settings._
+import org.ergoplatform.utils.ErgoCorePropertyTest
+import org.ergoplatform.utils.HistoryTestHelpers.BlocksInChain
+import org.ergoplatform.utils.generators.ChainGenerator._
+import org.ergoplatform.wallet.utils.FileUtils
+
+import scala.concurrent.duration._
+
+/**
+  * Reproduction of a restart deadlock during UTXO set snapshot bootstrapping.
+  *
+  * Scenario: a node bootstrapping via NiPoPoW proofs + UTXO set snapshot (nipopowBootstrap +
+  * utxoBootstrap) is restarted after the NiPoPoW proof headers were persisted but before the
+  * UTXO set snapshot application completed (e.g. killed in the middle of snapshot chunk
+  * downloading). After the restart the node never resumes bootstrapping:
+  *
+  * - `isHeadersChainSynced` is kept in memory only (FullBlockPruningProcessor.isHeadersChainSyncedVar)
+  *   and is set solely from `updateBestFullBlock` or `setHeadersChainSynced` (the latter is called
+  *   only right after a NiPoPoW proof is applied), so it is false after every restart
+  * - on restart the headers are read from the database, so no new NiPoPoW proof is requested
+  *   (ErgoNodeViewSynchronizer.sendSync asks for proofs only when `bestHeaderOpt` is empty)
+  * - the synchronizer asks for the UTXO set snapshot / block sections only when
+  *   `isHeadersChainSynced` is true (ErgoNodeViewSynchronizer.requestMoreModifiers), and
+  *   ToDownloadProcessor filters block-section invs out while the snapshot is not applied
+  *
+  * Result: headers are at tip, the state is at genesis, and the node loops sending sync
+  * messages forever. Observed on mainnet with a node restarted mid snapshot download.
+  *
+  * The property below simulates the restart by reopening the history database and asserts that
+  * the headers chain is considered synced after the restart, so that snapshot bootstrapping
+  * can resume. It fails before the fix.
+  */
+class UtxoBootstrapRestartSpecification extends ErgoCorePropertyTest with FileUtils {
+
+  import org.ergoplatform.utils.ErgoNodeTestConstants._
+
+  private def utxoBootstrapSettings(dir: java.io.File): ErgoSettings = {
+    val txCostLimit = initSettings.nodeSettings.maxTransactionCost
+    val txSizeLimit = initSettings.nodeSettings.maxTransactionSize
+    val nodeSettings = NodeConfigurationSettings(
+      StateType.Utxo,
+      verifyTransactions = true,
+      blocksToKeep = -1,
+      UtxoSettings(utxoBootstrap = true, 0, 2),
+      NipopowSettings(nipopowBootstrap = true, 1),
+      mining = false,
+      txCostLimit,
+      txSizeLimit,
+      blockCandidateGenerationInterval = 20.seconds,
+      useExternalMiner = false,
+      internalMinersCount = 1,
+      internalMinerPollingInterval = 1.second,
+      miningPubKeyHex = None,
+      offlineGeneration = false,
+      200,
+      5.minutes,
+      100000,
+      1.minute,
+      mempoolSorting = SortingOption.FeePerByte,
+      rebroadcastCount = 200,
+      1000000,
+      100,
+      adProofsSuffixLength = 112 * 1024,
+      extraIndex = false
+    )
+    ErgoSettings(dir.getAbsolutePath, NetworkType.TestNet, settings.chainSettings, nodeSettings,
+      null, null, settings.cacheSettings)
+  }
+
+  property("node restarted after nipopow headers persisted but before snapshot application resumes bootstrap") {
+    val dir = createTempDir
+    val historySettings = utxoBootstrapSettings(dir)
+
+    // headers-only chain, result of a NiPoPoW proof application (no full blocks downloaded yet)
+    val history = ErgoHistory.readOrGenerate(historySettings)(null)
+    val headers = genHeaderChain(BlocksInChain, history, diffBitsOpt = None, useRealTs = false)
+    val updHistory = applyHeaderChain(history, headers)
+
+    updHistory.bestFullBlockOpt shouldBe None
+    updHistory.isUtxoSnapshotApplied shouldBe false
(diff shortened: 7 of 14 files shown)


a-shannon:
Thanks @kushti. Reviewed `85044dc444f18235f2e33393acf005e3a6db4979`. The new restart change addresses the lost in-memory synchronization flag for NiPoPoW + UTXO bootstrap. I found one blocking regression in the cumulative PR, separate from that latest change:

**UTXO snapshot bootstrap without NiPoPoW loses its synchronization transition.** With `utxoBootstrap=true`, `nipopowBootstrap=false`, `verifyTransactions=true` and `blocksToKeep=-1`, the new [early return in `ToDownloadProcessor.toDownload`](https://github.com/ergoplatform/ergo/blob/85044dc444f18235f2e33393acf005e3a6db4979/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala#L114-L125) bypasses the fresh-header branch that sets `isHeadersChainSynced`. Consequently, `nextModifiersToDownload` keeps returning empty before it can request snapshot information. The new startup recovery condition requires NiPoPoW, so it cannot recover this configuration either.

Please preserve ordinary header-synchronization detection while suppressing premature block-section downloads. The focused regression should use this configuration, deliver headers through the normal append path until a fresh tip is reached, and assert that snapshot information is requested while block sections remain suppressed. It should not call `setHeadersChainSynced()` itself: the current tests do so and therefore bypass the missing transition. This guard is also carried by #2424, so fixing it here will let the descendant inherit the correction.

Two focused improvements would make the restart evidence reproducible:

- In [`UtxoBootstrapRestartSpecification`](https://github.com/ergoplatform/ergo/blob/85044dc444f18235f2e33393acf005e3a6db4979/src/test/scala/org/ergoplatform/nodeView/history/UtxoBootstrapRestartSpecification.scala#L75-L97), close the original history storage before reopening it, and assert the next snapshot request as well as the flag. `StoreRegistry` otherwise reuses the already-open underlying DB.
- Extend the [functional bootstrap test](https://github.com/ergoplatform/ergo/blob/85044dc444f18235f2e33393acf005e3a6db4979/src/it2/scala/org/ergoplatform/it2/TestUtxoSnapshotBootstrapOnMainNetSpec.scala#L35-L60) to observe partial chunk download, restart with the same data directory, and require subsequent snapshot completion and full-block progress. The committed version currently only starts once and waits for synchronization.

The current 8/8 CI checks are green, but the integration job runs `it:test`; the separate bootstrap workflow runs `it2` on release/manual dispatch. Your successful restart run may cover this separately. My feedback above is based on source review and visible CI; I have not independently rerun the mainnet scenario.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -111,6 +111,11 @@ trait ToDownloadProcessor
     if (!nodeSettings.verifyTransactions) {
       // A regime that do not download and verify transaction
       Nil
+    } else if (nodeSettings.utxoSettings.utxoBootstrap && !isUtxoSnapshotApplied) {
```
jozanek, line 114: This sits above L122-124, the only remaining writer of `isHeadersChainSyncedVar`. With `utxoBootstrap=true, nipopowBootstrap=false` (allowed by `ErgoSettingsReader.scala:191-194`, documented in README) nothing sets the flag: `ErgoHistory.scala:302` needs `nipopowBootstrap`, `ErgoNodeViewHolder.scala:322` needs a proof that config never requests. That node stalls permanently, restart included.

Would calling `setHeadersChainSynced()` inside this branch — not `updateBestFullBlock` — work? Hoisting the L94 snapshot case above the L83 gate would too, and drops changes 2 and 4.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -316,6 +316,12 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
       if (history().isEmpty) {
         history().applyPopowProof(proof)
         if (!history().isEmpty) {
+          // When UTXO set snapshot bootstrap is enabled, mark headers chain as synced right after
+          // a trusted NiPoPoW proof is applied. This allows the node to start requesting UTXO set
+          // snapshots immediately, instead of waiting for normal header sync to reach the tip.
+          if (settings.nodeSettings.utxoSettings.utxoBootstrap) {
```
jozanek, line 322: Gated on `utxoBootstrap`, but the `genesisId` that makes a proof safe to trust is required only under `nipopowBootstrap` (`ErgoSettingsReader.scala:195`), and `processNipopowProof` isn't gated either. `NipopowVerifier` warns that without it, a proof for any chain could be accepted.

I may be missing an upstream filter. Would `nipopowBootstrap && utxoBootstrap` here, or requiring `genesisId` whenever `utxoBootstrap` is set, close that?

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -583,6 +589,18 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
       )
   }
 
+  private def isPreparedUtxoSnapshotState(state: State, history: ErgoHistory): Boolean =
+    ErgoNodeViewHolder.isPreparedUtxoSnapshotState(
+      state.isInstanceOf[UtxoState],
```
jozanek, line 594: `isInstanceOf[UtxoState]` means a digest node never matches L612 and falls through to L615-618, which wipes the state dir. I found no settings rule rejecting `digest` together with `utxoBootstrap`.

`docs/review/architecture.md` asks for all four sync modes covered or excluded explicitly. Could you add either the rejection or a note?

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -294,6 +294,18 @@ object ErgoHistory extends ScorexLogging {
 
     repairIfNeeded(history)
 
+    // If the node is restarted in the middle of UTXO set snapshot bootstrapping (NiPoPoW proof
+    // headers are already in the database, but the snapshot was not applied yet), the headers
+    // chain is already complete, so mark it synced. Otherwise the synchronizer never asks for
+    // the snapshot manifest / block sections (ErgoNodeViewSynchronizer.requestMoreModifiers) and
+    // bootstrapping can never resume.
+    if (nodeSettings.utxoSettings.utxoBootstrap &&
+        nodeSettings.nipopowSettings.nipopowBootstrap &&
```
jozanek, line 303: This requires `nipopowBootstrap`; `ToDownloadProcessor.scala:114` and `ErgoNodeViewSynchronizer.scala:1152` require only `utxoBootstrap`. That gap is what makes the stall above survive a restart.

`bestHeaderOpt.isDefined` already proves a chain exists. Would dropping the `nipopowBootstrap` conjunct work — or better, one shared predicate across all three sites?

src/test/scala/org/ergoplatform/nodeView/history/UtxoBootstrapRestartSpecification.scala
```diff
@@ -0,0 +1,100 @@
...
+    val history = ErgoHistory.readOrGenerate(historySettings)(null)
+    val headers = genHeaderChain(BlocksInChain, history, diffBitsOpt = None, useRealTs = false)
+    val updHistory = applyHeaderChain(history, headers)
+
+    updHistory.bestFullBlockOpt shouldBe None
+    updHistory.isUtxoSnapshotApplied shouldBe false
+    updHistory.isHeadersChainSynced shouldBe false // set only via updateBestFullBlock / setHeadersChainSynced
+
+    // simulate node restart: reopen the same database
+    val restarted = ErgoHistory.readOrGenerate(historySettings)(null)
```
jozanek, line 88: The history at L79 isn't closed first, and `StoreRegistry.add` (`LDBFactory.scala:68-75`) is `getOrElseUpdate` on the directory — so this likely reuses the open DB rather than reopening.

The fresh in-memory flag still makes the assertion meaningful, but not the persistence claim. Would closing the first history, and asserting the follow-on snapshot request, give you what you want here?

src/test/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/UtxoBootstrapToDownloadSpecification.scala
```diff
@@ -0,0 +1,133 @@
...
+    (updHistory, chain)
+  }
+
+  property("snapshot request is issued when headers chain synced and no full blocks applied") {
+    var history = genUtxoBootstrapHistory()
+    val chain = genChain(BlocksInChain, history)
+    history = applyHeaderChain(history, HeaderChain(chain.map(_.header)))
+
+    history.bestFullBlockOpt shouldBe None
+    history.isHeadersChainSynced shouldBe false
```
jozanek, line 45: This asserts the flag stays false after a `genChain` chain (L41) with real timestamps — but on master I'd expect that tip to flip it via `ToDownloadProcessor` L122-124. So it reads as encoding the new early return rather than intended behaviour, which is part of why green CI says nothing about the `nipopowBootstrap=false` config.

I haven't run it. Could this assert the flag flips organically, with no manual `setHeadersChainSynced()`?

.github/workflows/bootstrap-mainnet.yml
```diff
@@ -0,0 +1,42 @@
+name: Mainnet UTXO snapshot bootstrap test
+
+# Boots a mainnet node with nipopowBootstrap + utxoBootstrap from an empty data dir and
+# waits for a full sync. Validates the fast-bootstrap path end-to-end against the real
+# network. Kept separate from release.yml so the multi-hour run cannot block publishing.
+
+on:
+  release:
+    types: [published]
```
jozanek, line 9: `release: published` fires after the release ships, so a red run can't stop a bad one — and this is exactly the job that would catch #2464. Thanks for adding it; it's the first CI to touch this path.

Also no `timeout-minutes`: the awaits total 6h (1h plus the 5h at the it2 spec L57) against a 360-minute default, so a stall gets cancelled rather than reported. Would a nightly `schedule:` plus a timeout fit?

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -1147,8 +1147,22 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+        // until the snapshot is applied. Headers and snapshot-related types are still processed.
+        val utxoBootstrapInProgress =
+          settings.nodeSettings.utxoSettings.utxoBootstrap &&
+            !hr.isUtxoSnapshotApplied &&
+            modifierTypeId != Header.modifierTypeId &&
+            modifierTypeId != ManifestTypeId.value &&
+            modifierTypeId != UtxoSnapshotChunkTypeId.value
+
+        if (utxoBootstrapInProgress) {
+          log.debug(s"Ignoring ${invData.ids.length} non-tx invs (of type $modifierTypeId) from $peer: UTXO snapshot bootstrap in progress")
```
jozanek, line 1160: At debug level, a bootstrap suppressing invs looks identical to a healthy one at default logging — the same blind spot that made #2464 take two clean-state reproductions to diagnose.

Would one INFO heartbeat naming the phase and what's awaited (snapshot info / manifest / N of M chunks) be worth adding here?

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
```diff
@@ -591,8 +609,12 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
       case (stateId, Some(block), _) if stateId == block.id =>
         log.info(s"State and history have the same version ${encoder.encode(stateId)}, no recovery needed.")
         Success(stateIn)
+      case (_, None, _) if isPreparedUtxoSnapshotState(stateIn, history) =>
+        log.info(s"Prepared UTXO snapshot state ${encoder.encode(stateIn.version)} restored before the first full block")
+        Success(stateIn)
       case (_, None, _) =>
         log.info("State and history are inconsistent. History is empty on startup, rollback state to genesis.")
+        stateIn.closeStorage()
```
jozanek, line 617: Reading `recreatedState()` with `StoreRegistry` (`LDBFactory.scala:69-76`), this looks like a genuine fix — without it the reopen gets a cached handle to a just-deleted directory. Good catch.

It's unconditional though, so it changes every node reaching this branch, and I see no test on the plain non-bootstrap path. Could it be named in the description, or split out?

jozanek (commented):
## What I understood this to be for

With `blocksToKeep = 1440`, the first fresh header after the proof reaches `updateBestFullBlock`, which writes `minimalFullBlockHeight` — and since `isUtxoSnapshotApplied` is derived from that height (`UtxoSetSnapshotProcessor.scala:50`), it flips true with no snapshot applied. The snapshot request is then skipped forever and the node downloads full blocks from 1.8M against a genesis state, which matches the report.

If that's right, the five changes are one mechanism rather than a bundle: the suppression keeps `updateBestFullBlock` from poisoning the flag the snapshot path keys off, and `setHeadersChainSynced()` is its side-effect-free replacement. Worth stating in the body — it's what makes the shape defensible, and reviewers have to re-derive it today.

**Verdict: right change, wrong shape.**

## The blocking concern

@a-shannon's regression traces end to end, and I reached it independently. With `utxoBootstrap=true, nipopowBootstrap=false` — permitted by `ErgoSettingsReader.scala:191-194`, documented in README — the early return at `ToDownloadProcessor.scala:114` precedes L122-124, the only remaining writer of the flag. Neither new setter covers that config, so the node stalls permanently and a restart doesn't clear it. Two possible fixes inline. #2424 is rebased on this branch and carries the same guard, so fixing it here propagates.

## Questions I'm holding

**1. Does this close #2464, or a neighbour?** The issue reproduces from a wiped data dir; `UtxoBootstrapRestartSpecification`'s comment describes a restart mid-download. Untouched by the diff: the recurring `Extension is empty while comparison is younger`, `headersHeight` reporting lower than a just-applied proof's tip, the 52 → 11 peer collapse, and the quorum-without-fallback the reporter hypothesised. Would you keep it open pending a mainnet run?

**2. Regression, or never worked end to end?** `git diff v5.0.13..v6.1.3` over the two processor files shows only import moves — this deadlock has been identical since 2023, `it2` has never run in CI, and #1995 is still open. Your issue comment mentions instability "after some changes in p2p layer"; naming the commit would say whether a second regression is still out there.

**3. Why does `ErgoHistory.scala:303` require `nipopowBootstrap` when the other guards don't?** Nothing explains it, and that asymmetry is what makes the blocker unrecoverable.

## One note without a line to sit on

Flipping the flag early moves rejection of a section below `minimalFullBlockHeight` from rule 304 (`recoverable`, no penalty) to rule 305 (`fatal`, `penalizeMisbehavingPeer`). Unsolicited modifiers are filtered at `:1261`, so I don't think it's reachable today — but the flag now gates a download policy and a peer-penalising validation rule at once. Free while `minimalFullBlockHeight == 1`; not once it can be raised.

a-shannon:
@jozanek, @kushti_ru, thanks for the detailed review. The ordinary-header blocker and the two test issues are addressed in [#2545](https://github.com/ergoplatform/ergo/pull/2545), targeting this PR's `i2464` branch.

The fix calls `setHeadersChainSynced()` when an ordinary fresh header arrives, while preserving the retention floor and suppressing premature block-section downloads. The regressions drive normal header append without manually setting the flag, and close the original history store before reopening it and checking the next snapshot request. A mutation using `updateBestFullBlock` fails the retention test. Nine focused tests pass, with **8/8 CI** on the current head.

The existing restart predicate is unchanged in this follow-up. A persisted header alone does not establish freshness; ordinary mode resumes discovery when the next fresh header arrives. Immediate recovery without another header remains outside this evidence.

The same shared commits are included in [#2424](https://github.com/ergoplatform/ergo/pull/2424) and the [combined wallet review](https://github.com/a-shannon/ergo/pull/2), both also at 8/8. The [tracker](https://github.com/ergoplatform/ergo/issues/2533) links the review increments.

I would keep [#2464](https://github.com/ergoplatform/ergo/issues/2464) open pending the separate end-to-end bootstrap/restart validation. This follow-up does not resolve the other trust/configuration, workflow, observability or state-recreation coverage questions, or identify a separate P2P regression commit.

src/it2/scala/org/ergoplatform/it2/TestUtxoSnapshotBootstrapOnMainNetSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+    val preKillInfo = Await.result(Async.async {
+      var info = nodeInfoAfterHeaders
+      while (info.bestBlockHeightOpt.isEmpty &&
+             System.currentTimeMillis() - snapshotPhaseStart < 90.seconds.toMillis) {
+        Thread.sleep(5.seconds.toMillis)
+        info = Async.await(node.waitFor[NodeInfo](_.info, _ => true, 1.minute))
+      }
+      info
+    }, 5.minutes)
+    preKillInfo.bestBlockHeightOpt shouldBe empty // still no full blocks: killed mid bootstrap
```
jozanek, line 76: The 90s window ends in a hard assertion, so "the snapshot finished faster than expected" becomes a red test rather than a skipped sub-scenario. In the other direction, if manifest discovery takes most of the window the restart lands after headers but before any chunk — still exercising the fix, but not the mid-chunk restart the comment describes. `docs/review/testing.md` flags wall-clock dependence as a flake source.

Would branching instead of asserting work — kill and restart when `bestBlockHeightOpt` is empty, otherwise log and continue to phase 3?

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -294,6 +294,20 @@ object ErgoHistory extends ScorexLogging {
...
+    // If the node is restarted in the middle of UTXO set snapshot bootstrapping (headers are
+    // already persisted, but the snapshot was not applied yet), the headers chain is already
+    // complete - via a NiPoPoW proof or via ordinary header sync - so mark it synced. Otherwise
+    // the synchronizer never asks for the snapshot manifest / block sections
+    // (ErgoNodeViewSynchronizer.requestMoreModifiers) and bootstrapping can never resume.
+    // Only the in-memory flag is set; no heights are written here (isUtxoSnapshotApplied is
+    // derived from the persisted minimalFullBlockHeight, which must stay untouched until the
+    // snapshot is actually applied).
+    if (nodeSettings.utxoSettings.utxoBootstrap &&
+        history.bestHeaderOpt.isDefined &&
```
jozanek, line 306: I suggested dropping the `nipopowBootstrap` conjunct and you did — but @a-shannon's objection that a persisted header alone doesn't establish freshness holds up, and on a second look their position is the more precise one. With the `ToDownloadProcessor` fix in place, this startup set now buys about one block interval, not correctness.

The cost when headers are stale or partial: `processSnapshotsInfo` logs an error per manifest per peer against `bestHeaderAtHeight == None`, and header pipelining drops to the 5s sync schedule until the chain catches up. Would a freshness test satisfy both — `history.bestHeaderOpt.exists(_.isNew(chainSettings.blockInterval * headerChainDiff))` in place of the dropped conjunct? That keeps your restart case and excludes the stale one, with the `toDownload` fix as the safety net.

src/test/scala/org/ergoplatform/nodeView/history/UtxoBootstrapNoNipopowSyncSpecification.scala
```diff
@@ -0,0 +1,125 @@
...
+    // fresh headers drive the ordinary headers-synced transition via the append path
+    val headers = genHeaderChain(BlocksInChain, history, diffBitsOpt = None, useRealTs = true)
+    val updHistory = applyHeaderChain(history, headers)
+    updHistory.isHeadersChainSynced shouldBe true
+
+    // ...but the transition must not persist minimalFullBlockHeight: isUtxoSnapshotApplied is
+    // derived from it (readMinimalFullBlockHeight() > GenesisHeight), so writing e.g.
+    // header.height - blocksToKeep + 1 here (as updateBestFullBlock would) would flip it to
+    // true with no snapshot applied and the snapshot would be skipped forever
+    updHistory.minimalFullBlockHeight shouldBe GenesisHeight
```
jozanek, line 113: With `BlocksInChain = 10` and `BlocksToKeep = 11`, `updateBestFullBlock` would compute `max(readMinimalFullBlockHeight()=1, 10-11+1=0) = 1` — so this assertion holds even with the wrong fix. The property still fails on the parent via `isHeadersChainSynced shouldBe true`, so it's a real deadlock regression; it just doesn't yet guard the poisoning trap the comment describes.

Would `blocksToKeep = 2` (or a chain longer than `blocksToKeep + 1`) make it discriminate?

src/test/scala/org/ergoplatform/nodeView/history/UtxoBootstrapRestartSpecification.scala
```diff
@@ -0,0 +1,119 @@
...
+    updHistory.isUtxoSnapshotApplied shouldBe false
+    updHistory.isHeadersChainSynced shouldBe false // set only via updateBestFullBlock / setHeadersChainSynced
+
+    // simulate node restart: close the storage and reopen the same database.
+    // Without the close, StoreRegistry hands back the already-open underlying DB and the
+    // "restart" is not a real cold reopen.
+    updHistory.historyStorage.close()
+    val restarted = ErgoHistory.readOrGenerate(historySettings)(null)
+
+    restarted.headersHeight shouldBe updHistory.headersHeight
```
jozanek, line 106: This reads `updHistory.headersHeight` after the close on L103. It resolves today only because ~40 entries fit the test caches (`indexesCacheSize = 1000`, `headersCacheSize = 100`); a miss would reach `db.get` on a closed LevelDB, which has no `Try` around it.

Thanks for adding the close — the reopen is a genuine cold one now. Would capturing the height into a val before L103, and using the public `updHistory.closeStorage()`, keep it robust as the fixture grows?

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -111,6 +111,21 @@ trait ToDownloadProcessor
...
+      // until the snapshot has been applied. Block sections downloaded before the snapshot
+      // would be stored as non-best and never applied to the freshly recreated state.
+      // Headers-chain sync detection must still run here: otherwise, with nipopowBootstrap = false,
+      // `isHeadersChainSynced` is never set, `nextModifiersToDownload` stays empty and the
+      // snapshot is never requested. Only the in-memory flag may be set though: running
+      // `updateBestFullBlock` would persist minimalFullBlockHeight, and `isUtxoSnapshotApplied`
+      // is derived from it (readMinimalFullBlockHeight() > GenesisHeight), so with blocksToKeep >= 0
+      // it would flip to true with no snapshot applied and the snapshot would be skipped forever.
+      if (!isHeadersChainSynced && header.isNew(chainSettings.blockInterval * headerChainDiff)) {
+        setHeadersChainSynced()
```
jozanek, line 125: This predicate and its log line are now identical to the ones at L132-135. Two copies of the freshness rule can drift apart later.

Would extracting a small private helper — `markHeadersSyncedIfFresh(header)` — used by both branches be worth it here?

jozanek (approved):
## Re-review at `9dca805`

**The blocker is fixed, and fixed well.** `setHeadersChainSynced()` inside the bootstrap branch, not `updateBestFullBlock` — so `minimalFullBlockHeight` stays at `GenesisHeight` and `isUtxoSnapshotApplied` can't flip early. I traced `toDownload` → `nextModifiersToDownload` L94 → `requestSnapshotsInfo()` and it reaches the snapshot request for `utxoBootstrap=true, nipopowBootstrap=false`. `headerChainDiff` and `chainSettings` are in scope, `verifyTransactions=false` still short-circuits first, and the flag write is in-memory only.

Three regression tests now fail on the parent commit — the new `UtxoBootstrapNoNipopowSyncSpecification` property 1, the parameterised restart spec, and the corrected `UtxoBootstrapToDownloadSpecification`. None of them sets the flag by hand. The restart spec's `close()` does evict from `StoreRegistry`, so the reopen is a real cold one. Thank you also for putting the `minimalFullBlockHeight` derivation into the code comments — that was the part reviewers had to re-derive.

## From the previous review

Addressed: the blocker, the guard asymmetry, both test issues, and the it2 restart scenario.

Still open, all outside the blocker's path: the `genesisId` trust gate at `ErgoNodeViewHolder.scala:322`; digest mode falling through to the state-wiping branch at `:594`; the unconditional `closeStorage()` still unnamed in the description; the `bootstrap-mainnet.yml` post-release trigger and missing timeout; `NodeViewTestConfig.allConfigs` still carrying no bootstrap configuration. Also still open: whether this actually closes #2464 — the body still says it does, while @a-shannon and I both read the evidence as not yet establishing it.

## One correction to my own earlier advice

I suggested dropping the `nipopowBootstrap` conjunct from `ErgoHistory`, and you did. @a-shannon declined that same change, and on re-examination their reasoning is the more precise one: with the `toDownload` fix in place, the startup set buys about one block interval rather than correctness, and it now marks a stale or partial header chain as synced. A freshness predicate satisfies both readings — details inline.

## Not blocking

The new findings are all minor: one test assertion that is vacuous at the current constants, the staleness point above, a read-after-close in the restart spec, and a wall-clock window in the it2 spec. Two coverage notes without a line to sit on: the it2 spec keeps `nipopowBootstrap = true`, so the configuration that produced the blocker has no integration coverage — and `it2` isn't in `ci.yml`, so this spec wasn't exercised by the green CI on this head.

Approving on the blocker being resolved and the regression tests that now cover it. That approval is scoped to this commit: the `genesisId` gate and digest mode from the previous round are still unanswered, and I'd like them addressed or explicitly ruled out of scope — but they predate this commit and needn't hold it.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
jozanek: Following up, since `9dca805` focused elsewhere. The narrow version: with `nipopowBootstrap=false` there is no `genesisId` requirement, and `processNipopowProof` is not gated on it either — so an unsolicited proof can be applied and now also flips the sync flag.

If there is an upstream filter I have missed, saying so closes this. Otherwise gating this line on `nipopowBootstrap && utxoBootstrap`, or requiring `genesisId` whenever `utxoBootstrap` is set, would do it. Happy to resolve either way.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
jozanek: Following up — this one was not in `9dca805`. The ask is small: either reject `stateType = "digest"` together with `utxoBootstrap = true` in settings validation, or say here that digest is out of scope for snapshot bootstrap and I will resolve.

As it stands such a node reaches L615-618 and has its state directory recreated. That may well be acceptable; nothing currently says so.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
jozanek: Still open after `9dca805`. To be clear, this is a good fix — the ask is only that it be visible: a line in the PR description noting it changes startup recovery for every node reaching this branch, not just bootstrapping ones.

A test on the plain non-bootstrap path would be a bonus rather than a requirement. I am happy to resolve on the description alone.

.github/workflows/bootstrap-mainnet.yml
jozanek: Following up. This matters a little more after `9dca805`: the it2 spec now carries the restart scenario, and `it2` still is not in `ci.yml`, so nothing exercised it before the green checks on this head.

A nightly `schedule:` plus a `timeout-minutes` would turn it into a signal you can act on. If you would rather keep it manual-dispatch only for now, say so and I will resolve.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
jozanek: Partly addressed — the new `log.info` in `ToDownloadProcessor` covers the sync transition, which was the part I most wanted. Thank you.

What is still missing is the bootstrap phase itself: this line stays at `debug`, so a node waiting on manifests or chunks still looks idle at default logging. One INFO line naming the phase would close it. Not a blocker — I will resolve if you would rather hold the log volume down.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
jozanek: Correcting my follow-up above: I asked this as though nothing existed for it, but #2546 already addresses exactly this, and cites this thread. It also finds a second digest issue I had missed — the snapshot-chunk receive branch requires a local UTXO reader, so Digest nodes drop incoming chunks.

So this is no longer a question for #2496; it is a merge decision on #2546. Resolving here and leaving it there. Apologies for the noise.

a-shannon:
**Update, 17 September:** [cb406808](https://github.com/ergoplatform/ergo/commit/cb4068084d6dc4577cc3006fd649fbd42db7b04b) applies the proposed ordinary-path correction in [#2496](https://github.com/ergoplatform/ergo/pull/2496), preserving the pending-snapshot path. The exact increment is one line. [CI for the corrected head](https://github.com/ergoplatform/ergo/actions/runs/35234286931) is in progress; validation of the correction remains pending. The finding and failed tests below refer to the previous head `03aad1f4`.

@kushti_ru, the changes in [#2496](https://github.com/ergoplatform/ergo/pull/2496) address the ordinary-header snapshot transition and improve the restart coverage, but the helper extraction in `03aad1f4` introduced a regression in normal pruned bootstrap.

Both calls to [`markHeadersSyncedIfFresh`](https://github.com/ergoplatform/ergo/blob/03aad1f40fbb92d39e5ac6b62b37849955dfbc8d/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala#L124-L148) use the default `updateBestBlock = false`. That is necessary while a UTXO snapshot is pending, but the ordinary branch previously called `updateBestFullBlock(header)`. It now leaves `minimalFullBlockHeight` at genesis for a fresh pruned node. Download planning starts too early, and a supplied suffix cannot initialize the full-block chain at its intended starting height.

The current CI reproduces both consequences:

- [Node job](https://github.com/ergoplatform/ergo/actions/runs/35210483250/job/105216952591): 819 passed, 2 failed, 2 ignored. Both failures are `PrunedNodeViewHolderSpec` suffix-bootstrap cases, with the state version remaining zero.
- [Integration job](https://github.com/ergoplatform/ergo/actions/runs/35210483250/job/105216952310): 12 passed, 1 failed. `PrunedDigestNodeSyncSpec` fails at line 71 because the digest node applies a block below the expected pruning start, before the final pruning assertion.

Please pass `updateBestBlock = true` in the ordinary branch and keep it `false` in the pending-snapshot branch. Removing the default would make both choices explicit. The existing failing tests already cover the regression; the two snapshot-discovery properties should remain green too. I have not executed that proposed correction.

This should be fixed before merging. It affects pruned bootstrap, not a demonstrated consensus fault. No separate PR is needed. The earlier snapshot-discovery fix also overlaps [#2545](https://github.com/ergoplatform/ergo/pull/2545); its remaining test coverage can be reconciled after this branch is stable.

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
jozanek: The helper landed in `03aad1f4`, thanks — and I owe you one here. Extracting it was my suggestion and I filed it as a nit, which was wrong: the two branches were only *nearly* identical, differing exactly in `updateBestFullBlock` vs `setHeadersChainSynced`. That is the difference `03aad1f4` flattened and `cb40680` restored.

What is left is the `updateBestBlock = false` default: the pending-snapshot call site still takes it implicitly, so the value you get by saying nothing is the one that broke pruned bootstrap. @a-shannon suggested dropping the default so both choices are explicit, which looks right to me. Would making it a required parameter work? Keeping this thread open for that.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
jozanek: Withdrawing this one, with a correction to my own framing. I presented it as something this PR widens; on a closer read it mostly is not.

Proof application is guarded by `history().isEmpty`, so the window is a fresh node. And once an attacker-supplied proof has been applied, the header chain already *is* that chain — the `setHeadersChainSynced()` added here only makes the snapshot request happen about one block interval sooner, against headers that were already going to be used. The substantive gap is that `processNipopowProof` (dispatched at :1622) runs regardless of `nipopowBootstrap`, while `ErgoSettingsReader:195` requires `genesisId` only when that flag is set. That predates this PR and is unchanged by it.

So it belongs in its own issue rather than blocking this branch. Resolving here.

src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
jozanek: Closing this one — it was only ever a request for a line in the description, the code itself is right, and it should not keep a thread open on an approved PR. Resolving.

.github/workflows/bootstrap-mainnet.yml
kushti: Checking on release is ok for now maybe, more frequent checks could be too heavy . On my laptop this kind of bootstrap takes ~15 min, so default timeout should be ok

PR #2501, opened 2026-09-05
Restore Matrix CI dependencies and mining test fixtures

a-shannon:
Build the declared Sigma snapshot from a verified source commit for each consuming Scala version, preserve cross-version codec behavior, and make Matrix mining fixtures observe the state required by their assertions.

Current head: [5fe55bb2](https://github.com/a-shannon/ergo/commit/5fe55bb2e3a1af8b1a2b42a024f757f55ee3dcec), target `weak-blocks`. The [latest one-file increment](https://github.com/a-shannon/ergo/compare/aaa73480397590c9ae6c3e024f87af0e64972dc9...5fe55bb2e3a1af8b1a2b42a024f757f55ee3dcec), +8/-4, reuses the shared fixture follow-up [b43a21c4](https://github.com/a-shannon/ergo/commit/b43a21c49b827f7df11448767d28c2b3c42ab5c9) on [e3493d84](https://github.com/a-shannon/ergo/commit/e3493d84bcbe70147bcc6716c642df9a031a952a), owned here. All production, bootstrap, codec and other test files are unchanged.

The previous-candidate regression now waits until work names the exact applied bootstrap parent before forcing regeneration. A block-application notification can reach the test before the candidate generator updates its cache. The test retains its existing candidate deadline, real PoW, rejection against the current candidate and exact previous-block application assertions. The changed case passed **1/1 after normal compilation on this actual composition**. The source-equal fixture also passed the19-case candidate suite in the #2500 consumer; its initial18/19 timeout remains recorded. The exact historical cache/PoW interleaving was not established.

Prior head [aaa734803](https://github.com/a-shannon/ergo/commit/aaa73480397590c9ae6c3e024f87af0e64972dc9) passed [all eight native CI jobs](https://github.com/ergoplatform/ergo/actions/runs/34735098843). That result is historical; native CI for5fe55bb2 remains pending. The earlier [c33→aaa increment](https://github.com/a-shannon/ergo/compare/c33e35a73d7885b8049a434af04518b4b01f8e2e...aaa73480397590c9ae6c3e024f87af0e64972dc9) and its scoped local6+7 mining/forwarding and45+5+32 selection/soft-field/observation checks retain their evidence. The initial shared snapshots-file lock preceded the successful fresh-directory actor run and remains historical.

This PR owns [bootstrap S](https://github.com/a-shannon/ergo/commit/13352c1e52aad4b63a90e16a51c7ecafda2ea98f), [Scala compatibility C](https://github.com/a-shannon/ergo/commit/1bf12ba7881ee2f477eeaf2564c558bb2139bd1a), their [common merge](https://github.com/a-shannon/ergo/commit/6e9be2ac43216fab895f1a161f2927fca819e48a), and the mining fixtures. It reuses production [E from #2504](https://github.com/a-shannon/ergo/commit/e77f77c2130521b0b9ddd7594d9d11aa5c51a56b), [F from #2505](https://github.com/a-shannon/ergo/commit/13d79b04a9e1d9ca7dbe57b4f701158a35b3b06d), and the [#2535 U/H observer](https://github.com/a-shannon/ergo/commit/bcfcd61166151d8000ae2a05a62fa4f389aa7126). Review each prerequisite with its owner before integrating this cumulative composition; other consumers can reuse the small common commits without depending on this whole branch.

Bootstrap accepts only `6.0.2-30-59782d92-SNAPSHOT` and `6.0.5-22-368a860b-SNAPSHOT`, mapped to source [59782d92](https://github.com/ergoplatform/sigmastate-interpreter/commit/59782d92065eec3abd0dc47af22e25141857b2ee) and [368a860b](https://github.com/ergoplatform/sigmastate-interpreter/commit/368a860be033af94aa14895381f42099b3646db6). It verifies the declaration, fetched commit, local Ivy metadata and main/test JARs. The older source describes v6.0.2-25-g59782d92; publication under the existing count-30 declaration establishes provenance, not identity with the unavailable artifact. Jobs use Java8 and pinned setup actions. Explicit Circe constructors and existing Either syntax preserve JSON behavior on Scala2.11.

The mining fixtures select their protocol, separate block notifications from replies, await actor shutdown after failure and wait for actual transaction processing. Real previous-candidate proofs, version6 input-block/ordering flows, confirmed UTXO assertions, funded box generation and deterministic mempool admission remain intact. [#2312](https://github.com/ergoplatform/ergo/pull/2312) covers related acknowledgment/funding concerns; [#2478](https://github.com/ergoplatform/ergo/pull/2478) covers candidate expiry. Matrix selection/fees/input-state cache refresh belong to [#2504](https://github.com/ergoplatform/ergo/pull/2504), lifecycle coverage and forwarding to [#2505](https://github.com/ergoplatform/ergo/pull/2505).

Shared validation includes17 shell contracts, both pinned Sigma sources compiled/published on Scala2.12.20, the older source also on2.11.12, and scoped three-case codec checks. The historical [combined df2df798 candidate](https://github.com/a-shannon/ergo/commit/df2df798b590e8162718eb1cd939611d8f35ce9d) passed [eight CI jobs](https://github.com/ergoplatform/ergo/actions/runs/33997997102), including1,116 node tests,696 wallet/core tests and six Matrix lifecycle cases. That run does not qualify the new head or the full legacy Docker suite. This PR remains a draft for coordinated maintainer review.

### DIFF
--- a/.github/actions/bootstrap-sigmastate/action.yml
+++ b/.github/actions/bootstrap-sigmastate/action.yml
@@ -1,0 +1,95 @@
+name: Bootstrap pinned SigmaState
+description: Build the unavailable SigmaState snapshot from its exact source commit
+
+inputs:
+  scala-version:
+    description: Full Scala version used by the consuming Ergo job
+    required: true
+  ergo-build-file:
+    description: Ergo build file that declares the pinned SigmaState version
+    required: false
+    default: build.sbt
+
+runs:
+  using: composite
+  steps:
+    - name: Set up JDK 17 for SigmaState
+      uses: actions/setup-java@cf277c60eb25467037889841efdb72551f06f6c3 # v4
+      with:
+        distribution: temurin
+        java-version: '17'
+
+    - name: Set up sbt
+      uses: sbt/setup-sbt@7b0c62920c0c6e456e829dcaad762cff5048167f # v1
+
+    - name: Build and publish SigmaState locally
+      shell: bash
+      env:
+        ERGO_BUILD_FILE: ${{ inputs.ergo-build-file }}
+        SIGMASTATE_REPOSITORY: https://github.com/ergoplatform/sigmastate-interpreter.git
+        SIGMASTATE_COMMIT: 368a860be033af94aa14895381f42099b3646db6
+        SIGMASTATE_VERSION: 6.0.5-22-368a860b-SNAPSHOT
+        SIGMASTATE_SCALA_VERSION: ${{ inputs.scala-version }}
+      run: |
+        set -euo pipefail
+
+        unset GITHUB_TOKEN SONATYPE_USERNAME SONATYPE_PASSWORD PGP_PASSPHRASE
+
+        ergo_build="${GITHUB_WORKSPACE}/${ERGO_BUILD_FILE}"
+        if [[ ! -f "${ergo_build}" ]]; then
+          echo "::error::Ergo build file not found: ${ERGO_BUILD_FILE}"
+          exit 1
+        fi
+
+        declared_version="$(sed -nE 's/^[[:space:]]*val[[:space:]]+sigmaStateVersion[[:space:]]*=[[:space:]]*"([^"]+)".*$/\1/p' "${ergo_build}")"
+        if [[ "${declared_version}" != "${SIGMASTATE_VERSION}" ]]; then
+          echo "::error::Ergo declares SigmaState ${declared_version:-<missing>}, but the bootstrap pins ${SIGMASTATE_VERSION}"
+          exit 1
+        fi
+
+        case "${SIGMASTATE_SCALA_VERSION}" in
+          2.11.12|2.12.20|2.13.16) ;;
+          *)
+            echo "::error::Unsupported Scala version: ${SIGMASTATE_SCALA_VERSION}"
+            exit 1
+            ;;
+        esac
+
+        sigma_dir="$(mktemp -d "${RUNNER_TEMP}/sigmastate.XXXXXX")"
+        git -C "${sigma_dir}" init
+        git -C "${sigma_dir}" remote add origin "${SIGMASTATE_REPOSITORY}"
+        git -C "${sigma_dir}" fetch --no-tags --depth=1 origin "${SIGMASTATE_COMMIT}"
+        git -C "${sigma_dir}" checkout --detach FETCH_HEAD
+
+        actual_commit="$(git -C "${sigma_dir}" rev-parse HEAD)"
+        if [[ "${actual_commit}" != "${SIGMASTATE_COMMIT}" ]]; then
+          echo "::error::SigmaState checkout mismatch: expected ${SIGMASTATE_COMMIT}, got ${actual_commit}"
+          exit 1
+        fi
+
+        (
+          cd "${sigma_dir}"
+          sbt -batch \
+            "++${SIGMASTATE_SCALA_VERSION}!" \
+            "set ThisBuild / version := \"${SIGMASTATE_VERSION}\"" \
+            "show ThisBuild / version" \
+            "show scalaVersion" \
+            publishLocal
+        )
+
+        scala_binary="${SIGMASTATE_SCALA_VERSION%.*}"
+        ivy_dir="${HOME}/.ivy2/local/org.scorexfoundation/sigma-state_${scala_binary}/${SIGMASTATE_VERSION}"
+        ivy_file="${ivy_dir}/ivys/ivy.xml"
+        main_jar="${ivy_dir}/jars/sigma-state_${scala_binary}.jar"
+        tests_jar="${ivy_dir}/jars/sigma-state_${scala_binary}-tests.jar"
+
+        if [[ ! -s "${ivy_file}" || ! -s "${main_jar}" || ! -s "${tests_jar}" ]]; then
+          echo "::error::SigmaState local publication is missing its Ivy descriptor or JARs"
+          exit 1
+        fi
+
+        if ! grep -Fq "revision=\"${SIGMASTATE_VERSION}\"" "${ivy_file}" || \
+           ! grep -Fq "name=\"scala-library\" rev=\"${SIGMASTATE_SCALA_VERSION}\"" "${ivy_file}"; then
+          echo "::error::SigmaState local publication does not match the requested version and Scala toolchain"
+          exit 1
+        fi
--- a/.github/actions/bootstrap-sigmastate/action.yml
+++ b/.github/actions/bootstrap-sigmastate/action.yml
@@ -13,11 +13,11 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Set up JDK 17 for SigmaState
+    - name: Set up JDK 8 for SigmaState
       uses: actions/setup-java@cf277c60eb25467037889841efdb72551f06f6c3 # v4
       with:
         distribution: temurin
-        java-version: '17'
+        java-version: '8'
 
     - name: Set up sbt
       uses: sbt/setup-sbt@7b0c62920c0c6e456e829dcaad762cff5048167f # v1
--- a/.github/actions/bootstrap-sigmastate/action.yml
+++ b/.github/actions/bootstrap-sigmastate/action.yml
@@ -69,7 +69,7 @@ runs:
 
         (
           cd "${sigma_dir}"
-          sbt -batch \
+          sbt -batch -J-Xms512m -J-Xmx4g \
             "++${SIGMASTATE_SCALA_VERSION}!" \
             "set ThisBuild / version := \"${SIGMASTATE_VERSION}\"" \
             "show ThisBuild / version" \
--- a/.github/actions/bootstrap-sigmastate/action.yml
+++ b/.github/actions/bootstrap-sigmastate/action.yml
@@ -1,0 +1,102 @@
+name: Bootstrap pinned SigmaState
+description: Build the unavailable SigmaState snapshot from its exact source commit
+
+inputs:
+  scala-version:
+    description: Full Scala version used by the consuming Ergo job
+    required: true
+  ergo-build-file:
+    description: Ergo build file that declares the pinned SigmaState version
+    required: false
+    default: build.sbt
+
+runs:
+  using: composite
+  steps:
+    - name: Set up JDK 8 for SigmaState
+      uses: actions/setup-java@cf277c60eb25467037889841efdb72551f06f6c3 # v4
+      with:
+        distribution: temurin
+        java-version: '8'
+
+    - name: Set up sbt
+      uses: sbt/setup-sbt@7b0c62920c0c6e456e829dcaad762cff5048167f # v1
+
+    - name: Build and publish SigmaState locally
+      shell: bash
+      env:
+        ERGO_BUILD_FILE: ${{ inputs.ergo-build-file }}
+        SIGMASTATE_REPOSITORY: https://github.com/ergoplatform/sigmastate-interpreter.git
+        SIGMASTATE_SCALA_VERSION: ${{ inputs.scala-version }}
+      run: |
+        set -euo pipefail
+
+        unset GITHUB_TOKEN SONATYPE_USERNAME SONATYPE_PASSWORD PGP_PASSPHRASE
+
+        ergo_build="${GITHUB_WORKSPACE}/${ERGO_BUILD_FILE}"
+        if [[ ! -f "${ergo_build}" ]]; then
+          echo "::error::Ergo build file not found: ${ERGO_BUILD_FILE}"
+          exit 1
+        fi
+
+        declared_version="$(sed -nE 's/^[[:space:]]*val[[:space:]]+sigmaStateVersion[[:space:]]*=[[:space:]]*"([^"]+)".*$/\1/p' "${ergo_build}")"
+        case "${declared_version}" in
+          6.0.2-30-59782d92-SNAPSHOT)
+            SIGMASTATE_COMMIT=59782d92065eec3abd0dc47af22e25141857b2ee
+            ;;
+          6.0.5-22-368a860b-SNAPSHOT)
+            SIGMASTATE_COMMIT=368a860be033af94aa14895381f42099b3646db6
+            ;;
+          *)
+            echo "::error::Unsupported SigmaState declaration: ${declared_version:-<missing>}"
+            exit 1
+            ;;
+        esac
+        SIGMASTATE_VERSION="${declared_version}"
+
+        case "${SIGMASTATE_SCALA_VERSION}" in
+          2.11.12|2.12.20|2.13.16) ;;
+          *)
+            echo "::error::Unsupported Scala version: ${SIGMASTATE_SCALA_VERSION}"
+            exit 1
+            ;;
+        esac
+
+        sigma_dir="$(mktemp -d "${RUNNER_TEMP}/sigmastate.XXXXXX")"
+        git -C "${sigma_dir}" init
+        git -C "${sigma_dir}" remote add origin "${SIGMASTATE_REPOSITORY}"
+        git -C "${sigma_dir}" fetch --no-tags --depth=1 origin "${SIGMASTATE_COMMIT}"
+        git -C "${sigma_dir}" checkout --detach FETCH_HEAD
+
+        actual_commit="$(git -C "${sigma_dir}" rev-parse HEAD)"
+        if [[ "${actual_commit}" != "${SIGMASTATE_COMMIT}" ]]; then
+          echo "::error::SigmaState checkout mismatch: expected ${SIGMASTATE_COMMIT}, got ${actual_commit}"
+          exit 1
+        fi
+
+        (
+          cd "${sigma_dir}"
+          sbt -batch -J-Xms512m -J-Xmx4g \
+            "++${SIGMASTATE_SCALA_VERSION}!" \
+            "set ThisBuild / version := \"${SIGMASTATE_VERSION}\"" \
+            "show ThisBuild / version" \
+            "show scalaVersion" \
+            publishLocal
+        )
+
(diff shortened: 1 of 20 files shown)


.github/actions/bootstrap-sigmastate/action.yml
```diff
@@ -0,0 +1,102 @@
...
+          exit 1
+        fi
+
+        declared_version="$(sed -nE 's/^[[:space:]]*val[[:space:]]+sigmaStateVersion[[:space:]]*=[[:space:]]*"([^"]+)".*$/\1/p' "${ergo_build}")"
+        case "${declared_version}" in
+          6.0.2-30-59782d92-SNAPSHOT)
+            SIGMASTATE_COMMIT=59782d92065eec3abd0dc47af22e25141857b2ee
+            ;;
+          6.0.5-22-368a860b-SNAPSHOT)
+            SIGMASTATE_COMMIT=368a860be033af94aa14895381f42099b3646db6
```
cafebedouin, line 48: **[Maintainability] Make the version→commit table check itself.**

The build forces `ThisBuild / version`, which a shallow, tagless fetch requires. The pin is therefore only as correct as this table: a row mapping a version string to the wrong commit still passes the `actual_commit` check at :71.

Forcing is right. The distance segment cannot be derived: `git describe` gives `v6.0.2-25-g59782d920` for the commit the first row names `6.0.2-30-…`. The 8-hex segment is the only part of the string that identifies the source.

Recommended: after the `case`, fail unless `SIGMASTATE_COMMIT` starts with the hash embedded in `declared_version`. Run as a standalone script, this passes both rows and rejects a swapped row and a release string such as `6.0.6`:

```bash
embedded="$(sed -nE 's/^[0-9]+(\.[0-9]+)*-[0-9]+-([0-9a-f]{8})-SNAPSHOT$/\2/p' <<<"${declared_version}")"
if [[ -z "${embedded}" || "${SIGMASTATE_COMMIT}" != "${embedded}"* ]]; then
  echo "::error::SigmaState pin mismatch: ${declared_version} does not embed the prefix of ${SIGMASTATE_COMMIT}"
  exit 1
fi
```

.github/actions/bootstrap-sigmastate/action.yml
```diff
@@ -0,0 +1,102 @@
...
+
+        ergo_build="${GITHUB_WORKSPACE}/${ERGO_BUILD_FILE}"
+        if [[ ! -f "${ergo_build}" ]]; then
+          echo "::error::Ergo build file not found: ${ERGO_BUILD_FILE}"
+          exit 1
+        fi
+
+        declared_version="$(sed -nE 's/^[[:space:]]*val[[:space:]]+sigmaStateVersion[[:space:]]*=[[:space:]]*"([^"]+)".*$/\1/p' "${ergo_build}")"
+        case "${declared_version}" in
+          6.0.2-30-59782d92-SNAPSHOT)
```
cafebedouin, line 44: **[Nit] The `6.0.2-30-59782d92-SNAPSHOT` row is not exercised by this PR's CI.**

All eight jobs of run 34740376766 resolve `6.0.5-22-368a860b-SNAPSHOT`. `build.sbt` was read on `weak-blocks`, `master` (`6.0.6`) and `v6.0.3-RC1` (`6.0.3`): none declares the `6.0.2-30` string. Code search covers default branches only, so another branch may still declare it; that was not checked.

Recommended: name the consuming branch in a comment beside the row, or drop the row until a branch needs it. Which tree declares it?

cafebedouin (commented):
Review carried out by Claude (Anthropic, Claude Fable 5.1) for cafebedouin. Taken from the #2533 review table — Build/CI unit.

**Commits:** `13352c1e5` (pinned Sigma bootstrap), then `1bf12ba78` (Scala 2.11 codec compatibility), reconciled with the head `5fe55bb2e`.
**Scope:** `.github/actions/bootstrap-sigmastate/action.yml`, `.github/workflows/ci.yml`, the two `AutolykosSolution` codec lines and their spec. The mining fixtures and the other commits of this PR were not reviewed.
**Answer to the unit's question — do source pins, declared snapshots and each consuming Scala version agree? Yes, for the entry this branch consumes.**

| | declared / allowed | checked against |
|---|---|---|
| Sigma snapshot | `build.sbt` at the head and on `weak-blocks`: `6.0.5-22-368a860b-SNAPSHOT` | `git describe` of `368a860be` in sigmastate-interpreter: `v6.0.5-22-g368a860be` |
| Scala versions | `build.sbt` `scala211/212/213` = 2.11.12 / 2.12.20 / 2.13.16 | CI matrices = the same three; the action's allow-list = the same three |
| it job | checkout under `it/` | `uses: ./it/.github/actions/…` and `ergo-build-file: it/build.sbt` both follow it |
| cache | `~/.ivy2/cache` is cached, `~/.ivy2/local` is not | so a stale locally-published Sigma cannot be restored |

**Tests run:** no local build. All eight job logs of the head's CI run [34740376766](https://github.com/ergoplatform/ergo/actions/runs/34740376766) were read: every job checked out `368a860`, printed `ThisBuild / version 6.0.5-22-368a860b-SNAPSHOT`, printed its own matrix Scala version for every Sigma module (2.11.12 ×2, 2.12.20 ×4, 2.13.16 ×2), published `sigma-state_<binary>/6.0.5-22-368a860b-SNAPSHOT`, and then passed its tests (57 / 178 / 1077 / 45). The check proposed inline was run as a standalone script on both table rows and on two wrong ones.

**Findings:** two inline, both non-blocking, both on the version table. Nothing found in `1bf12ba78`: `Encoder.instance` / `Decoder.instance` is the right fix for Scala 2.11 (no SAM conversion), and the new spec pins exact JSON fields against independently written expectations.

**Two notes, no action needed now:**
- `master` is on Sigma `6.0.6` (a release) and Scala `2.13.18`. When `weak-blocks` catches up, the action fails closed on the Scala allow-list with a clear message — the right behaviour, just expect it.
- Sigma is rebuilt in all eight jobs (~3 min each). Not caching `~/.ivy2/local` is what keeps stale artifacts out; if a cache is added later, key it by the Sigma commit and the Scala version.

.github/actions/bootstrap-sigmastate/action.yml
cafebedouin: Correction to the snippet above: its pattern is too strict. It rejects legitimate sbt-dynver strings whose tag carries a suffix, such as `6.1.0-RC2-5-abcdef12-SNAPSHOT`. It fails closed, so it never passes a wrong row, but it would reject a valid one. Anchoring only the tail is enough:

```bash
embedded="$(sed -nE 's/^.*-([0-9a-f]{8})-SNAPSHOT$/\1/p' <<<"${declared_version}")"
```

Run on seven strings: both table rows and two suffixed tags pass; a swapped row, `6.0.6` and a 7-character hash are rejected.

Also, the 25-versus-30 distance cited above restates what the PR description already says. It was not a new observation.

.github/actions/bootstrap-sigmastate/action.yml
cafebedouin: Correction: this question is withdrawn. The PR description already answers it, and it was not read before this review was posted. Its fourth paragraph says the older source describes `v6.0.2-25-g59782d92`, that the count-30 declaration is kept for provenance, and that both pinned sources were compiled and published on Scala 2.12.20, the older one also on 2.11.12.

What remains is narrower: that row is exercised by the local validation, not by this PR's CI jobs. No action needed.

cafebedouin:
Correction to the review above, first of its two closing notes: it names the wrong gate. With `master`'s `6.0.6` declared, the action stops at the version `case` (`Unsupported SigmaState declaration`, `action.yml:43-54`) before it reaches the Scala allow-list at `:58`, and that allow-list tests the matrix input, not `build.sbt`. Failing closed is still the behaviour; the message will be the version one.

Also, "~3 min each" rests on one timed job (169 s). Two corrections are posted inline as replies.

.github/workflows/ci.yml
```diff
@@ -47,6 +42,16 @@ jobs:
               ~/Library/Caches/Coursier/v1
           key: ${{ runner.os }}-sbt-cache-v4-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
 
+      - name: Build pinned SigmaState dependency
+        uses: ./.github/actions/bootstrap-sigmastate
+        with:
+          scala-version: ${{ matrix.scala }}
+
+      - name: Setup Java and Scala
+        uses: olafurpg/setup-scala@v10
```
cafebedouin, line 51: **[Maintainability] Each job builds Sigma under one JDK and tests the node under another.**

Noticed after the review above. In the 2.11.12 core job of run 34740376766, the bootstrap's SHA-pinned `setup-java` gives `sbt 1.12.11 (Temurin Java 1.8.0_504)`. This step then downloads jabba 0.11.2, installs `adopt@1.8.0-292` and switches `JAVA_HOME`, so the tests report `sbt 1.11.1 (AdoptOpenJDK Java 1.8.0_292)`: a 2021 build of a discontinued distribution, fetched by an archived action pinned by tag.

The step predates this PR; the PR moves it after the bootstrap. The description's "pinned setup actions" holds for the composite action, not for this step, nor for `checkout@v4` and `cache@v4`.

Recommended: drop this step and let the jobs keep the Temurin 8 and sbt the action already installs, or pin it by SHA. Not run here.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -884,18 +889,20 @@ object CandidateGenerator extends ScorexLogging {
     ).headOption
   }
 
+  val MaxFeeBoxesPerTransaction: Int = 100
```
jozanek: #2462 (open, base `master`) defines this same constant with the same value and the same `.grouped(...)` shape for issue #2185. If both land, `master` and `weak-blocks` end up with two implementations to reconcile at merge time. Would you be open to settling with that PR which one is canonical? I may be missing a reason they are intentionally separate.

.github/actions/bootstrap-sigmastate/action.yml
```diff
@@ -0,0 +1,108 @@
...
+            SIGMASTATE_COMMIT=59782d92065eec3abd0dc47af22e25141857b2ee
+            ;;
+          6.0.5-22-368a860b-SNAPSHOT)
+            SIGMASTATE_COMMIT=368a860be033af94aa14895381f42099b3646db6
+            ;;
+          *)
```
jozanek, line 50: An unlisted `sigmaStateVersion` exits 1 here, so a branch declaring a *published* version — `master` has `6.0.6` — fails rather than skipping the bootstrap. That makes the action hard to reuse outside `weak-blocks`, and puts every cross-branch merge inside this table. Would a no-op path when the version is not a pinned snapshot work? `build.sbt:50` already reads `SIGMASTATE_VERSION` from the env.

.github/workflows/ci.yml
```diff
@@ -47,6 +42,16 @@ jobs:
               ~/Library/Caches/Coursier/v1
           key: ${{ runner.os }}-sbt-cache-v4-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
 
+      - name: Build pinned SigmaState dependency
```
jozanek, line 45: This step's `publishLocal` writes to `~/.ivy2/local` (`action.yml:94`), and the cache above lists `~/.ivy2/cache`, so I think Sigma gets rebuilt from source in all 8 matrix cells on every run. Adding `~/.ivy2/local` to the cache paths, keyed on the pinned commit, would cut most of that. Have I read the paths right?

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -1040,37 +1029,150 @@ object CandidateGenerator extends ScorexLogging {
                   maxBlockSize: Int,
                   us: UtxoStateReader,
                   upcomingContext: ErgoStateContext,
-                  transactions: Seq[ErgoTransaction]
+                  transactions: Seq[ErgoTransaction],
+                  inputBlockTransactions: Seq[ErgoTransaction] = Seq.empty
```
jozanek: Both production callers (`:723`, `:818`) pass this argument — the default serves the 12 test call sites. The description records that omitting it on the retry path already rejected a valid dependent transaction. Would you be open to making it required and updating the tests, so that failure mode can't recur silently?

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -1040,37 +1029,150 @@ object CandidateGenerator extends ScorexLogging {
...
-    var remainingTxs = transactions
+    val version = upcomingContext.sigmaPreHeader.version
+    val prefixIds = inputBlockTransactions.map(_.id).toSet
+    val prefixState = us.withTransactions(inputBlockTransactions)
+    val prefix = inputBlockTransactions.map { tx =>
+      val cost = prefixState.validateWithCost(tx, upcomingContext,
+        upcomingContext.currentParameters.maxBlockCost, Some(verifier), softFieldsAllowed = true).get
+      tx -> cost
+    }
+    require(correctLimits(prefix, upcomingContext.currentParameters.maxBlockCost, maxBlockSize, version),
```
jozanek: A failing `require` here reaches `createCandidate`'s `Try`, which logs and stops without scheduling a retry — unlike the limit checks below, which set `done = true` and truncate. Since the prefix comes from persisted input-block state, I'm not sure the condition clears on its own. Could this degrade by truncating the prefix instead?

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -1040,37 +1029,150 @@ object CandidateGenerator extends ScorexLogging {
...
+      size
+    }
+
+    def correctLimits(blockTxs: Seq[CostedTransaction], costLimit: Long,
+                      sizeLimit: Long, blockVersion: Header.Version): Boolean = {
+      if (blockTxs.map(_._2.toLong).sum > costLimit) false
+      else if (blockTxs.isEmpty) true
+      else {
+        // Match BlockTransactionsSerializer, including its version marker and count.
+        val framing = 32L + unsignedSize(blockTxs.size.toLong) +
```
jozanek: This framing mirrors `BlockTransactionsSerializer.serialize`, but nothing binds the two, so a later wire change would silently mis-size candidates. Would a shared `sizeOf(txs, blockVersion)` on the serializer be workable? Separately, `correctLimits` moved from a top-level `private def` into this closure, so it can no longer be tested on its own.

jozanek (commented):
What I understood this PR to be for: unbreaking `weak-blocks` CI, which has failed 8/8 runs since 2026-05-22 because `6.0.5-22-368a860b-SNAPSHOT` resolves nowhere. On that it delivers — 7/8 jobs green from a base where nothing passes. Thanks for the candour about what isn't verified.

What I'd most like your read on: the chunked `collectFees` and `MaxFeeBoxesPerTransaction = 100` look near-identical to #2462, which implements the same issue (#2185) against `master`. Two in flight will collide when `weak-blocks` merges — is one meant to supersede the other? Relatedly, `CandidateGenerator.scala` differs from #2504's copy by ~164 lines, while `AutolykosSolution`, `ErgoTransaction` and `ErgoMiner` are byte-identical across #2501/#2504/#2505.

What would move me to approve: splitting the ~194 lines that actually fix CI from the candidate-assembly work. I ran no builds, and couldn't verify the `ForkResolutionSpec` timeout — no log artifact.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -1040,37 +1029,150 @@ object CandidateGenerator extends ScorexLogging {
...
+    val prefix = inputBlockTransactions.map { tx =>
+      val cost = prefixState.validateWithCost(tx, upcomingContext,
+        upcomingContext.currentParameters.maxBlockCost, Some(verifier), softFieldsAllowed = true).get
+      tx -> cost
+    }
+    require(correctLimits(prefix, upcomingContext.currentParameters.maxBlockCost, maxBlockSize, version),
+      "Selected input transactions exceed the ordering block limits")
+    // A mandatory prefix can consume the conservative mining cost gap. Keep that valid prefix,
+    // but give optional transactions no additional cost allowance in that case.
+    val orderingCostLimit = math.max(maxBlockCost.toLong, prefix.map(_._2.toLong).sum)
```
jozanek: Thanks for making this explicit in the comment. Raising the ceiling to the prefix's own cost does mean a peer-supplied prefix can consume the `safeGap` margin, which exists to absorb interpreter cost-estimation differences. What's the reasoning that this is safe here? No action needed if the gap only guards pre-5.0 AOT costing.

ergo-core/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala
```diff
@@ -493,6 +494,20 @@ case class ErgoTransaction(override val inputs: IndexedSeq[Input],
...
+    def unapply(error: Throwable): Option[SoftFieldAccessException] = {
+      @scala.annotation.tailrec
+      def unwrap(current: Throwable, remaining: Int): Option[SoftFieldAccessException] = current match {
+        case cause: SoftFieldAccessException => Some(cause)
+        case wrapper: InvocationTargetException if remaining > 0 =>
+          unwrap(wrapper.getTargetException, remaining - 1)
+        case _ => None
+      }
+      // Only reflection wrappers are transparent; unrelated failures keep their normal rejection path.
+      unwrap(error, 16)
```
jozanek: This lands in `ergo-core`, which SPV clients consume, and it widens what becomes `SoftFieldsAccessError` — which then drives the input-vs-ordering split. Is an upstream fix in sigma, rethrowing unwrapped, on the table? I'm also curious what 16 levels protects against; the reflective call looked like it wraps once.

kushti (approved):

PR #2505, opened 2026-09-05
Exercise Matrix lifecycle on three nodes and fix API dispatch

a-shannon:
Exercise Matrix lifecycle on three devnet nodes through the normal mining REST API: input-block wallet and mempool updates, competing input chains, restart recovery, and payment fees reaching the ordering miner and becoming spendable after maturity. This PR also owns the typed SolutionFound forwarding and full-block route termination needed by that API path.

Head `be808056e9c842fcab537dab67036e1e4d8582c2` targets `weak-blocks`. The [new nineteen-file prerequisite increment](https://github.com/a-shannon/ergo/compare/6a346b38a519d3e9a1c705658047e4cc4db1218d...28bf47722c18054b6b516fb2c634238a8ba5de95) preserves the six lifecycle scenarios and their assertions. It reuses existing production corrections in six paths plus their direct and shared tests.

The fee and selected-input-state prerequisites are owned by [#2504](https://github.com/ergoplatform/ergo/pull/2504): [E](https://github.com/a-shannon/ergo/commit/e77f77c2130521b0b9ddd7594d9d11aa5c51a56b) restores fee collection, prefix-aware transaction selection and soft-field classification; [3c709](https://github.com/a-shannon/ergo/commit/3c70954de416d9e3d0d8b4245754d47eff94ce6a) keys cached mining work by selected input tip and processed transaction prefix; its exact [3fd test follow-up](https://github.com/a-shannon/ergo/commit/3fd1ef21d8e7028c3401a3eb75566421a83ae61f) retains the independently allocated digest assertion. These production inputs now appear in this PR's actual ancestry. Exact [b43 mining fixtures](https://github.com/a-shannon/ergo/commit/b43a21c49b827f7df11448767d28c2b3c42ab5c9), owned by [#2501](https://github.com/ergoplatform/ergo/pull/2501), retain actor ownership and wait for the applied parent before testing a previous solution. Forwarding [F](https://github.com/a-shannon/ergo/commit/13d79b04a9e1d9ca7dbe57b4f701158a35b3b06d) remains owned here and unchanged.

Synchronization uses exact [G](https://github.com/a-shannon/ergo/commit/d62c945a4d6314c54a7b57f18a94aee8fd258aa7), owned by [#2511](https://github.com/ergoplatform/ergo/pull/2511), on [R](https://github.com/a-shannon/ergo/commit/76e5ec4bfc377334fdced4bd60a61df06e6b4cb6), owned by [#2535](https://github.com/ergoplatform/ergo/pull/2535), and [X](https://github.com/a-shannon/ergo/commit/a772a5cbf666584db3ed981838dbc11223d502ed), owned by [#2549](https://github.com/ergoplatform/ergo/pull/2549). G corrects full-summary genesis anchors, tip/mode cache identity and retained-common-ancestor download scheduling; X wakes cached sections after their headers arrive. The shared [test compatibility follow-up](https://github.com/a-shannon/ergo/commit/ec9c33682f3cd07266ef6c4c4cd1a763f71200d0) preserves the parser bound of 50 and all serialization round trips on branches where the constant is private. Review R before G and merge the owning production prerequisites before this lifecycle consumer, using these focused commits rather than whole owner branches.

The existing [#2502 prerequisite 518d8bea](https://github.com/a-shannon/ergo/commit/518d8beaef3130ad337d577cce7630f85476f383) remains unchanged: available-prefix selection, disjoint rollback, wallet rollback independent of cache availability, and pooled-conflict cleanup. Existing [D](https://github.com/a-shannon/ergo/commit/71f33c6e45a9394557cf3af1bfb422cbc53d26cf) and [H](https://github.com/a-shannon/ergo/commit/0f9f75c1fc1df3ff3309482b83c0475a05cdd7a2), owned by #2535, preserve DeepRollBack's settled seed, isolation, final oracle, twenty-minute bound and restored connectivity. The [S/C bootstrap increment](https://github.com/a-shannon/ergo/compare/9d02fae38927ed255a2d971d3f8beae7822f8aeb...f960fa5f27c6142930f90131fbf54de821b3e5c5), owned by #2501, is unchanged and retains the declared pinned Sigma coordinate and codec behavior.

At `6a346b38`, [native IT](https://github.com/ergoplatform/ergo/actions/runs/34738411580/job/103673822249) passed 50/52 with one ignored case. The fee lifecycle failed because the payment transaction was absent from the ordering block, before the fee-collector assertions; DeepRollBack separately stalled at tips 246/70 after a settled 20/20 seed, with connected peers. The old cache omitted selected input state, and fee collection still returned None. Their existing owner corrections are therefore required by the retained scenario. G/X address separate synchronization defects; no exclusive attribution of the old CI failures is claimed. The old node log also contained candidate failures; no final node failure count is inferred from its partial, repeating log.

The exact [P proof-test prerequisite](https://github.com/a-shannon/ergo/commit/79531dcf145ac1b18ca17f61c7634baf6c53edea) adds an ordering-only test profile and a deterministic regression for continuing past an input-block hit in the same nonce window. It changes one core test file, without changing production mining or weakening proof validation.

The preceding composition at `28bf4772` passed normal compilation and 178/178 focused checks: full weak synchronizer 88, V2 cache/history/scheduling 21, UTXO/Digest wake-up 2, fee/transaction selection 45, mining cache 11, soft-field validation 5, the DevNet60 input-to-ordering actor case 1 in a fresh JVM, and all five Scala 2.12 core proof tests after adding P. The 173 earlier checks remain valid because P changes only that core test file. Independent review covered the weak reader fixture and exact fee/cache/mining joins. Prior unchanged #2502/H/D composition evidence was 106/106. Those local tests did not run Docker or establish full CI success; new-head native CI remains pending in [#2533](https://github.com/ergoplatform/ergo/issues/2533).

The six owner scenarios retain exact transaction, wallet-balance, box, UTXO, mempool-conflict and ordering-progress assertions. They cover normal inclusion, competing chains, settled and pending-payment restarts, local confirmed-state restoration before connecting peers, and mature fee spending. Pending-payment recovery checks one inclusion and absence from the next ordering block. Fee checks follow exact fee-box consumption on all three nodes, the ordering miner key, reward value and creation height, emission separately, scan movement at maturity, and spending with a normal funding input. Premature-spend rejection remains in collector properties. Harness configuration, image selection and restart-preserving data binds retain existing defaults and normal genesis/PoW producers.

Fork and pending-payment recovery resume input production after reconnection. They do not prove idle reconnection or abrupt-crash recovery; offline restart does not assert persistence of pending input blocks or the mempool. Historical [combined CI](https://github.com/a-shannon/ergo/actions/runs/33997997102) at [df2df798](https://github.com/a-shannon/ergo/commit/df2df798b590e8162718eb1cd939611d8f35ce9d) passed eight jobs, including 1,116 node tests, 696 wallet/core tests and six lifecycle scenarios, on a combined #2500-#2505 candidate. That result is not this new head's CI.

With a matching combined node image, run `ERGO_MATRIX_TEST_IMAGE=<node image ID> sbt "it:testOnly org.ergoplatform.it.MatrixLifecycleSpec"`; the image variable is optional for the default tag. This remains a draft for coordinated review.

The [new three-file test increment](https://github.com/a-shannon/ergo/compare/28bf47722c18054b6b516fb2c634238a8ba5de95...be808056e9c842fcab537dab67036e1e4d8582c2) merges exact shared [M/funding fixtures](https://github.com/a-shannon/ergo/commit/29e6b6c641c6b2f8e179a9a596fc566c494a8a7d); P is already present. Review the common fixture commit once. The six Matrix lifecycle scenarios, fee-flow assertions and forwarding production correction remain unchanged. Normal Scala 2.12 compilation and 16/16 affected checks passed on this exact tree: mempool clearing 13, funding 2 and the original large-mempool property 1. Each group used one test task in a fresh JVM and data directory. Other earlier results are input-qualified. At the preceding public head, native integration had 46 passed, zero failed and one ignored case, but MatrixLifecycleSpec aborted at startup with an address-already-in-use error before its cases executed. The funding fixture does not fix that Docker startup failure or establish the six scenarios on this head. New-head native CI must supply that evidence; this remains a draft.

### DIFF
--- a/src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
+++ b/src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
@@ -2,7 +2,7 @@ package org.ergoplatform.mining
 
 import akka.actor.{Actor, ActorRef, ActorRefFactory, Props, Stash}
 import akka.pattern.StatusReply
-import org.ergoplatform.AutolykosSolution
+import org.ergoplatform.{AutolykosSolution, SolutionFound}
 import org.ergoplatform.mining.CandidateGenerator.GenerateCandidate
 import org.ergoplatform.nodeView.state.DigestState
 import org.ergoplatform.nodeView.ErgoNodeViewHolder.ReceivableMessages.GetDataFromCurrentView
@@ -191,6 +191,9 @@ class ErgoMiner(
       minerState.candidateGeneratorRef forward genCandidate
 
     case solution: AutolykosSolution =>
+      minerState.candidateGeneratorRef forward solution
+
+    case solution: SolutionFound =>
       minerState.candidateGeneratorRef forward solution
 
     case ReadMinerPk => // used in /mining/rewardAddress API method
--- a/.github/actions/bootstrap-sigmastate/action.yml
+++ b/.github/actions/bootstrap-sigmastate/action.yml
@@ -1,0 +1,102 @@
+name: Bootstrap pinned SigmaState
+description: Build the unavailable SigmaState snapshot from its exact source commit
+
+inputs:
+  scala-version:
+    description: Full Scala version used by the consuming Ergo job
+    required: true
+  ergo-build-file:
+    description: Ergo build file that declares the pinned SigmaState version
+    required: false
+    default: build.sbt
+
+runs:
+  using: composite
+  steps:
+    - name: Set up JDK 8 for SigmaState
+      uses: actions/setup-java@cf277c60eb25467037889841efdb72551f06f6c3 # v4
+      with:
+        distribution: temurin
+        java-version: '8'
+
+    - name: Set up sbt
+      uses: sbt/setup-sbt@7b0c62920c0c6e456e829dcaad762cff5048167f # v1
+
+    - name: Build and publish SigmaState locally
+      shell: bash
+      env:
+        ERGO_BUILD_FILE: ${{ inputs.ergo-build-file }}
+        SIGMASTATE_REPOSITORY: https://github.com/ergoplatform/sigmastate-interpreter.git
+        SIGMASTATE_SCALA_VERSION: ${{ inputs.scala-version }}
+      run: |
+        set -euo pipefail
+
+        unset GITHUB_TOKEN SONATYPE_USERNAME SONATYPE_PASSWORD PGP_PASSPHRASE
+
+        ergo_build="${GITHUB_WORKSPACE}/${ERGO_BUILD_FILE}"
+        if [[ ! -f "${ergo_build}" ]]; then
+          echo "::error::Ergo build file not found: ${ERGO_BUILD_FILE}"
+          exit 1
+        fi
+
+        declared_version="$(sed -nE 's/^[[:space:]]*val[[:space:]]+sigmaStateVersion[[:space:]]*=[[:space:]]*"([^"]+)".*$/\1/p' "${ergo_build}")"
+        case "${declared_version}" in
+          6.0.2-30-59782d92-SNAPSHOT)
+            SIGMASTATE_COMMIT=59782d92065eec3abd0dc47af22e25141857b2ee
+            ;;
+          6.0.5-22-368a860b-SNAPSHOT)
+            SIGMASTATE_COMMIT=368a860be033af94aa14895381f42099b3646db6
+            ;;
+          *)
+            echo "::error::Unsupported SigmaState declaration: ${declared_version:-<missing>}"
+            exit 1
+            ;;
+        esac
+        SIGMASTATE_VERSION="${declared_version}"
+
+        case "${SIGMASTATE_SCALA_VERSION}" in
+          2.11.12|2.12.20|2.13.16) ;;
+          *)
+            echo "::error::Unsupported Scala version: ${SIGMASTATE_SCALA_VERSION}"
+            exit 1
+            ;;
+        esac
+
+        sigma_dir="$(mktemp -d "${RUNNER_TEMP}/sigmastate.XXXXXX")"
+        git -C "${sigma_dir}" init
+        git -C "${sigma_dir}" remote add origin "${SIGMASTATE_REPOSITORY}"
+        git -C "${sigma_dir}" fetch --no-tags --depth=1 origin "${SIGMASTATE_COMMIT}"
+        git -C "${sigma_dir}" checkout --detach FETCH_HEAD
+
+        actual_commit="$(git -C "${sigma_dir}" rev-parse HEAD)"
+        if [[ "${actual_commit}" != "${SIGMASTATE_COMMIT}" ]]; then
+          echo "::error::SigmaState checkout mismatch: expected ${SIGMASTATE_COMMIT}, got ${actual_commit}"
+          exit 1
+        fi
+
+        (
+          cd "${sigma_dir}"
+          sbt -batch -J-Xms512m -J-Xmx4g \
+            "++${SIGMASTATE_SCALA_VERSION}!" \
+            "set ThisBuild / version := \"${SIGMASTATE_VERSION}\"" \
+            "show ThisBuild / version" \
+            "show scalaVersion" \
+            publishLocal
+        )
+
+        scala_binary="${SIGMASTATE_SCALA_VERSION%.*}"
+        ivy_dir="${HOME}/.ivy2/local/org.scorexfoundation/sigma-state_${scala_binary}/${SIGMASTATE_VERSION}"
+        ivy_file="${ivy_dir}/ivys/ivy.xml"
+        main_jar="${ivy_dir}/jars/sigma-state_${scala_binary}.jar"
+        tests_jar="${ivy_dir}/jars/sigma-state_${scala_binary}-tests.jar"
+
+        if [[ ! -s "${ivy_file}" || ! -s "${main_jar}" || ! -s "${tests_jar}" ]]; then
+          echo "::error::SigmaState local publication is missing its Ivy descriptor or JARs"
+          exit 1
+        fi
+
+        if ! grep -Fq "revision=\"${SIGMASTATE_VERSION}\"" "${ivy_file}" || \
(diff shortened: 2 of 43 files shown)


src/main/scala/org/ergoplatform/mining/ErgoMiner.scala
```diff
@@ -193,6 +193,9 @@ class ErgoMiner(
     case solution: AutolykosSolution =>
       minerState.candidateGeneratorRef forward solution
 
+    case solution: SolutionFound =>
+      minerState.candidateGeneratorRef forward solution
```
cafebedouin, line 197: **[Possible bug] A forwarded solution that misses the guard gets no reply.**

The defect is in `CandidateGenerator.scala`, outside this PR's diff hunks, so the note sits on the arm that routes REST solutions into it.

The processing arm is `case sf: SolutionFound if state.solvedBlock.isEmpty && state.cachedCandidate.nonEmpty` (`:238-239`). The fallback that answers "Block already solved" is still `case _: AutolykosSolution` (`:298`), and that receive ends at `:302` with no catch-all. A `SolutionFound` that fails the guard matches nothing: no reply, and the route's ask waits out its timeout. The guard fails once `solvedBlock` is set, and also after a successful input block, which clears `cachedCandidate` at `:284` without setting `solvedBlock`.

Read at this head, not run.

Recommended: keep the raw arm, which `ErgoMiner.scala:193` still feeds, and add one for the typed message with a reason that is true in both cases. Not compiled here:

```scala
case _: SolutionFound =>
  val why = state.solvedBlock.map(b => s"Block already solved : ${b.id}").getOrElse("No cached candidate")
  sender() ! StatusReply.error(why)
```

cafebedouin (commented):
Review carried out by Claude (Anthropic, Claude Fable 5.1) for cafebedouin, with independent passes by Grok (grok-4.7-build) and Gemini (3.8 Flash). Taken from the #2533 review table — Scala/Akka, the typed-forwarding unit of #2505.

**Commit:** `13d79b04a` (`ErgoMiner.scala` +4/−1, `ErgoMinerSolutionForwardingSpec.scala` +70), reconciled with the head `be808056e`.
**Scope:** the new `SolutionFound` arm in `ErgoMiner.started` and its spec, plus the two ends of that path: `MiningApiRoute.scala:79-93` and the `SolutionFound` arm of `CandidateGenerator`. The REST-route correction, the lifecycle scenarios and the shared prerequisites were not reviewed.
**Answer to the unit's question — are both typed variants forwarded with the original sender, without changing proof acceptance? Forwarded with the original sender: yes, once the miner is in `started`. Acceptance code: untouched — but REST solutions now reach it, so its two gaps below become reachable from the API.**

- `SolutionFound` is a sealed trait with exactly two cases, `InputSolutionFound` and `OrderingSolutionFound` (`mining.scala:27-33`), and the spec sends both.
- The new arm uses `forward`, so `CandidateGenerator` sees the route's `askWithStatus` sender and replies to it directly. The spec asserts `candidateGenerator.lastSender shouldBe requester.ref`, then that the probe's reply reaches the requester.
- This commit does not edit `CandidateGenerator`, and the message is forwarded unmodified.
- Before it, a `SolutionFound` reaching `started` fell to `case m => log.warn("Unexpected message …")` and was never answered. With `useExternalMiner` on, the route's ask then timed out; with it off, the route fails at once and never asks. Without the new arm the spec's `expectMsg` times out. By reading; the revert was not executed here.
- One limit, unchanged by this commit: in `starting` (`ErgoMiner.scala:133-184`) `GenerateCandidate` gets "Miner has not started yet", but a `SolutionFound` matches nothing and is dropped without a reply.

**Tests run:** no local build. The description still lists new-head CI as pending; the tracker links run [34750648241](https://github.com/ergoplatform/ergo/actions/runs/34750648241) at this head, and its node job's log shows both `ErgoMinerSolutionForwardingSpec` cases passing, within 1127 succeeded / 0 failed. The spec stubs the generator with a probe, so that run says nothing about the inline finding.

**Findings:** two, both in `CandidateGenerator.scala` and both outside this PR's diff hunks: what a forwarded solution meets at the other end. The first is inline on the forwarding arm; the second follows.

**Finding 2 — [Possible bug] The previous-candidate fallback is not validated on this branch.** For an ordering solution, `CandidateGenerator.scala:253-262` passes the current candidate's block through `.filter { powScheme.validate(block.header).isSuccess }`, but the `.orElse` to `cachedPreviousCandidate` has no such filter. A solution that fails against the current candidate is therefore accepted whenever a previous candidate exists: the block is sent to the node view, `solvedBlock` is set and the sender gets success. `master` validates whichever block it picked (`CandidateGenerator.scala:264`), and #2411 keeps that when it removes the duplicate call, with tests for the fallback. This is already on `weak-blocks`; the forwarding arm is what lets a REST solution reach it. Read, not run.

Recommended: apply the same `validate` filter to the fallback before `sendOrderingToNodeView`; #2411 shows the shape. Not compiled here:

```scala
.orElse {
  state.cachedPreviousCandidate
    .map(c => c -> completeOrderingBlock(c.candidateBlock, solution))
    .filter { case (_, block) => ergoSettings.chainSettings.powScheme.validate(block.header).isSuccess }
}
```

arkadianet:
The `pathEndOrSingleSlash` change to `getFullBlockByHeaderIdR` in f5d322a90 is the same one-line
fix we reached for the input-block transaction routes. We carried your exact line, co-authored to
you, on a small standalone branch so it could be reviewed apart from the mining changes. Your
route spec already covers ids, bodies and trailing-slash full-block lookup, so we won't duplicate
those. One behaviour it doesn't pin: an unknown suffix after a full-block id
(`GET /blocks/{headerId}/thisDoesNotExist`) used to return 200 with the full block, and now falls
through. Through the node's `ApiRejectionHandler` it currently answers 500
(`MethodRejection(POST)`, because the handler has no case for it), which is a pre-existing handler
gap, not caused by this hunk. Happy to add that test to your branch with attribution. We can
assert only "not 200 and not the full block", or also pin the current 500, whichever you prefer.

PR #2506, opened 2026-09-06
Replay processed input tips during peer sync

a-shannon:
Replay the processed input tip during periodic sync and sync replies so peers reconnecting at the same ordering height can discover that input chain while mining is idle. Replay retains the existing version, UTXO, height-proximity and compact weak-ID rules, and excludes an announced descendant whose transactions are still pending. [#2500](https://github.com/ergoplatform/ergo/pull/2500) and [#2503](https://github.com/ergoplatform/ergo/pull/2503) retain the adjacent request-admission and connection-state scopes.

Head `0d7b983061780bf7720e762b09d1b603328d02f6` targets `weak-blocks` at `c216c5b6e3310911d5f3822b40c5e57eee6fc45a`. The [new increment](https://github.com/a-shannon/ergo/compare/89bb84847223d57e0234e590384c19650408b084...0d7b983061780bf7720e762b09d1b603328d02f6) preserves this PR's replay production and tests. It reuses selection [E](https://github.com/a-shannon/ergo/commit/e77f77c2130521b0b9ddd7594d9d11aa5c51a56b), owned by [#2504](https://github.com/ergoplatform/ergo/pull/2504); typed-solution forwarding [F](https://github.com/a-shannon/ergo/commit/13d79b04a9e1d9ca7dbe57b4f701158a35b3b06d), owned by [#2505](https://github.com/ergoplatform/ergo/pull/2505); and mining fixtures [b43](https://github.com/a-shannon/ergo/commit/b43a21c49b827f7df11448767d28c2b3c42ab5c9), owned by [#2501](https://github.com/ergoplatform/ergo/pull/2501). The [focused fixture closure](https://github.com/a-shannon/ergo/commit/29e6b6c641c6b2f8e179a9a596fc566c494a8a7d) combines exact #2500 ordering-proof support with #2501 mempool-admission and funded-box support. Review these shared commits with their owners before this consumer; their whole branches are not prerequisites.

The final composition passed normal node compilation and **16/16 focused checks**: two deterministic funding regressions, the actual large-mempool property, and all thirteen mempool cases. Its five core proof checks passed on unchanged core inputs. Earlier in this update, five owner replay cases and a DevNet mining case passed, covering equal-height discovery, outbound timestamps, processed-tip selection and both compact-payload boundaries. Those earlier node sessions used the previous generator and are not presented as final-input results.

The prior native node job failed in old miner/candidate fixtures and mempool admission setup. The added funding prerequisite also addresses a subsequently reproduced underfunded-box premise in the shared large-mempool test. Assertions still require valid PoW, transaction admission, removal, rollback, value conservation and token preservation. Independent source review found no composition blocker. These focused results do not establish one cause for every historical failure or replace new-head native CI.

Existing [H/K observer support](https://github.com/a-shannon/ergo/compare/612c7eadf7698f8aaba6ce78a63433c046704ed2...89bb84847223d57e0234e590384c19650408b084), owned by #2535/#2511, is unchanged, including fixed targets, progress latches and the fifteen-minute fork deadline. Its earlier actual composition passed normal IT compilation and 37/37 observer/diagnostic cases. At `612c7ead`, native IT passed twelve cases with one fork timeout and one ignored case; a separate core failure occurred during tool download before project tests. Local observer checks did not run the Docker scenario.

The earlier [S/C increment](https://github.com/a-shannon/ergo/compare/89cc88cedb0f3280feb74136da6d2853f0b20473...612c7eadf7698f8aaba6ce78a63433c046704ed2), owned by #2501, preserves the declared Sigma coordinate, verified source and codec expressions. Shared bootstrap evidence includes seventeen shell contracts, both pinned sources on Scala 2.12.20, the older source on 2.11.12, and three codec cases on each relevant old/new composition.

Historical owner evidence includes thirty direct actor properties and [combined Linux node CI](https://github.com/a-shannon/ergo/actions/runs/34001725870): 1,146 tests across 105 suites, with two ignored. [Final lifecycle CI](https://github.com/a-shannon/ergo/actions/runs/34010622069) passed eight scenarios, including idle reconnection, a nonproducing relay, pending-input restart and offline catch-up, before further mining and with full serialized-body comparisons. Runtime sources matched node validation; idle reconnection failed twice on the baseline. These results apply to the combined #2505 harness/#2501 bootstrap stack. Separate Windows runs had a fork timeout and a native JVM error. The [test-only follow-up](https://github.com/a-shannon/ergo/commit/c8680dda90d0583fd6d90e2450befb1501eb3389) retains the Linux-validated test files and cancels dependent scenarios after the first failure. Current individual-PR CI and integration remain tracked in [#2533](https://github.com/ergoplatform/ergo/issues/2533).

### DIFF
--- a/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
+++ b/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
@@ -382,6 +382,32 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
         val v2SyncInfo = getV2SyncInfo(history, full = true)
         networkControllerRef ! SendToNetwork(Message(syncInfoSpec, Right(v2SyncInfo), None), SendToPeers(peersV2))
       }
+      sendProcessedInputTip(history, peers)
+    }
+  }
+
+  /** Share the processed input tip during ordinary sync, including when production is idle. */
+  private def sendProcessedInputTip(history: ErgoHistory, peers: Seq[ConnectedPeer]): Unit = {
+    val localHeight = history.fullBlockHeight
+    val recipients = peers.filter { peer =>
+      localHeight > 0 &&
+        SubBlocksFilter.condition(peer) &&
+        peer.mode.exists(_.stateType == StateType.Utxo) &&
+        syncTracker.statuses.get(peer).exists { status =>
+          status.status != Unknown && status.status != Nonsense && status.height > 0 &&
+            math.abs(status.height.toLong - localHeight.toLong) <= 2
+        }
+    }
+    if (recipients.nonEmpty) {
+      // The announced tip may still await transactions. Only replay the processed prefix.
+      history.bestInputBlocksChain().headOption.flatMap(history.getInputBlock).foreach { tip =>
+        val announcement = if (tip.weakTxIds.getOrElse(Seq.empty).size <= 3) tip
+          else tip.copy(weakTxIds = None)
+        val message = Message(InputBlockMessageSpec, Right(announcement), None)
+        recipients.foreach { peer =>
+          networkControllerRef ! SendToNetwork(message, SendToPeer(peer))
+        }
+      }
     }
   }
 
@@ -469,6 +495,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     if (syncSendNeeded) {
       val ownSyncInfo = getV1SyncInfo(hr)
       sendSyncToPeer(remote, ownSyncInfo)
+      sendProcessedInputTip(hr, Seq(remote))
     }
   }
 
@@ -515,6 +542,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     if (syncSendNeeded) {
       val ownSyncInfo = getV2SyncInfo(hr, full = true)
       sendSyncToPeer(remote, ownSyncInfo)
+      sendProcessedInputTip(hr, Seq(remote))
     }
   }
 
--- a/.github/actions/bootstrap-sigmastate/action.yml
+++ b/.github/actions/bootstrap-sigmastate/action.yml
@@ -1,0 +1,102 @@
+name: Bootstrap pinned SigmaState
+description: Build the unavailable SigmaState snapshot from its exact source commit
+
+inputs:
+  scala-version:
+    description: Full Scala version used by the consuming Ergo job
+    required: true
+  ergo-build-file:
+    description: Ergo build file that declares the pinned SigmaState version
+    required: false
+    default: build.sbt
+
+runs:
+  using: composite
+  steps:
+    - name: Set up JDK 8 for SigmaState
+      uses: actions/setup-java@cf277c60eb25467037889841efdb72551f06f6c3 # v4
+      with:
+        distribution: temurin
+        java-version: '8'
+
+    - name: Set up sbt
+      uses: sbt/setup-sbt@7b0c62920c0c6e456e829dcaad762cff5048167f # v1
+
+    - name: Build and publish SigmaState locally
+      shell: bash
+      env:
+        ERGO_BUILD_FILE: ${{ inputs.ergo-build-file }}
+        SIGMASTATE_REPOSITORY: https://github.com/ergoplatform/sigmastate-interpreter.git
+        SIGMASTATE_SCALA_VERSION: ${{ inputs.scala-version }}
+      run: |
+        set -euo pipefail
+
+        unset GITHUB_TOKEN SONATYPE_USERNAME SONATYPE_PASSWORD PGP_PASSPHRASE
+
+        ergo_build="${GITHUB_WORKSPACE}/${ERGO_BUILD_FILE}"
+        if [[ ! -f "${ergo_build}" ]]; then
+          echo "::error::Ergo build file not found: ${ERGO_BUILD_FILE}"
+          exit 1
+        fi
+
+        declared_version="$(sed -nE 's/^[[:space:]]*val[[:space:]]+sigmaStateVersion[[:space:]]*=[[:space:]]*"([^"]+)".*$/\1/p' "${ergo_build}")"
+        case "${declared_version}" in
+          6.0.2-30-59782d92-SNAPSHOT)
+            SIGMASTATE_COMMIT=59782d92065eec3abd0dc47af22e25141857b2ee
+            ;;
+          6.0.5-22-368a860b-SNAPSHOT)
+            SIGMASTATE_COMMIT=368a860be033af94aa14895381f42099b3646db6
+            ;;
+          *)
+            echo "::error::Unsupported SigmaState declaration: ${declared_version:-<missing>}"
+            exit 1
+            ;;
+        esac
+        SIGMASTATE_VERSION="${declared_version}"
+
+        case "${SIGMASTATE_SCALA_VERSION}" in
+          2.11.12|2.12.20|2.13.16) ;;
+          *)
+            echo "::error::Unsupported Scala version: ${SIGMASTATE_SCALA_VERSION}"
+            exit 1
+            ;;
+        esac
+
+        sigma_dir="$(mktemp -d "${RUNNER_TEMP}/sigmastate.XXXXXX")"
+        git -C "${sigma_dir}" init
+        git -C "${sigma_dir}" remote add origin "${SIGMASTATE_REPOSITORY}"
+        git -C "${sigma_dir}" fetch --no-tags --depth=1 origin "${SIGMASTATE_COMMIT}"
+        git -C "${sigma_dir}" checkout --detach FETCH_HEAD
+
+        actual_commit="$(git -C "${sigma_dir}" rev-parse HEAD)"
+        if [[ "${actual_commit}" != "${SIGMASTATE_COMMIT}" ]]; then
+          echo "::error::SigmaState checkout mismatch: expected ${SIGMASTATE_COMMIT}, got ${actual_commit}"
+          exit 1
+        fi
+
+        (
+          cd "${sigma_dir}"
+          sbt -batch -J-Xms512m -J-Xmx4g \
+            "++${SIGMASTATE_SCALA_VERSION}!" \
+            "set ThisBuild / version := \"${SIGMASTATE_VERSION}\"" \
+            "show ThisBuild / version" \
+            "show scalaVersion" \
+            publishLocal
+        )
+
+        scala_binary="${SIGMASTATE_SCALA_VERSION%.*}"
+        ivy_dir="${HOME}/.ivy2/local/org.scorexfoundation/sigma-state_${scala_binary}/${SIGMASTATE_VERSION}"
+        ivy_file="${ivy_dir}/ivys/ivy.xml"
+        main_jar="${ivy_dir}/jars/sigma-state_${scala_binary}.jar"
+        tests_jar="${ivy_dir}/jars/sigma-state_${scala_binary}-tests.jar"
+
+        if [[ ! -s "${ivy_file}" || ! -s "${main_jar}" || ! -s "${tests_jar}" ]]; then
+          echo "::error::SigmaState local publication is missing its Ivy descriptor or JARs"
+          exit 1
+        fi
+
+        if ! grep -Fq "revision=\"${SIGMASTATE_VERSION}\"" "${ivy_file}" || \
+           ! grep -Fq "name=\"scala-library\" rev=\"${SIGMASTATE_SCALA_VERSION}\"" "${ivy_file}"; then
+          echo "::error::SigmaState local publication does not match the requested version and Scala toolchain"
+          exit 1
+        fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,11 +30,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Java and Scala
-        uses: olafurpg/setup-scala@v10
-        with:
-          java-version: ${{ matrix.java }}
-
       - name: Cache sbt
         uses: actions/cache@v4
         with:
@@ -46,6 +41,16 @@ jobs:
               ~/AppData/Local/Coursier/Cache/v1
               ~/Library/Caches/Coursier/v1
           key: ${{ runner.os }}-sbt-cache-v4-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
+
+      - name: Build pinned SigmaState dependency
+        uses: ./.github/actions/bootstrap-sigmastate
+        with:
+          scala-version: ${{ matrix.scala }}
+
+      - name: Setup Java and Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: ${{ matrix.java }}
 
       - name: Runs wallet tests
         run: sbt ++${{ matrix.scala }} ergoWallet/test
@@ -73,11 +78,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Java and Scala
-        uses: olafurpg/setup-scala@v10
-        with:
-          java-version: ${{ matrix.java }}
-
       - name: Cache sbt
         uses: actions/cache@v4
         with:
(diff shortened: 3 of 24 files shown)


cafebedouin (commented):
Review carried out by Claude (Anthropic), model Claude Opus 5, for cafebedouin.

**No blocking issue at `d5df48ce95246fde9d7d85761d56db062f2337a2`. The replay sends only the processed input tip, and each of eight mutated guards and hooks is caught by a test. The node-job mining failures also occur on the base.**

Scope: the `ErgoNodeViewSynchronizer.scala` change (+28) and `MatrixInputTipSyncSpec`, against the #2533 lane question "replay of the processed tip without pending descendants".

- **Processed tip only (read).** `bestInputBlocksChain()` is `bestChain.reverse`, and `bestChain` is `chain.take(processedIndex + 1)` (`InputBlocksProcessor.scala`:334-339, 1028-1035). So `headOption` is the last processed input block. `InputBlocksTree.bestTip` (:323) would be the wrong accessor: it returns the best fork's last block, processed or not.
- **Tests (executed locally at the head, JDK 8, sbt 1.11.1, Scala 2.12.20).** `sbt "testOnly org.ergoplatform.network.MatrixInputTipSyncSpec"`: 30/30 pass. Each of eight single mutations fails at least one test (table below). This is a local run, not CI.
- **Receiver side (read).** A peer that already has the input block returns at `ErgoNodeViewSynchronizer.scala`:1509-1512 ("already known"), so each replay costs that peer one lookup.
- **Mining failures in the node job (executed).** The base `b5a02204d`'s own CI node job ([run 35735669473](https://github.com/ergoplatform/ergo/actions/runs/35735669473)) failed 23 mining tests, including all six in this head's partial log. Locally, `sbt "testOnly org.ergoplatform.mining.ErgoMinerSpec org.ergoplatform.mining.CandidateGeneratorSpec"` fails 19 of 27 at both the head and the base.

Recommended: track the mining failures against `weak-blocks` rather than this PR. From this scope, nothing here blocks marking it ready.

Not run: the Docker integration suite, so there is no evidence either way on the `ForkResolutionSpec` timeout. Also not run: Scala versions other than 2.12.

<details><summary>Mutations (each applied alone to src/main; all compiled)</summary>

| mutation | tests failing |
|---|---|
| replay disabled (`if (false && recipients.nonEmpty)`) | 11 |
| announced tip: `inputBlocksTree().flatMap(_.bestTip)` instead of `bestInputBlocksChain().headOption` | 1: "an announced but unprocessed child does not replace the replayed processed tip" |
| height window `<= 2` → `<= 1000` | 4 |
| UTXO-mode peer filter removed | 7 |
| `SubBlocksFilter.condition(peer)` removed | 3 |
| weak-ID compaction removed (always `tip`) | 1: the 4-transaction case |
| V1 reply hook removed | 3 |
| V2 reply hook removed | 1 |

The `status != Unknown`, `!= Nonsense`, `height > 0` and `localHeight > 0` guards were not mutated. Patches are available on request.

</details>

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -382,6 +382,32 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
         val v2SyncInfo = getV2SyncInfo(history, full = true)
         networkControllerRef ! SendToNetwork(Message(syncInfoSpec, Right(v2SyncInfo), None), SendToPeers(peersV2))
       }
+      sendProcessedInputTip(history, peers)
```
jozanek: The stated case is a reconnecting peer, and the reply hooks at 498/545 already fire on its first sync. This periodic call also re-sends full announcements of tips mined by other nodes every 20–60 s, which brings back the non-local relay that 546eee98e removed (`// todo: send only id out`, :2338). Would you be open to dropping this call, or checking with kushti that relay on a timer is wanted?

jozanek (commented):
I read this as: a peer that reconnects while no input block is being mined should still learn the processed input chain. Replaying the processed prefix rather than `bestTip` looks right to me.

What I'm holding:
- **Recovery.** A replayed tip needs the receiver to back-fill parents. `applyDisconnected` (InputBlocksProcessor.scala:380–396) matches against `acc`, not `a`, so with 2+ missing ancestors I think the tip stays disconnected, and re-replays stop at "already known" (:854). Could a receiver test cover 3+ blocks?
- **CI.** IT passed at `b5a02204d` and `90a129733` but `ForkResolutionSpec` timed out here; its phase 4 reconnects UTXO nodes on forks. Could you rebase and rerun IT a few times?
- **Who.** With ~4 s input blocks on mainnet, who waits long enough to notice?

A recovery test and green IT would move me toward approval.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -382,6 +382,32 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+  /** Share the processed input tip during ordinary sync, including when production is idle. */
+  private def sendProcessedInputTip(history: ErgoHistory, peers: Seq[ConnectedPeer]): Unit = {
+    val localHeight = history.fullBlockHeight
+    val recipients = peers.filter { peer =>
+      localHeight > 0 &&
+        SubBlocksFilter.condition(peer) &&
+        peer.mode.exists(_.stateType == StateType.Utxo) &&
+        syncTracker.statuses.get(peer).exists { status =>
+          status.status != Unknown && status.status != Nonsense && status.height > 0 &&
+            math.abs(status.height.toLong - localHeight.toLong) <= 2
```
jozanek: The receiver acts on this tip only when its full-block height equals ours; at the other four offsets it drops it with an info log, and at −1 it also requests a header. Line 398 compares the peer's header height with our full-block height and admits Fork/Older peers. Would limiting recipients to `Equal` status keep sends to the peers that can use them?

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -382,6 +382,32 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
...
+        }
+    }
+    if (recipients.nonEmpty) {
+      // The announced tip may still await transactions. Only replay the processed prefix.
+      history.bestInputBlocksChain().headOption.flatMap(history.getInputBlock).foreach { tip =>
+        val announcement = if (tip.weakTxIds.getOrElse(Seq.empty).size <= 3) tip
+          else tip.copy(weakTxIds = None)
```
jozanek: This compaction and the recipient filter above copy `NewBestInputBlock` (:2321–2334), and the two filters already differ (the Unknown/Nonsense and `height > 0` checks exist only here). Would you be open to one shared helper so the two push paths stay in step?

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
```diff
@@ -469,6 +495,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     if (syncSendNeeded) {
       val ownSyncInfo = getV1SyncInfo(hr)
       sendSyncToPeer(remote, ownSyncInfo)
+      sendProcessedInputTip(hr, Seq(remote))
```
jozanek: As I read `ErgoSyncTracker.scala:91`, a V1 sync keeps the height from an earlier V2 one, so this fires only for a peer that sends V2 and then V1, yet peers passing `SubBlocksFilter` (≥6.5.0) send V2. Is there a real peer on this path? If not, would you consider dropping the hook?

jozanek (commented):
Thanks for the rework in b5c7a3a21. From my first round, the periodic relay, the V1 hook and the recipient window are resolved, and IT passed here. The ForkResolution timeout on #2560, which leaves sync untouched, suggests flakiness. The back-fill stall now lives in #2560, and I agree it should land first.

One question stays open, not blocking: with ~4 s input blocks on mainnet and no relay of non-local blocks, who gains most from a one-shot replay at reconnect: quiet devnets/testnets, or peers not connected to a miner?

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
jozanek: Resolved in b5c7a3a21: the periodic call is gone and replay fires only from the V2 reply. Thank you.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
jozanek: Resolved: `Equal` plus `status.height == localHeight` targets the peers that can use the tip, and the scaladoc names the header-vs-full-state limit. Thank you.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
jozanek: The filters now differ on purpose, so I'm dropping the helper suggestion. The ≤3 compaction (:404 and :2380) is still a copy; sharing it is optional from my side.

src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
jozanek: Resolved: the V1 hook is gone, and the V1 case now asserts no replay. Thank you.

cafebedouin:
Review carried out by Claude (Anthropic, Claude Opus 5.5) for cafebedouin, using forkbench; executed one added unit property, a variant of the change, and two- and three-node networks, merge base `90a12973` against `b5c7a3a21`. Adds only what is new since the 2026-09-22 review of `d5df48ce9`, and answers the open question from the 2026-09-23 review.

**The replay is not one-shot at reconnect: an already Equal peer receives the processed tip again on the next sync reply sent after a 60 s gap.**

**[Likely bug] Replay at most once per connection, not on every reply that `syncSendNeeded` requires**

Observed: `sendProcessedInputTip` runs whenever `syncSendNeeded` is true (`ErgoNodeViewSynchronizer.scala:539-542`), which includes `notSyncedOrOutdated(peer)` (`ErgoSyncTracker.scala:78`), true once this node's last sync send to the peer is older than one minute (`ErgoSyncTracker.scala:21`, `:57`). The class doc scopes replay "to the first V2 response which establishes an equal peer" (`MatrixInputTipSyncSpec.scala:34`).

Executed (JDK 8): an added property (replay once, `markOutdated(peer)`, a second equal V2 sync) passes 15 of 15: status stays `Equal` and the tip is announced again. With a per-connection flag in place, the PR's 14 tests still pass and the added property fails (`MatrixInputTipSyncSpec.scala:135`). On two nodes (600 s per jar, JDK 21), the follower sent the miner's own input blocks back to it 4 times on this branch and 0 on the merge base, each with its sync reply after an Equal exchange. Each repeat costs one message and one lookup, and brings back the timed re-send raised in the first review round.

Recommended: keep a per-peer "replayed" set, added to on the first replay and cleared on disconnect next to `clearStatus` (`ErgoNodeViewSynchronizer.scala:2101`), and keep the outdated-Equal case as a negative test. If recurring replay is intended, the class doc and description should say so.

**On who gains: neither option, as measured.** On a miner–relay–leaf line (no miner–leaf link; relay–leaf cut for longer than 60 s, then restored), the leaf never held the miner's current input tip on either build: 0 of 36 samples in each of the four valid runs. On the merge base no input block reached the leaf. On this branch the leaf got one replayed tip after the link returned, already 8 input blocks behind the miner, and the next ordering block replaced it. Quiet networks were not run.

Recommended: scope the stated benefit to directly connected peers.

<details><summary>Runs and counts</summary>

Unit (JDK 8, sbt 1.11.1), `MatrixInputTipSyncSpec` at `b5c7a3a21` plus the property below:
```
[info] - review-only: an Equal peer's next V2 sync replays again once the last local sync send is 60 s old
[info] Tests: succeeded 15, failed 0, canceled 0, ignored 0, pending 0

Same, with the per-connection variant (patch below) applied to src/main:
[info] - review-only: an Equal peer's next V2 sync replays again once the last local sync send is 60 s old *** FAILED ***
[info]   List() had size 0 instead of expected size 1 (MatrixInputTipSyncSpec.scala:135)
[info] Tests: succeeded 14, failed 1, canceled 0, ignored 0, pending 0
```

Two nodes (miner and follower), one 600 s run per jar, JDK 21. Both runs end with the same header chain and state root on both nodes. Counts from the node logs:
```
jar                    ordering  input blocks    follower's input-block  "already known" lines  after an Equal     during ordering-block catch-up
                       blocks    miner/follower  messages sent           (both nodes)           exchange           (present on both jars)
merge base 90a12973    60        695/686         0                       3                      0                  3
this branch            65        696/604         4                       24                     6 (+1 at 270 ms)   17
```
The 6 include the miner answering with its own tip twice. All 4 follower sends fall in 18:16-18:19; later the follower's replies came less than 60 s apart. The follower's lower input-block count on this branch is one run and is not attributed.

Miner's log on this branch (trimmed; each id was mined by the miner itself):
```
18:16:19.033 Input block 9fc72e04… already known, ignoring
18:17:19.267 Input block 36bc9afb… already known, ignoring
18:18:19.515 Input block 6280b603… already known, ignoring
18:19:44.847 Input block 3339a8cb… already known, ignoring
```

Both jars on one network: merge base mining with this branch following, and the reverse; in both, each node follows the other (one run per role).

Three nodes, miner M, relay R (peer: M), leaf L (peer: R only), one jar per run, JDK 21; settle, cut R–L for longer than the 60 s window while M mines, restore, observe L; L's best input tip sampled every 5 s. Four valid runs; two further runs on this branch are excluded because the relay itself held the miner's tip in under half of its samples.
```
run                 mean input-block interval  L tip = M tip  R->L input-block messages
merge base, run 1   3.46 s                     0 of 36        0
merge base, run 2   3.65 s                     0 of 36        0
merge base, run 3   0.53 s                     0 of 36        0
this branch         3.46 s                     0 of 36        2 (1 replay after the link returned, 8 input blocks behind M; 1 parent L requested)
```
One host, no link latency, one relay hop, coinbase-only blocks. The ~3.5 s interval comes from lowering the miner's polling interval.

jars (sha256): merge base `ae8b58581d84a2dd0d44ef91355aaadeeb1d9094fe8a7f0e23fb2fd441624774`; this branch (b5c7a3a21's tree applied to 90a12973) `7504be802220873910fdddddb720884b956bc5c03a5d396e4f7bb1e17cbe39ea`.

The added property, verbatim:
```scala
 
  property("review-only: an Equal peer's next V2 sync replays again once the last local sync send is 60 s old") {
    withFixture { f =>
      val tip = f.processedTip()
      val peer = f.peer()
      f.sync.inboundV2(f.history, peer)
      f.requireReplay(f.messages(), tip, peer)
      f.tracker.getStatus(peer) shouldBe Some(Equal)

      // no reconnect, no status change: only the time since this node last sent sync to the peer passes 60 s
      f.markOutdated(peer)
      f.sync.inboundV2(f.history, peer)
      f.tracker.getStatus(peer) shouldBe Some(Equal)
      val messages = f.messages()
      f.requireSync(messages)
      f.requireReplay(messages, tip, peer)
    }
  }
```

The per-connection variant (review-only, not a proposed patch):
```diff
diff --git a/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala b/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
index fbb64c74b..242e3f524 100644
--- a/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
+++ b/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
@@ -390,6 +390,8 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
     * matches ours at the local full-block height. `Equal` only compares header chains; the
     * matching height is a routing precondition and does not prove the peer has full blocks.
     */
+  private val replayedTo = mutable.Set[ConnectedPeer]() // review-only mutation: at most one replay per connection
+
   private def sendProcessedInputTip(history: ErgoHistory, peer: ConnectedPeer): Unit = {
     val localHeight = history.fullBlockHeight
     val eligible = localHeight > 0 &&
@@ -398,13 +400,14 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
       syncTracker.statuses.get(peer).exists { status =>
         status.status == Equal && status.height == localHeight
       }
-    if (eligible) {
+    if (eligible && !replayedTo.contains(peer)) {
       // The announced tip may still await transactions. Only replay the processed prefix.
       history.bestInputBlocksChain().headOption.flatMap(history.getInputBlock).foreach { tip =>
         val announcement = if (tip.weakTxIds.getOrElse(Seq.empty).size <= 3) tip
         else tip.copy(weakTxIds = None)
         val message = Message(InputBlockMessageSpec, Right(announcement), None)
         networkControllerRef ! SendToNetwork(message, SendToPeer(peer))
+        replayedTo += peer
       }
     }
   }
@@ -2099,6 +2102,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
 
     case DisconnectedPeer(connectedPeer) =>
       syncTracker.clearStatus(connectedPeer)
+      replayedTo -= connectedPeer
   }
 
   protected def sendLocalSyncInfo(historyReader: ErgoHistory): Receive = {
```
</details>

Issue #1307, opened 2021-03-03
The node can't read JSON file with encrypted secret if more than file in directory


kushti:
Following some tutorial, users set ergo,directory = "c:/ergo" , ergo.wallet.secretDir = "c:/ergo". Then the node can't read the encrypted seed file as reader fails if there's more than one file in the secretDir.

As a fix, the node may try to read the first JSON file found. Should cover most of the cases.

Issue #671, opened 2019-03-25
Improved secret storage


kushti:
1. If no wallet created yet, it is possible to create it by running the node with a certain flag. Password and, optionally, seed are to be provided during wallet generation (via command line), if no seed is provided it is generated from system randomness. Password is used to encrypt a wallet file (w. seed and wallet historical operations).

A wallet could be imported using either
* seed directly
* BIP39 mnemonic (if supported)

To password-protect the seed on the filesystem (in the wallet file), the encryption key should be derived from the password using a password hashing function (Argon2, PBKDF2, or scrypt), with sufficiently many iterations and a random salt. For example 50,000 PBKDF2 iterations.

If a wallet needs be generated then  system randomness should be fine (SecureRandom's). It may be better to generate a BIP39 mnemonic, because many interfaces and hardware wallets only support import via BIP39. And of course maybe the most important is to make sure there is a backup of the wallet/seed/mnemonic .

2.  If wallet is created, it is possible to open it, again, by providing a certain flag when starting a node.

PR #2507, opened 2026-09-06
Propagate wallet secret-file persistence failures

a-shannon:
Wallet keystore creation used `PrintWriter`, which can hide write and close failures. This change propagates those failures, closes a complete staging file before atomic publication, and erases the supplied seed on both success and failure.

The [scope and compatibility correction](https://github.com/a-shannon/ergo/compare/d9032c373c157f37fa02c46bea29a91f6e6fb732...15a1e04c5d7f6fbbac9d7bfdb801e320520d283f) is now one wallet-only commit on the target branch. The shared integration commits have been removed from its ancestry as well as its diff, so merging it cannot mark those changes as merged and then reverted. The complete PR now contains exactly three files: `JsonSecretStorage.scala`, its direct persistence spec, and `ErgoWalletServiceSpec.scala`. The five integration files have been removed from this PR. The service spec isolates each initialization case's directories and closes its own storage handles.

### Review corrections

- [Initialization](https://github.com/ergoplatform/ergo/pull/2507#discussion_r4000876172) now consults discovery. A discovered wallet blocks initialization; explicit absence permits it; ambiguous or unreadable inventories remain errors. Node data without a discoverable wallet no longer blocks initialization. Two JSON wallets cannot acquire an undiscoverable third wallet through this path.
- [Discovery](https://github.com/ergoplatform/ergo/pull/2507#discussion_r4000876242) preserves [#1307's shared-directory support](https://github.com/ergoplatform/ergo/commit/19d7a9d9d1ed4c5c970eeae28aa06e466bf1bcb6): exactly one JSON wallet can coexist with `history/`, `state/`, `peers/` and other files. A sole legacy filename remains supported. Multiple JSON candidates fail explicitly. Discovery does not parse or decrypt a selected file.

### Validation

Current head: `15a1e04c5d7f6fbbac9d7bfdb801e320520d283f`.

- The complete source tree is identical to the tested `78e011163` revision; only its parent history changed.
- The three layout/recovery regressions failed on the preceding production code and passed after this correction.
- Local JDK 8 / Scala 2.12.20: persistence 21/21, existing secret storage 4/4, wallet service 14/14. These tests ran without the removed integration helpers.
- Independent review found no blocker in the three-file scope. A direct composition with [#2543](https://github.com/ergoplatform/ergo/pull/2543)'s initialization preflight passed four layout cases, including rejection before opening stores. The POSIX symlink case was unavailable on the Windows test host.
- Current-head GitHub CI passed [8/8](https://github.com/ergoplatform/ergo/actions/runs/34784694835), with the test-execution step successful in each job. The preceding `d9032c373` head passed all eight jobs in [attempt 2](https://github.com/ergoplatform/ergo/actions/runs/34724201467/attempts/2). Attempt 1's Java-setup failure did not execute integration tests and is not test-pass evidence. Historical approval and CI do not apply to the new head.

### Limits and integration

Publication is atomic visibility, not power-loss durability; neither the file nor its parent directory is forced to stable storage. This change does not add cross-process initialization coordination. Cleanup removes only the failed operation's own staging file. POSIX staging permissions are restricted to `rw-------`; Linux CI supplies the platform-specific checks.

[#2543](https://github.com/ergoplatform/ergo/pull/2543) consumes the explicit absence result and should follow this correction. It has not been rebased as part of this review. [#2508](https://github.com/ergoplatform/ergo/pull/2508)'s cipher-parameter correction is independent. Integration order and exact review/CI pins are maintained in [#2533](https://github.com/ergoplatform/ergo/issues/2533).


jozanek (approved):

kushti (approved):

jozanek (commented):
Thanks for this — I read it as: `PrintWriter` swallows `IOException` into a flag nobody checks, so a write failing mid-flush leaves a truncated keystore while the node reports success. Worth fixing.

Two framing points that changed how I read the diff:

**Propagation already exists.** Both call sites wrap `init` in `Try` (`ErgoWalletService.scala:310`, `:339`), `ErgoWalletActor.scala:87-99` logs, `WalletApiRoute.scala:392-404` folds to a 400. The defect is narrower than the title: `PrintWriter` never throws, and with ~250 bytes against an 8192-char buffer the silent window is the flush in `close()`. Is an error lost anywhere else?

**Who has hit this?** Nothing in the tracker for corrupt or empty keystores, and #671 states no atomicity or durability requirement. `/wallet/init` also hands back the mnemonic, so a bad keystore surfaces at the next unlock while it's still on screen — which makes the cost side hard to weigh. Is there a report I missed?

Smaller: `ATOMIC_MOVE` turns a working `init` into a hard failure where rename is unavailable — does that bite any supported deployment? #2508's branch doesn't contain `1452249f3` (ahead 14 / behind 1), so I couldn't find the "tested together" composite — is it pushed? And nothing logs; `JsonSecretStorage` doesn't extend `ScorexLogging`.

What would close it out: the `Files.exists` guard either implementing the contract `openapi.yaml:4255` documents or going away with the test name that implies it (inline), plus a line in the description on the permission change and the atomicity/durability trade. The `readFile` ambiguity is fine as its own PR — I'd just like it named.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
```diff
@@ -89,29 +92,60 @@ final class JsonSecretStorage(val secretFile: File, encryptionSettings: Encrypti
...
+      val file = new File(s"${settings.secretDir}/$uuid.json")
+      persist(file, encryptedSecret.asJson.noSpaces)
+      new JsonSecretStorage(file, settings.encryption)
+    } finally {
+      util.Arrays.fill(seed, 0: Byte)
+    }
+  }
+
+  /** Writes and closes a staging file before publishing it atomically. */
+  private[secrets] def persist(file: File, jsonRaw: String,
```
jozanek, line 152: `openWriter` has no non-test caller, and `private[secrets]` compiles to public bytecode plus a synthetic `persist$default$3`, so it ships in the 2.11/2.12/2.13 artifacts. It does earn its keep — close-time ENOSPC isn't reachable by filesystem tricks.

Would two overloads work instead of a default argument, so the injectable form is opt-in?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
```diff
@@ -89,29 +92,60 @@ final class JsonSecretStorage(val secretFile: File, encryptionSettings: Encrypti
...
+                               openWriter: Path => Writer = path => Files.newBufferedWriter(path, UTF_8)): Unit = {
+    val target = file.toPath.toAbsolutePath
+    Files.createDirectories(target.getParent)
+    if (Files.exists(target, LinkOption.NOFOLLOW_LINKS)) {
```
jozanek, line 156: `target` is `nameUUIDFromBytes(ciphertext)` with a fresh IV and salt per call, so this can never fire — a second init still succeeds and leaves two `.json` files. `openapi.yaml:4255` already promises "Fails if a wallet file already exists", and the test "persistence does not overwrite an existing destination" reads as if that promise were now kept.

Would you implement the documented contract — fail when `readFile(settings).isSuccess` — or drop the check and rename the test?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
```diff
@@ -89,29 +92,60 @@ final class JsonSecretStorage(val secretFile: File, encryptionSettings: Encrypti
...
+      throw new FileAlreadyExistsException(target.toString)
+    }
+    val staging = Files.createTempFile(target.getParent, StagingPrefix, ".tmp")
```
jozanek: `Files.createTempFile` yields owner-only permissions on POSIX; `new PrintWriter(file)` used umask default. Tighter is what I'd want for a spending key, but it's a `TempFileHelper` side effect, spec'd only as "may be more restrictive", and unmentioned.

Would you set it explicitly via `PosixFilePermissions.asFileAttribute` and assert it?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
```diff
@@ -89,29 +92,60 @@ final class JsonSecretStorage(val secretFile: File, encryptionSettings: Encrypti
...
+          writeFailure = Some(error)
+          throw error
+      } finally {
+        writeFailure match {
+          case Some(error) => Try(writer.close()).failed.foreach(error.addSuppressed)
+          case None => writer.close()
+        }
+      }
+      // Do not expose a partial destination when atomic publication is unavailable.
+      Files.move(staging, target, StandardCopyOption.ATOMIC_MOVE)
```
jozanek, line 181: No `force` on the staging fd or parent dir, so this buys atomicity, not durability. I think that's the right half — `/wallet/init` returns the mnemonic (`WalletApiRoute.scala:398`), so a lost keystore is recoverable by restore while a partial one that `readFile` returns is not.

Would you put that in the description? It reads as a limitation today rather than the threat model.

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
```diff
@@ -89,29 +92,60 @@ final class JsonSecretStorage(val secretFile: File, encryptionSettings: Encrypti
...
+    } catch {
+      case NonFatal(error) =>
+        Try(Files.deleteIfExists(staging)).failed.foreach(error.addSuppressed)
```
jozanek, line 184: Cleanup runs only on a caught exception, so a `kill -9` between `createTempFile` and the move orphans a `.ergo-secret-staging-*.tmp` permanently — and the new `readFile` filter hides it from the node too. Encrypted and owner-only, so not exposure; just nothing reclaims it.

Worth a prefix sweep at the top of `persist`?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
```diff
@@ -133,7 +167,7 @@ object JsonSecretStorage {
   def readFile(settings: SecretStorageSettings): Try[JsonSecretStorage] = {
     val dir = new File(settings.secretDir)
     if (dir.exists()) {
-      dir.listFiles().toList match {
+      dir.listFiles().toList.filterNot(_.getName.startsWith(StagingPrefix)) match {
```
jozanek: Good to see this filter — `19d7a9d9d`/#1307 shows stray files in `secretDir` really do break wallet loading. What stops me is the branch below: the `.json` filter only runs when `size > 1`, and `init` never removes a prior keystore, so a re-initialized node has two `.json` files and `headOption` picks by `listFiles()` order. A restore silently shadowed by the old keystore looks like the larger loss path.

In scope here, or its own PR?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
jozanek: Thanks — the guard landed.

One correction of mine: `openapi.yaml:4255` calling restore the overwrite path has been wrong since `67d04827d` (`ErgoWalletActor.scala:102` blocks restore once a secret is set), so I pointed you at a contract that doesn't exist.

Still open though: `storageEntries(...).nonEmpty` is broader than "a wallet exists". When `readFile` fails but the directory has files, init used to recover the node; now it throws and the only repair is cleaning `secretDir` by hand. Would gating on `readFile(settings)` work instead?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
jozanek: (Collapsed as outdated — this is now `readFile` at `:203-208`.)

Thanks for taking it here. One thing I missed when I raised it: `case _ => throw` restores what `19d7a9d9d` removed — kushti's fix for #1307, users with `ergo.directory` == `secretDir`. There the node puts `history/`, `state/`, `peers/` inside `secretDir`, so discovery fails again, and `:130` also blocks init with no wallet present.

Master had no test on that path, so it would go out silently. Could the description name the revert, or the `.json` fallback stay for the single-json case?

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
a-shannon: Agreed. In [#2507](https://github.com/ergoplatform/ergo/pull/2507) at `15a1e04c5`, initialization now consults `readFile(settings)`: a discovered wallet blocks initialization, explicit `SecretFileNotFoundException` permits it, and ambiguous or unreadable inventories remain errors. Node data directories without a discoverable wallet no longer prevent initialization.

This does not treat every discovery failure as permission to create another wallet, and discovery does not parse or decrypt a selected file. The three layout regressions failed before the correction and passed afterward. Persistence, existing storage and service tests passed 21/21, 4/4 and 14/14; [current-head CI passed all eight jobs](https://github.com/ergoplatform/ergo/actions/runs/34784694835).

ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/JsonSecretStorage.scala
a-shannon: You were right: that branch lost the shared-directory behavior from #1307. [#2507](https://github.com/ergoplatform/ergo/pull/2507) at `15a1e04c5` restores discovery of exactly one JSON wallet alongside `history/`, `state/`, `peers/` and other files, and preserves a sole legacy filename. Multiple JSON candidates fail explicitly. The direct regression coverage includes discovery and initialization in these layouts.

The PR now contains only `JsonSecretStorage.scala`, its persistence spec and the service spec. The unrelated integration commits were removed from both its diff and ancestry. [All eight current-head CI jobs passed](https://github.com/ergoplatform/ergo/actions/runs/34784694835).

jozanek (approved):
LGTM

PR #2508, opened 2026-09-06
Unlock wallets with their persisted cipher parameters

a-shannon:
Wallet unlock derives the key using the cipher parameters stored in the wallet file. Changing the node's current encryption configuration therefore no longer prevents an existing wallet from reopening with its original password. New wallet files use the current configuration; opening an existing file does not rewrite it.

Focused coverage includes legacy wallet compatibility, both derivation modes, independent PRF, iteration-count and key-length changes, and invalid-parameter handling. Unsupported or nonpositive parameters return the existing failure result through the cryptographic provider. The correction introduces no arbitrary iteration limit. Independent source review is complete.

[#2507](https://github.com/ergoplatform/ergo/pull/2507) adds checked persistence and atomic publication in the same file. The two corrections apply independently and have also been tested together. The unlock correction is isolated in commit `a785374f386ac2b1c02f41ae2cef61f9eeab0a91`.

The [published composition](https://github.com/a-shannon/ergo/commit/438df263256f1ff7a694d4bd6c0f4c61b902e83d) adds only that unlock correction to #2507's reviewed storage head `f17d145d62f4c0375401f7aea64915625d0fe1a8`. Review the [two-file increment](https://github.com/a-shannon/ergo/compare/f17d145d62f4c0375401f7aea64915625d0fe1a8...438df263256f1ff7a694d4bd6c0f4c61b902e83d): all 35 focused wallet tests passed on the combination. Its production patch matches the owning unlock commit, so the synchronization changes carried by this PR are not part of that composition.

This branch also carries the shared synchronization changes from [#2511](https://github.com/ergoplatform/ergo/pull/2511), which should be reviewed separately. Their scheduling correction extends backward to retained ancestors while preserving the forward ceiling anchored to the best full block, existing pruning and filtering, request-count logic and terminal-height handling. The port has independent composition review. This dependency is separate from wallet key derivation.

Included CI support retains Ergologica's contributions from [#2452](https://github.com/ergoplatform/ergo/pull/2452) and [#2455](https://github.com/ergoplatform/ergo/pull/2455), with authorship and provenance.

## Current CI

Commit [`8af2db9156003417b72d9a96e88330ea9ece667f`](https://github.com/ergoplatform/ergo/commit/8af2db9156003417b72d9a96e88330ea9ece667f): [workflow run 34144500555](https://github.com/ergoplatform/ergo/actions/runs/34144500555). All eight checks passed, including node and integration tests and wallet/core tests across the three Scala versions.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
@@ -71,7 +71,7 @@ trait ToDownloadProcessor
             val toDownload = headersAtThisHeight.flatMap(requiredModifiersForHeader).filter { case (mtid, mid) => condition(mtid, mid) }
             // add new modifiers to download to accumulator
             val newAcc = toDownload.foldLeft(acc) { case (newAcc, (mType, mId)) => newAcc.adjust(mType)(_.fold(Vector(mId))(_ :+ mId)) }
-            continuation(height + 1, newAcc, maxHeight)
+            if (height == maxHeight) newAcc else continuation(height + 1, newAcc, maxHeight)
           } else {
             acc
           }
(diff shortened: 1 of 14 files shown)


src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -84,8 +84,20 @@ trait ToDownloadProcessor
...
+        } else {
+          val parentSteps = Math.max(0L, nodeSettings.keepVersions.toLong)
+          val limit = Math.min(fb.height.toLong, parentSteps + 1L).toInt
+          headerChainBack(limit, fb.header, h => isInBestChain(h.id)).headOption
+            .filter(h => isInBestChain(h.id))
+        }
+        val fromHeight = commonAncestor
+          .map(h => Math.max(minimalFullBlockHeight, h.height + 1))
+          .getOrElse(fb.height + 1)
+        val maxHeight = Math.min(fromHeight.toLong + FullBlocksToDownloadAhead - 1L, Int.MaxValue.toLong).toInt
```
jozanek: **[P1] Keep the download window beyond the existing full tip**

Anchoring the upper bound to the common ancestor can permanently stall a retained deep fork. For example, with a common ancestor at height 20, the old full tip at 219, best headers through 500, and the default `keepVersions = 200`, this scans only heights 21–212. If the replacement bodies through 219 are already stored, the download filter excludes that entire window and every call returns empty. At equal difficulty, `FullBlockProcessor.isBetterChain` requires strictly greater cumulative score, so the replacement through 219 does not move `bestFullBlock`; block 220 is needed but is never requested. This is reachable after a restart with the headers already stored, when fresh header-arrival requests cannot rescue it. The previous window, 220–411, could progress in this state.

Please preserve the previous upper bound while extending the starting point backward, or advance the window past completed replacement blocks. I reproduced the empty result with a focused regression in `BlockDownloadSchedulingSpecification`; the 19 existing focused synchronization tests passed.

PR #2512, opened 2026-09-06
Fix Digest state recovery checkpoint root

a-shannon:
Digest-state recovery derives the stored version and state root from the same epoch checkpoint used to recover the context. This keeps checkpoint metadata consistent before subsequent headers are replayed. The production change is limited to the recovery caller; the adjacent recovery work in [#2424](https://github.com/ergoplatform/ergo/pull/2424) is separate.

The tests cover checkpoint recovery and reopening, actual node-view startup with later headers in history, persisted checkpoint metadata after replay, and reopening and rolling back to that checkpoint. The approved test-only synchronization support remains included, with selected-header agreement and bounded observations under the existing deadlines.

This PR now targets `v6.0.6`. Commit [`d4a509121e4f2b0abba3db96afe9ebabf43ede72`](https://github.com/ergoplatform/ergo/commit/d4a509121e4f2b0abba3db96afe9ebabf43ede72) replays the three approved commits onto release base `db0f1a4299e6c2561152dc6ca07990aef5403ae2`. All seven changed files have exactly the same contents as approved commit [`2f005d3bc00e78f6ac83e9aa3d41cd1c7629acb3`](https://github.com/ergoplatform/ergo/commit/2f005d3bc00e78f6ac83e9aa3d41cd1c7629acb3), and the feature diff has the same stable patch ID. Release build configuration and dependency versions are retained. Independent review confirmed the exact port.

Validation on the release branch: JDK 8 / Scala 2.12.20; all nine Digest recovery/state tests and all nine observation contracts passed, and integration sources compiled. [Current release CI](https://github.com/ergoplatform/ergo/actions/runs/34198630972) passed all eight checks, including node and integration tests and the three Scala versions for wallet/core. GitHub reports no merge conflicts with `v6.0.6`.

### DIFF
--- a/src/test/scala/org/ergoplatform/nodeView/state/DigestStateSpecification.scala
+++ b/src/test/scala/org/ergoplatform/nodeView/state/DigestStateSpecification.scala
@@ -20,6 +20,40 @@ class DigestStateSpecification extends ErgoCorePropertyTest {
   private val emptyVersion: VersionTag = bytesToVersion(Array.fill(32)(0: Byte))
   private val emptyAdDigest: ADDigest = ADDigest @@ Array.fill(32)(0: Byte)
 
+  property("recover a valid checkpoint and reopen it") {
+    val (us, bh) = createUtxoState(settings)
+    try {
+      val block = validFullBlock(parentOpt = None, us, bh)
+      val checkpoint = block.header
+      val checkpointContext = us.stateContext.appendFullBlock(block).get
+      val dir = createTempDir
+
+      def checkCheckpoint(state: DigestState): Unit = {
+        state.version shouldEqual checkpoint.id
+        state.rootDigest shouldEqual checkpoint.stateRoot
+        state.stateContext.lastHeaderOpt.map(_.id) shouldEqual Some(checkpoint.id)
+        state.stateContext.bytes shouldEqual checkpointContext.bytes
+      }
+
+      val recovered = DigestState.recover(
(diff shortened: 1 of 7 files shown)


src/test/scala/org/ergoplatform/nodeView/state/DigestStateSpecification.scala
```diff
@@ -20,6 +20,40 @@ class DigestStateSpecification extends ErgoCorePropertyTest {
...
+      val dir = createTempDir
+
+      def checkCheckpoint(state: DigestState): Unit = {
+        state.version shouldEqual checkpoint.id
+        state.rootDigest shouldEqual checkpoint.stateRoot
+        state.stateContext.lastHeaderOpt.map(_.id) shouldEqual Some(checkpoint.id)
+        state.stateContext.bytes shouldEqual checkpointContext.bytes
+      }
+
+      val recovered = DigestState.recover(
```
jozanek, line 38: The checkpoint-root change looks correct, but the new test doesn’t exercise the caller being fixed. It supplies consistent values directly to `DigestState.recover`; I confirmed it also passes against the unchanged base.

Please add a regression test that triggers startup recovery with different epoch-checkpoint and tip roots, then inspects the persisted checkpoint or rolls back to it and checks its root. Asserting only the final tip won’t catch this bug because header replay replaces the active root. The test should fail with the original expression and pass with this fix.

jozanek (approved):

kushti (approved):

PR #2513, opened 2026-09-06
Propagate state rollback failures and verify recovered roots

a-shannon:
Uses shared test support from [#2535](https://github.com/ergoplatform/ergo/pull/2535). Review the [six-file state rollback increment](https://github.com/a-shannon/ergo/compare/b3ebc6d6f5a36514d28631d0852d1ba91c9f1781...bb43ed4dd8b3421da7b16557ba857c4bb929c90a). For reviewers of the earlier `c7a2fd34899d86dc77332b8a0e0f31e5193324bb` head, the [new Digest-only increment](https://github.com/a-shannon/ergo/compare/f1cfeea3aeeca03ae8567ace041b28da8a47f619...bb43ed4dd8b3421da7b16557ba857c4bb929c90a) contains the additional correction and direct tests; the earlier four owned files are unchanged. The upstream diff is cumulative until shared support merges.

AVL and Digest rollback propagate returned or thrown store failures before continuing restoration. UTXO rollback checks the complete recovered digest before returning the requested state. The Digest correction stops unwanted cleanup/root reads and preserves the original error; its successful path is unchanged. Existing caller recovery policy is retained.

Current validation: eight focused state tests pass on the composed tree, including the three new Digest cases, the existing UTXO rollback suite and ordinary Digest rollback. Independent source and test review completed. The earlier 38-test result belongs to the previous `c7a2fd3` head. All eight checks passed at the current `bb43ed4dd` head ([CI run](https://github.com/ergoplatform/ergo/actions/runs/34292536724)).

The already-current rollback work in [#2277](https://github.com/ergoplatform/ergo/pull/2277) addresses a separate condition.

### DIFF
--- a/avldb/src/main/scala/scorex/crypto/authds/avltree/batch/VersionedLDBAVLStorage.scala
+++ b/avldb/src/main/scala/scorex/crypto/authds/avltree/batch/VersionedLDBAVLStorage.scala
@@ -12,7 +12,7 @@ import scorex.db.{LDBKVStore, LDBVersionedStore}
 import scorex.util.ScorexLogging
 
 import scala.collection.mutable
-import scala.util.{Failure, Try}
+import scala.util.{Failure, Success, Try}
 
 /**
   * Persistent versioned authenticated AVL+ tree implementation on top of versioned LevelDB storage
@@ -44,8 +44,8 @@ class VersionedLDBAVLStorage(store: LDBVersionedStore)
     if (!this.version.contains(version)) { // do not rollback to self
       log.info(s"Doing rollback from ${this.version.map(Base16.encode)} to ${Base16.encode(version)}:")
       store.rollbackTo(version)
-    }
-  }.flatMap(_ => restorePrunedRootNode())
+    } else Success(())
+  }.flatten.flatMap(_ => restorePrunedRootNode())
     .recoverWith { case e =>
       log.warn(s"Failed to recover tree for digest ${Base16.encode(version)}:", e)
       Failure(e)
--- a/avldb/src/test/scala/scorex/crypto/authds/avltree/batch/VersionedLDBAVLStorageSpecification.scala
+++ b/avldb/src/test/scala/scorex/crypto/authds/avltree/batch/VersionedLDBAVLStorageSpecification.scala
@@ -19,7 +19,7 @@ import scorex.utils.{Random => RandomBytes}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
-import scala.util.{Success, Try}
+import scala.util.{Failure, Success, Try}
 
 class VersionedLDBAVLStorageSpecification
   extends AnyPropSpec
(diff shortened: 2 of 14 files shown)


cafebedouin:
Review carried out by Claude (Anthropic, Claude Opus 5.5) for cafebedouin, using forkbench; executed the three state specs with production hunks reverted (JDK 8), and two-node and three-node network runs of v6.0.5 (the merge base) against this branch (JDK 21).

**Each hunk has a new test that fails without it and passes with it; successful UTXO rollbacks still complete on a network. One maintainability note on the UTXO path.**

Executed: specs at `bb43ed4`, production files reverted to `5528ef5`. Reverting only the AVL hunk fails 2 of the 5 new AVL tests (unavailable version; returned store failure). Reverting all three hunks fails the Digest returned-failure test: the `DigestState` constructor throws `requirement failed: version should always be equal to store.lastVersionID` after `store.clean` has run. Reverting only the UTXO hunk fails both wrong-root tests. All pass with the change. Seven other new tests pass both ways and guard existing behaviour.

On a network, a candidate follower switched forks after one `Rollback UtxoState` with every node responsive (released 6.0.6 on this host: no switch in 8 of 16 runs), and candidate nodes rolled back two one-block reorgs and applied the next block. Payments: 10 of 10 confirmed on both builds (6.0.6: 10 of 10). Follower-agreement runs are inconclusive here: they also failed with v6.0.5 alone, with no failed rollback in any retained log. With the change, a rollback failure now reaches the existing stop branch (`ErgoNodeViewHolder.scala:196`, `:215-219`; read). Not run: the project CI.

**[Maintainability] Restore the prover when the recovered root is wrong, or document that the state must be discarded**

Observed: `UtxoState.rollbackTo` calls `p.rollback(rootHash)` and then `require(p.digest.sameElements(rootHash), …)` (`UtxoState.scala:56-57`). scrypto 3.1.1's `PersistentBatchAVLProver.rollback` installs the recovered tree before returning, so when the `require` fails, the shared prover already holds the wrong root. The PR's own test asserts this (`UtxoStateRollbackSpecification.scala:43`).

The AVL and Digest paths now return the store's failure before touching the prover or the root read; the UTXO path returns its failure after replacing the prover. Nothing breaks today: both callers stop on failure (`ErgoNodeViewHolder.scala:219` is `???`; `:606` calls `.get`). This reading is wrong if a caller keeps the receiver after a `Failure`.

Recommended: a doc line on `rollbackTo` saying the receiver is unusable after a `Failure` is enough given both callers; restoring the pre-call digest is the stricter option. (read; the asserting test was executed)

```
revert (class | with | without):
all three files   DigestStateRollbackFailureSpec  | pass | fail
all three files   UtxoStateRollbackSpecification  | pass | fail
UtxoState only    UtxoStateRollbackSpecification  | pass | fail   (wrong hash, wrong height)
AVL only          avldb/testOnly VersionedLDBAVLStorageSpecification: with 13/0, without 11/2
pass both ways: Digest thrown + success path; UTXO prover failure + stored-root publish; AVL thrown, self-rollback, reopen

network (base jar ergo-6.0.5 2a7e2978cb09…, candidate 307746d8d750…):
fork switch, 1 pair: base switched=false; candidate switched=true, 1 Rollback UtxoState; all nodes responsive
payments:            base 10/10 confirmed; candidate 10/10 confirmed
digest/pruned followers settled: base 2/4, candidate 0/4 (all failures: follower behind at the deadline)
release/candidate agreement: 1/8 runs with the candidate on either side; 6.0.5 alone: 1/2
"Unexpected root digest after rollback" / "Rollback failed" in retained candidate node logs: 0
```

</details>

PR #2520, opened 2026-09-06
Budget complete serialized transaction sections during candidate assembly

a-shannon:
Candidate assembly budgets the complete serialized transaction section: header ID, version marker, transaction count and payloads. Measurement uses the section writer's version context, including fee replacement. Exact size fits remain eligible; the existing strict cost limit is unchanged.

The budget is captured once for selection and emission-only fallback. Before returning collected transactions, the actual section writer checks the result. An oversized result or writer failure restores an earlier accepted prefix with its matching fee and invalid-transaction snapshot. If no nonempty candidate fits, construction fails without issuing work. The emission-only proof fallback has its own size check.

### September 13 review follow-up

Review the [three-file increment](https://github.com/a-shannon/ergo/compare/d904fd1da9697ce6d5471e4416570c7bd0cf8ced...c9ddcaaa5207d5373a720a4399ea27489faa6a9e). Current head: `c9ddcaaa5207d5373a720a4399ea27489faa6a9e`.

- `BlockTransactions.sizeOf` now documents its existing nonempty-transaction precondition.
- Expected oversize recovery logs at DEBUG; reconstruction or measurement failures log at WARN.
- The requested memory measurement found that shared Vector nodes still retained every historical fee transaction. Each fee transaction grows with the accepted prefix. Snapshots now retain shared costed prefixes and reconstruct historical fees only when fallback needs them. Ordinary selection, limits, fee construction and no-fee accumulator transitions are unchanged. A failed reconstruction continues to an earlier snapshot.

The measurement invoked the compiled recipe factory and verified that each closure retains the original prefix by identity, without eagerly copying it. With 256 synthetic fee-producing transactions, retained input objects fell from 33,152 to 512, including ordinary inputs. For a 524,166-byte synthetic section with 2,864 ordinary transactions, the new strategy retained 2,865 transaction objects, 5,728 inputs and about 4.13 MB of traversed structure and payload. This measures retained objects for that fixture, not peak heap, transient allocation, execution complexity or a state/cost-valid full block. Fees already embedded in accepted candidate content are preserved.

### Validation

- Current JDK 8 / Scala 2.12.20 execution: `CandidateGeneratorPropSpec` 19/19, with no failures or skips. Cases include exact fit, one-byte-under budgets, frozen limits and actual incremental/final writer failure with and without fees.
- Recovery checks preserve exact section bytes and the corresponding invalid-ID snapshot, skip failed reconstruction, terminate on exhaustion, and perform no historical reconstruction when the current candidate succeeds.
- The new reconstruction regression fails against the preceding production code: the exception escapes the old `checkedCandidate`. It passes with this increment.
- Independent review of all three changed files found no blocking issue. The retained-object measurement and focused tests bind the source above.
- Current-head GitHub CI passed [8/8](https://github.com/ergoplatform/ergo/actions/runs/34784284996), with the test-execution step successful in each job. The preceding approved `d904fd1da` head passed [all eight jobs](https://github.com/ergoplatform/ergo/actions/runs/34721554177). Its approval and CI remain historical evidence.

Earlier serializer evidence, unchanged by this follow-up, covers writer-byte comparisons and parser round trips for versions 1 through 4 and counts 1, 127 and 128 (6/6). The earlier actor suite initially passed 19/20; its forced-cache case passed separately on both baseline and corrected code. That original failed result is retained. Standalone sizing does not imply acceptance of SOption registers: the pinned parser still rejects V6-only register types.

### Integration

The requested full-section assertions are now published on [#2479](https://github.com/ergoplatform/ergo/pull/2479) itself at [`f75bda15c`](https://github.com/a-shannon/ergo/commit/f75bda15cb1b859a505a264c5b5a5b167d2c9809), with 16/16 tests. They use the existing writer and preserve its `v6.0.7` target.

For a common release line, integrate this sizing correction before adapting #2479's deferral changes. The [earlier combined revision](https://github.com/a-shannon/ergo/commit/5f171a53a65e6b22615780647aeceb2386f87858) passed 48/48 at its pinned inputs; it predates this retention increment and is not validation of the new combination. If [#2462](https://github.com/ergoplatform/ergo/pull/2462) is combined later, accounting must include every fee chunk. Current review increments and integration order are tracked in [#2533](https://github.com/ergoplatform/ergo/issues/2533).

### DIFF
--- a/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
@@ -15,7 +15,8 @@ import org.ergoplatform.serialization.ErgoSerializer
 import scorex.crypto.authds.LeafData
 import scorex.crypto.authds.merkle.{Leaf, MerkleProof, MerkleTree}
 import scorex.crypto.hash.Digest32
-import scorex.util.serialization.{Reader, Writer}
+import scorex.util.ByteArrayBuilder
+import scorex.util.serialization.{Reader, VLQByteBufferWriter, Writer}
 import scorex.util.{ModifierId, bytesToId, idToBytes}
 import scorex.util.Extensions._
 import sigma.VersionContext
@@ -141,23 +142,42 @@ object BlockTransactionsSerializer extends ErgoSerializer[BlockTransactions] {
 
   override def serialize(bt: BlockTransactions, w: Writer): Unit = {
     w.putBytes(idToBytes(bt.headerId))
-    val blockVersion = bt.blockVersion
+    serializeMetadata(bt.blockVersion, bt.txs.size, w)
+    bt.txs.foreach(tx => serializeTransaction(tx, bt.blockVersion, w))
+  }
+
+  private def serializeMetadata(blockVersion: Version, transactionCount: Int, w: Writer): Unit = {
     if (blockVersion > 1) {
       // see comments in parse()
-      w.putUInt(MaxTransactionsInBlock.toLong + bt.blockVersion)
+      w.putUInt(MaxTransactionsInBlock.toLong + blockVersion)
     }
-    w.putUInt(bt.txs.size.toLong)
-    bt.txs.foreach { tx =>
-      if (blockVersion >= VersionContext.V6SoftForkVersion) {
-        // since 6.0 we use versioned serializers
-        VersionContext.withVersions(blockVersion, blockVersion) {
-          ErgoTransactionSerializer.serialize(tx, w)
-        }
-      } else {
-        // before 6.0 activation, VersionContext is not used
+    w.putUInt(transactionCount.toLong)
+  }
+
+  private def serializeTransaction(tx: ErgoTransaction, blockVersion: Version, w: Writer): Unit = {
+    if (blockVersion >= VersionContext.V6SoftForkVersion) {
+      // since 6.0 we use versioned serializers
+      VersionContext.withVersions(blockVersion, blockVersion) {
         ErgoTransactionSerializer.serialize(tx, w)
       }
+    } else {
+      // before 6.0 activation, VersionContext is not used
+      ErgoTransactionSerializer.serialize(tx, w)
     }
+  }
+
+  /** Size of a transaction in the section's serialization context, independent of its cached standalone size. */
+  def transactionSize(tx: ErgoTransaction, blockVersion: Version): Int = {
+    val w = new VLQByteBufferWriter(new ByteArrayBuilder())
+    serializeTransaction(tx, blockVersion, w)
+    w.result().toBytes.length
+  }
+
+  /** Full section size from the already measured transaction payloads. */
+  def sectionSize(blockVersion: Version, transactionCount: Int, payloadSize: Long): Long = {
+    val w = new VLQByteBufferWriter(new ByteArrayBuilder())
+    serializeMetadata(blockVersion, transactionCount, w)
+    Constants.ModifierIdSize.toLong + w.result().toBytes.length + payloadSize
   }
 
   override def parse(r: Reader): BlockTransactions = {
--- a/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
+++ b/ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
@@ -101,6 +101,10 @@ object BlockTransactions extends ApiCodecs {
 
   val modifierTypeId: NetworkObjectTypeId.Value = BlockTransactionsTypeId.value
 
+  /** Complete wire size, measured by the section writer in its existing block-version context. */
+  def sizeOf(txs: Seq[ErgoTransaction], blockVersion: Version): Int =
+    BlockTransactions(Header.GenesisParentId, blockVersion, txs).bytes.length
+
   // Used in the miner when a BlockTransaction instance is not generated yet (because a header is not known)
   def transactionsRoot(txs: Seq[ErgoTransaction], blockVersion: Version): Digest32 = {
     if (blockVersion == Header.InitialVersion) {
@@ -166,15 +170,14 @@ object BlockTransactionsSerializer extends ErgoSerializer[BlockTransactions] {
     }
   }
 
-  /** Size of a transaction in the section's serialization context, independent of its cached standalone size. */
-  def transactionSize(tx: ErgoTransaction, blockVersion: Version): Int = {
+  // Internal incremental accounting uses the same writer and version rules as sizeOf.
+  private[ergoplatform] def transactionSize(tx: ErgoTransaction, blockVersion: Version): Int = {
     val w = new VLQByteBufferWriter(new ByteArrayBuilder())
     serializeTransaction(tx, blockVersion, w)
     w.result().toBytes.length
   }
 
-  /** Full section size from the already measured transaction payloads. */
-  def sectionSize(blockVersion: Version, transactionCount: Int, payloadSize: Long): Long = {
+  private[ergoplatform] def sectionSize(blockVersion: Version, transactionCount: Int, payloadSize: Long): Long = {
     val w = new VLQByteBufferWriter(new ByteArrayBuilder())
     serializeMetadata(blockVersion, transactionCount, w)
     Constants.ModifierIdSize.toLong + w.result().toBytes.length + payloadSize
--- a/ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
+++ b/ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
@@ -1,6 +1,15 @@
 package org.ergoplatform.modifiers.history
 
 import org.ergoplatform.utils.ErgoCorePropertyTest
+import org.ergoplatform.{ErgoBox, ErgoBoxCandidate, Input}
+import org.ergoplatform.modifiers.history.header.Header
+import org.ergoplatform.modifiers.mempool.{ErgoTransaction, ErgoTransactionSerializer}
+import scorex.util.{ByteArrayBuilder, idToBytes}
+import scorex.util.serialization.VLQByteBufferWriter
+import sigma.VersionContext
+import sigma.ast.{Constant, ErgoTree, SInt, SOption}
+import sigma.data.TrivialProp.TrueProp
+import sigma.interpreter.ProverResult
 
 class BlockTransactionsSpec extends ErgoCorePropertyTest {
   import org.ergoplatform.utils.generators.CoreObjectGenerators._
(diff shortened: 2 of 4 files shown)


jozanek (approved):

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -917,15 +918,20 @@ object CandidateGenerator extends ScorexLogging {
...
   }
 
-  /**
-    * Helper function which decides whether transactions can fit into a block with given cost and size limits
-    */
-  def correctLimits(
+  /** Check the complete serialized section before returning work, including the emission-only fallback. */
+  private[mining] def ensureCandidateSize(candidate: CandidateBlock, maxBlockSize: Int): Unit = {
+    val section = BlockTransactions(Header.GenesisParentId, candidate.version, candidate.transactions)
+    require(section.bytes.length <= maxBlockSize, "Candidate transaction section exceeds the block size limit")
```
jozanek: This `require` becomes `log.error` + `StatusReply.error` at line 214-220 with no `context.become`, so `cachedCandidate` stays `None` and the next poll re-runs the same deterministic `collectTxs` and fails identically. The loop above already checked this exact set, with this function, at this version.

`AGENTS.md` asks for `Try`/`Either` over exceptions here. Would you be open to logging and taking the existing `current -> invalidTxs` exit, so a miscount costs one transaction rather than all work?

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -705,6 +705,7 @@ object CandidateGenerator extends ScorexLogging {
       }
 
       def deriveWorkMessage(block: CandidateBlock) = {
+        ensureCandidateSize(block, state.stateContext.currentParameters.maxBlockSize)
```
jozanek: `ErgoStateReader.stateContext` is a `def` that re-reads the store on each access, so this is a second read of `maxBlockSize` after line 693 — the two can disagree across a parameter epoch flip.

Would you be open to reading the budget once above and passing it in?

jozanek (commented):
Thanks for this — the defect is real and I think the instrument is right, which is worth saying because the obvious alternative fails: a `safeGap`-style headroom constant would close the framing gap and nothing else.

But I think the body undersells the change. It leads with ~38 framing bytes, which is real but narrow. The version-context half gets one clause, and that looks like the actual justification: `collectTxs` has no `withVersions` wrapper, so `tx.size` in the old `correctLimits` is forced under the default `VersionContext(1, 1)`, where sigma 6.0.6 gates `SOption`/`SUnsignedBigInt` on `isV3OrLaterErgoTreeVersion` and the fallthrough raises. If that's right, a 6.0 transaction with such a register doesn't undercount on master — it raises, and the candidate is lost. Have I got that right? I couldn't confirm it by running anything. If so, I'd lead with it and add a test at the `tx.size` level.

Three things I'm holding:

1. **#2479 contradicts this PR.** It sets `maxSize = smallTx.size + smallFeeTx.size + 1` and then expects `smallTx` to be selected. Under this PR `payload + 1` can't fit `payload + 33..39`, so `collectTxs` returns empty. #2462 also removes the `lastFeeTx: Option[...]` binding that the new `blockPayloadSize` line reads. Could you state the intended merge order and land the later one rebased and green, rather than leaving it to whoever merges second?

2. **The guard withholds work rather than degrading** — inline at line 924.

3. **Nothing parses a section at block version ≥ 2** — inline at the spec.

Minor, no code implied: `transactionSize` uses `withVersions(v, v)` while `parse` in the same file uses `v − 1`, as do `ErgoMemPool` and `ErgoNodeViewSynchronizer`. Worth a sentence on which is intended, since this makes it public API.

What would move me to approve: merge order settled with #2479, the guard degrading, and round-trip coverage above version 1. I didn't run the build; CI is green on 9c877b4.

ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
```diff
@@ -141,25 +142,44 @@ object BlockTransactionsSerializer extends ErgoSerializer[BlockTransactions] {
...
 
+  /** Size of a transaction in the section's serialization context, independent of its cached standalone size. */
+  def transactionSize(tx: ErgoTransaction, blockVersion: Version): Int = {
+    val w = new VLQByteBufferWriter(new ByteArrayBuilder())
+    serializeTransaction(tx, blockVersion, w)
+    w.result().toBytes.length
+  }
+
+  /** Full section size from the already measured transaction payloads. */
+  def sectionSize(blockVersion: Version, transactionCount: Int, payloadSize: Long): Long = {
```
jozanek: `sectionSize` takes a `payloadSize` it can't validate, so it's confidently wrong if the caller accumulated under a different version — and `ensureCandidateSize` partly exists to check the caller got it right.

Line 102 has the pattern for this same caller: `transactionsRoot(txs, blockVersion)`. Would you consider a single `sizeOf(txs, blockVersion)` beside it? `ergo-core` cross-publishes with no MiMa, so a two-method surface is effectively permanent.

ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
```diff
@@ -16,4 +25,61 @@ class BlockTransactionsSpec extends ErgoCorePropertyTest {
...
+  }
+
+  property("section sizing preserves original bytes across versions and count framing") {
+    for (version <- Seq[Byte](1, 2, 3, 4); count <- Seq(1, 127, 128)) {
+      val txs = (0 until count).map(i => transaction(i, versionedRegister = false))
+      val section = BlockTransactions(Header.GenesisParentId, version, txs)
+      val original = originalBytes(section)
+      section.bytes shouldBe original
+      val payloadSize = txs.map(BlockTransactionsSerializer.transactionSize(_, version).toLong).sum
+      BlockTransactionsSerializer.sectionSize(version, count, payloadSize) shouldBe original.length.toLong
```
jozanek, line 69: `original` is a copy of the pre-PR writer inlined above, so this asserts the extraction is byte-identical — I'd expect it to pass against the unchanged base too.

What I can't find: the only round-trip property (`SerializationCoreTests.scala:76`) feeds `invalidBlockTransactionsGen`, which hardcodes `Header.InitialVersion` — so nothing parses a section at version ≥ 2, including the `VersionContext` branch you refactor. Would you be open to giving that generator a generated block version?

jozanek (changes requested):

ergo-core/src/test/scala/org/ergoplatform/modifiers/history/BlockTransactionsSpec.scala
jozanek: The parser round-trips at v1–4 across counts 1/127/128, plus the two v4 properties, close what I was actually worried about — nothing parsed a section above version 1 before. The rejection test also answered the SOption question from my top-level comment, where my reading was wrong.

`invalidBlockTransactionsGen` still hardcodes `Header.InitialVersion`, so the `SerializationCoreTests` property keeps seeing only v1. That reads as a nice-to-have now rather than a gap — fine as a separate change. Resolving this.

jozanek (commented):
Thanks — this addresses everything I raised, and the prefix recovery goes further than I asked. `sizeOf` beside `transactionsRoot` is exactly the shape I had in mind, and the injected-failure tests are much stronger evidence than the earlier wiring tests. I've resolved the four inline threads; my read is now that this is the right change.

One thing I'd still like settled before I lift the block. The body states the order — this PR before the adapted #2479 — and links the composition commit, which I appreciate. But #2479's own head is still `1b276802e`, carrying `maxSize = smallTx.size + smallFeeTx.size + 1` and `selected.map(_.size).sum should be < maxSize`. So the adaptation lives on your fork rather than on the PR that needs it. Could you push it to #2479, so the sequencing is visible on the PR itself? Different release targets are fine — it's the assertions I'd like to see move.

Three small things inline, none blocking.

On my earlier SOption claim: your v4 rejection test answers it, and my reading was wrong. Thanks for pinning it with a test.

ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
```diff
@@ -100,6 +101,10 @@ object BlockTransactions extends ApiCodecs {
 
   val modifierTypeId: NetworkObjectTypeId.Value = BlockTransactionsTypeId.value
 
+  /** Complete wire size, measured by the section writer in its existing block-version context. */
+  def sizeOf(txs: Seq[ErgoTransaction], blockVersion: Version): Int =
```
jozanek, line 110: `sizeOf` builds a `BlockTransactions`, which carries `assert(txs.nonEmpty, …)`, so `sizeOf(Seq.empty, v)` throws `AssertionError`. `checkedCandidate` guards for that at line 938, so callers here are fine — but this is the one method now public in `ergo-core`.

Would you be open to noting the precondition in the scaladoc?

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -984,51 +1021,65 @@ object CandidateGenerator extends ScorexLogging {
...
           }
         case None => // mempool is empty
-          current -> invalidTxs
+          finish
       }
     }
 
-    val res = loop(transactions, Seq.empty, None, Seq.empty)
+    // Vector prefixes share their storage; retaining fallback snapshots must not retain quadratic copies.
+    val res = loop(transactions, Vector.empty, 0L, None, Seq.empty, Nil)
```
jozanek: The comment says Vector prefixes share storage, so retaining a snapshot per accepted transaction stays sub-quadratic. I follow that for `acc :+ tx`, but each snapshot is `acc ++ lastFeeTx` — a concatenation rather than an append.

Have you measured retention for a full 512 KB block? If it holds, a line saying which operation preserves the sharing would save the next reader the same question.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -917,15 +922,40 @@ object CandidateGenerator extends ScorexLogging {
...
+    candidates: Iterator[(Seq[ErgoTransaction], Seq[ModifierId])], blockVersion: Byte, maxBlockSize: Int
+  ): (Seq[ErgoTransaction], Seq[ModifierId]) = {
+    if (!candidates.hasNext) Seq.empty -> Seq.empty
+    else {
+      val candidate = candidates.next()
+      val txs = candidate._1
+      (if (txs.isEmpty) Success(true) else candidateSizeWithinLimit(txs, blockVersion, maxBlockSize)) match {
+        case Success(true) => candidate
+        case result =>
+          log.warn(s"Cannot return transaction section ($result); restoring the previous accepted prefix")
```
jozanek: This fires on the ordinary oversize path too, where `$result` renders as `Success(false)` — so a normal size boundary logs at WARN with a value that reads like an error.

Would you be open to splitting the two: `debug` when the section is merely too large, `warn` when measurement actually failed?

jozanek (approved):

ergo-core/src/main/scala/org/ergoplatform/modifiers/history/BlockTransactions.scala
a-shannon: Documented in [#2520](https://github.com/ergoplatform/ergo/pull/2520) at `c9ddcaaa5`: `sizeOf` now states that `txs` must be nonempty and documents the `AssertionError` for an empty sequence. This documents the existing serializer precondition without changing it.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
a-shannon: The measurement confirmed your concern: Vector sharing alone did not prevent snapshots from retaining every historical fee transaction, each growing with the prefix. The [three-file correction in #2520](https://github.com/a-shannon/ergo/compare/d904fd1da9697ce6d5471e4416570c7bd0cf8ced...c9ddcaaa5207d5373a720a4399ea27489faa6a9e) retains the original costed prefix and reconstructs a historical fee only when fallback needs it. The measurement invoked the compiled recipe factory and checked prefix identity directly.

For 256 synthetic fee-producing transactions, retained inputs fell from 33,152 to 512. A 524,166-byte synthetic section with 2,864 ordinary transactions retained 2,865 transaction objects, 5,728 inputs and about 4.13 MB of traversed structure and payload. This is retained-object evidence for that fixture, not peak heap, transient allocations or a state/cost-valid full block.

The focused suite passed 19/19, including failed reconstruction and no historical reconstruction when the current candidate succeeds. [All eight current-head CI jobs passed](https://github.com/ergoplatform/ergo/actions/runs/34784284996). The requested full-section assertions are also now on [#2479](https://github.com/ergoplatform/ergo/pull/2479) itself at `f75bda15c`, with 16/16 local tests; its separate integration CI timeout remains unresolved.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
a-shannon: Split in [#2520](https://github.com/ergoplatform/ergo/pull/2520) at `c9ddcaaa5`: expected oversize recovery logs at DEBUG; reconstruction or measurement failures log at WARN. Expected `Success(false)` no longer produces the warning.

jozanek (approved):
All three points are addressed, and the retention one turned out to be a real bug rather than a documentation gap — thanks for measuring it. The recipe thunk is a good shape: the happy path materialises nothing, and `reconstructed shouldBe 0` pins that.

#2479 at `f75bda15c` moves both assertions to complete-section measurement, which was the one thing I was holding the block for. Its integration CI timeout is its own matter.

Approving. One thing worth its own issue — not yours, not blocking: the no-fee-proposition branch recurses with `acc = newTxs ++ lastFeeTx` while still carrying `lastFeeTx`, so the next `currentCosted` can include it twice. It predates this PR, but this PR now measures and returns `currentCosted`. I haven't verified reachability.

PR #2524, opened 2026-09-06
Propagate snapshot writes and protect all retained content

a-shannon:
Snapshot availability now follows successful subtree, manifest and metadata writes. The asynchronous persistence owner observes failures and skips pruning after a failed write.

Pruning protects content referenced by every retained snapshot, including a manifest advertised at several heights. If a retained manifest is missing or cannot be read, pruning leaves content and metadata intact. Previously, protection considered only the newest snapshot.

Validation: JDK 8, Scala 2.12.20; 29 tests pass across five affected suites with the pure-Java LevelDB backend. Ordinary producer-generated snapshots cover each write boundary, the actual asynchronous caller, shared retained content, unavailable retained manifests, duplicate-height references, retention limits and snapshot reconstruction. Independent source and test review completed.

This change does not make content and metadata one transaction or add crash recovery. Pruning write-failure handling and concurrent scheduling remain separate concerns. Snapshot acquisition cleanup is addressed in [#2519](https://github.com/ergoplatform/ergo/pull/2519).

All eight [CI checks](https://github.com/ergoplatform/ergo/actions/runs/34064391786) pass at `03f8ce889c93594fc5ea4085147ebccb3533568d`.

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/state/SnapshotsDb.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/SnapshotsDb.scala
@@ -11,7 +11,6 @@ import scorex.util.ScorexLogging
 import scorex.util.encode.Base16
 
 import scala.collection.mutable
-import scala.collection.mutable.ArrayBuffer
 import scala.util.{Failure, Success, Try}
 
 /**
@@ -45,17 +44,24 @@ class SnapshotsDb(store: LDBKVStore) extends ScorexLogging {
 
     if (manifests.size > toStore) {
 
-      val lastManifestBytesOpt = store.get(manifests.last._2)
-      val lastManifestSubtrees =
-        lastManifestBytesOpt
-          .flatMap(bs => manifestSerializer.parseBytesTry(bs).toOption)
-          .map(_.subtreesIds)
-          .getOrElse(ArrayBuffer.empty)
-          .map(Algos.encode)
-          .toSet
-
       val toPrune = manifests.dropRight(toStore)
       val toLeave = manifests.takeRight(toStore)
+      // Resolve every retained reference before deleting anything. Missing references
+      // cannot be treated as an empty tree, since other snapshots may share its content.
+      val retainedKeys = Try {
+        toLeave.iterator.flatMap { case (_, manifestId) =>
+          val bytes = store.get(manifestId).getOrElse {
+            throw new IllegalStateException(s"Retained manifest ${Base16.encode(manifestId)} not found")
+          }
+          val manifest = manifestSerializer.parseBytesTry(bytes).get
+          Iterator(Algos.encode(manifestId)) ++ manifest.subtreesIds.iterator.map(Algos.encode)
+        }.toSet
+      } match {
+        case Success(keys) => keys
+        case Failure(e) =>
+          log.error("Can't resolve retained snapshots; skipping pruning", e)
+          return
+      }
 
       toPrune.foreach { case (h, manifestId) =>
         val pt0 = System.currentTimeMillis()
@@ -66,10 +72,12 @@ class SnapshotsDb(store: LDBKVStore) extends ScorexLogging {
                 val keys = mutable.ArrayBuilder.make[Array[Byte]]()
                 val subtrees = m.subtreesIds
                 keys.sizeHint(subtrees.size + 1)
-                keys += manifestId
-                // filter out subtrees which are the same in the latest version of tree snapshot
+                if (!retainedKeys.contains(Algos.encode(manifestId))) {
+                  keys += manifestId
+                }
+                // A fragment may be shared with any retained snapshot.
                 subtrees.foreach { subtreeId =>
-                  if (!lastManifestSubtrees.contains(Algos.encode(subtreeId))) {
+                  if (!retainedKeys.contains(Algos.encode(subtreeId))) {
                     keys += subtreeId
                   }
                 }
@@ -109,10 +117,9 @@ class SnapshotsDb(store: LDBKVStore) extends ScorexLogging {
                     height: Height,
                     expectedRootHash: Array[Byte],
                     manifestDepth: Byte = ManifestSerializer.MainnetManifestDepth): Try[Array[Byte]] = {
-    pullFrom.dumpSnapshot(store, manifestDepth, expectedRootHash).map { manifestId =>
+    pullFrom.dumpSnapshot(store, manifestDepth, expectedRootHash).flatMap { manifestId =>
       val si = readSnapshotsInfo.withNewManifest(height, Digest32 @@ manifestId)
-      writeSnapshotsInfo(si)
-      manifestId
+      writeSnapshotsInfo(si).map(_ => manifestId)
     }
   }
 
--- a/src/main/scala/org/ergoplatform/nodeView/state/UtxoSetSnapshotPersistence.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/state/UtxoSetSnapshotPersistence.scala
@@ -62,10 +62,12 @@ trait UtxoSetSnapshotPersistence extends ScorexLogging {
       Future {
         log.info("Started work within future")
         val ft0 = System.currentTimeMillis()
-        dumpSnapshot(height, expectedRootHash)
+        dumpSnapshot(height, expectedRootHash).get
         snapshotsDb.pruneSnapshots(ergoSettings.nodeSettings.utxoSettings.storingUtxoSnapshots)
         val ft = System.currentTimeMillis()
         log.info("Work within future finished in: " + (ft - ft0) + " ms.")
+      }(scala.concurrent.ExecutionContext.Implicits.global).failed.foreach { e =>
+        log.error(s"Snapshot persistence failed at height $height", e)
       }(scala.concurrent.ExecutionContext.Implicits.global)
       val ms = System.currentTimeMillis()
       log.info("Main thread time to dump utxo set snapshot: " + (ms - ms0) + " ms.")
--- a/src/test/scala/org/ergoplatform/nodeView/state/SnapshotsDbSpecification.scala
+++ b/src/test/scala/org/ergoplatform/nodeView/state/SnapshotsDbSpecification.scala
@@ -31,7 +31,7 @@ class SnapshotsDbSpecification extends ErgoCorePropertyTest {
     }
   }
 
-  property("pruneSnapshots choosing snapshots correctly") {
+  property("pruneSnapshots preserves metadata when retained manifests are unavailable") {
     forAll(Gen.nonEmptyListOf(modifierIdGen)) { manifestIds =>
       val (si, db) = seededDatabase(manifestIds)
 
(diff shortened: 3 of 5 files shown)


cafebedouin:
Review carried out by Claude (Anthropic, Claude Opus 5.5) for cafebedouin, using forkbench; executed the changed spec classes with and without the fix, and a devnet run on the merge base and on this branch.

**Supported: on a live devnet this branch keeps every fragment of every retained snapshot, where the merge base loses part of the older one; both changed test classes fail without the fix.**

Executed: `sbt testOnly` of `SnapshotPersistenceSpecification` and `SnapshotsDbSpecification` (JDK 8) on `03f8ce88`, and on `03f8ce88` with its three production files restored to `5528ef56`: both classes pass with the fix and fail without it; 14 of 19 tests fail without it, the other 5 are controls that pass both ways.

Live (OpenJDK 21), one run per build, same workload, each build mining its own chain: three nodes, `makeSnapshotEvery = 64`, two with `storingUtxoSnapshots = 2`, about 24,000 boxes minted. On both builds each storing node took seven snapshots (heights 63 to 447), pruned five, kept 383 and 447, logged no snapshot processing error, and served a third node every chunk of the height-255 snapshot. With the nodes stopped, each storing node's snapshot database was read: is every subtree each advertised manifest lists present?

| build | height 383 (older retained) | height 447 |
|---|---|---|
| merge base, both nodes | 226 of 12,592 subtrees missing | 0 of 12,616 missing |
| this branch, both nodes | 0 of 12,290 missing | 0 of 12,312 missing |

Control: after one subtree is deleted from a copy, the same check reports it missing. What would make this wrong: a missing subtree on this branch after more than the five pruning cycles run here.

Not run: a write failure or a missing retained manifest on a live node (the tests cover both); mainnet-sized snapshots; other platforms; this repository's CI. Bootstrap completion was not compared: the merge base predates #2496.

Recommended: merge.

```
class | with change | without change | verdict
org.ergoplatform.nodeView.state.SnapshotPersistenceSpecification | pass | fail | fails-without/passes-with
org.ergoplatform.nodeView.state.SnapshotsDbSpecification | pass | fail | fails-without/passes-with
per test: guarding=14 passes-both=5 fails-both=0

base A: Error-during-snapshot-processing=0 futures-finished=7 prune-finished=5 pruned-heights=63 127 191 255 319
candidate A: Error-during-snapshot-processing=0 futures-finished=7 prune-finished=5 pruned-heights=63 127 191 255 319
base C: Downloading manifest for height 255; Downloaded or waiting 12522 chunks out of 12522; transfer-finished=1
candidate C: Downloading manifest for height 255; Downloaded or waiting 12230 chunks out of 12230; transfer-finished=1

== base-A
height 383 manifest 8dc78608c7f03829 subtrees 12592 missing 226
height 447 manifest a8af6b3d74ce30c5 subtrees 12616 missing 0
== candidate-A
height 383 manifest 651ce7c3478799e0 subtrees 12290 missing 0
height 447 manifest 655266f5a36e430e subtrees 12312 missing 0
== control-candidate-A (one retained subtree deleted first)
height 383 manifest 651ce7c3478799e0 subtrees 12290 missing 1
height 447 manifest 655266f5a36e430e subtrees 12312 missing 1
```

(B nodes identical to A in every row.)

</details>

<details><summary>Subtree check: RetainedCheck.java (46 lines, sha256 9975f2fcc3b69d83f9cb9f8c9eebe305fcf6ffb6fa3f6f880925623fc231fb2f)</summary>

Run on copies of each stopped storing node's `snapshots` directory; it only reads, except that `delete-one` first removes one subtree from the copy (the control). Compiled and run with JDK 8 against a node jar (the released 6.0.6 jar here): `javac -cp ergo.jar RetainedCheck.java && java -cp ergo.jar:. RetainedCheck <copy-of-snapshots-dir> [delete-one]`.

```java
// RetainedCheck: offline check of a stopped ergo node's snapshots database (a copy of <data>/snapshots).
// For every manifest listed in the snapshots metadata: is the manifest present, and is every subtree it lists present?
// Optional 2nd arg "delete-one": first delete one subtree key of the newest retained manifest (capability control).
// usage: java -cp <node.jar>:. RetainedCheck <snapshots-dir> [delete-one]
import org.ergoplatform.nodeView.state.SnapshotsDb;
import org.ergoplatform.serialization.ManifestSerializer;
import scorex.db.LDBFactory;
import scorex.db.LDBKVStore;
import scala.collection.JavaConverters;
import java.util.*;

public class RetainedCheck {
  static String hex(byte[] b) { StringBuilder s = new StringBuilder(); for (byte x : b) s.append(String.format("%02x", x)); return s.toString(); }
  @SuppressWarnings({"unchecked","rawtypes"})
  static List<byte[]> subs(ManifestSerializer ser, byte[] bytes) {
    scorex.crypto.authds.avltree.batch.serialization.BatchAVLProverManifest man = (scorex.crypto.authds.avltree.batch.serialization.BatchAVLProverManifest) ser.parseBytesTry(bytes).get();
    return new ArrayList<byte[]>((List) JavaConverters.seqAsJavaListConverter(man.subtreesIds()).asJava());
  }
  public static void main(String[] a) throws Exception {
    LDBKVStore store = LDBFactory.createKvDb(a[0]);
    SnapshotsDb db = new SnapshotsDb(store);
    Map<Object, byte[]> m = JavaConverters.mapAsJavaMapConverter(db.readSnapshotsInfo().availableManifests()).asJava();
    TreeMap<Integer, byte[]> sorted = new TreeMap<>();
    for (Map.Entry<Object, byte[]> e : m.entrySet()) sorted.put((Integer) e.getKey(), e.getValue());
    ManifestSerializer ser = ManifestSerializer.defaultSerializer();
    if (a.length > 1 && a[1].equals("delete-one")) {
      byte[] mid = sorted.lastEntry().getValue();
      List<byte[]> ids = subs(ser, store.get(mid).get());
      byte[] victim = ids.get(ids.size() / 2);
      store.remove(new byte[][]{victim}).get();
      System.out.println("CONTROL deleted subtree " + hex(victim).substring(0, 16) + " of height " + sorted.lastKey());
    }
    int totalMissing = 0;
    for (Map.Entry<Integer, byte[]> e : sorted.entrySet()) {
      scala.Option<byte[]> mb = store.get(e.getValue());
      if (mb.isEmpty()) { System.out.println("height " + e.getKey() + " manifest " + hex(e.getValue()).substring(0, 16) + " MISSING"); totalMissing++; continue; }
      List<byte[]> ids = subs(ser, mb.get());
      int missing = 0;
      for (byte[] id : ids) if (store.get(id).isEmpty()) missing++;
      totalMissing += missing;
      System.out.println("height " + e.getKey() + " manifest " + hex(e.getValue()).substring(0, 16) + " subtrees " + ids.size() + " missing " + missing);
    }
    System.out.println("TOTAL_MISSING " + totalMissing);
    store.close();
  }
}
```

</details>

src/test/scala/org/ergoplatform/nodeView/state/SnapshotsDbSpecification.scala
```diff
@@ -31,7 +31,7 @@ class SnapshotsDbSpecification extends ErgoCorePropertyTest {
     }
   }
 
-  property("pruneSnapshots choosing snapshots correctly") {
+  property("pruneSnapshots preserves metadata when retained manifests are unavailable") {
```
jozanek: Seeded ids here have no stored manifests, so for `toStore > 0` this property only reaches the fail-closed branch. Which heights survive pruning isn't exercised on random input anymore. The new spec uses ≤4 ascending heights, which I believe small Scala maps keep in order even without `sortBy(_._1)`. Would you be open to storing a parseable manifest per seeded id and keeping the `takeRight(toStore)` assertion?

src/main/scala/org/ergoplatform/nodeView/state/UtxoSetSnapshotPersistence.scala
```diff
@@ -62,10 +62,12 @@ trait UtxoSetSnapshotPersistence extends ScorexLogging {
       Future {
         log.info("Started work within future")
         val ft0 = System.currentTimeMillis()
-        dumpSnapshot(height, expectedRootHash)
+        dumpSnapshot(height, expectedRootHash).get
         snapshotsDb.pruneSnapshots(ergoSettings.nodeSettings.utxoSettings.storingUtxoSnapshots)
         val ft = System.currentTimeMillis()
         log.info("Work within future finished in: " + (ft - ft0) + " ms.")
+      }(scala.concurrent.ExecutionContext.Implicits.global).failed.foreach { e =>
+        log.error(s"Snapshot persistence failed at height $height", e)
```
jozanek: The comment at L58–60 treats a root-hash change before the DB snapshot is taken as expected. With the new `.get`, that case now lands here as ERROR with a stack trace, which operators may read as a fault. Would you consider logging the `Root node hash changed` case at WARN without the trace?

src/main/scala/org/ergoplatform/nodeView/state/SnapshotsDb.scala
```diff
@@ -45,17 +44,24 @@ class SnapshotsDb(store: LDBKVStore) extends ScorexLogging {
...
-          .map(_.subtreesIds)
-          .getOrElse(ArrayBuffer.empty)
-          .map(Algos.encode)
-          .toSet
-
       val toPrune = manifests.dropRight(toStore)
       val toLeave = manifests.takeRight(toStore)
+      // Resolve every retained reference before deleting anything. Missing references
+      // cannot be treated as an empty tree, since other snapshots may share its content.
+      val retainedKeys = Try {
```
jozanek: Optional, readability only: `throw` and `.get` inside `Try` are correct here, but I had to trace the catch boundary to confirm it. A fold keeps it all in `Try`:

```scala
toLeave.foldLeft(Try(Set.empty[String])) { case (acc, (_, id)) =>
  for {
    keys  <- acc
    bytes <- store.get(id).toRight(new IllegalStateException(s"Retained manifest ${Base16.encode(id)} not found")).toTry
    m     <- manifestSerializer.parseBytesTry(bytes)
  } yield keys + Algos.encode(id) ++ m.subtreesIds.map(Algos.encode)
}
```

No action needed if you prefer the current form.

jozanek (commented):
My read of the purpose: with mainnet's `storingUtxoSnapshots = 2`, pruning protected only the newest manifest. So the older retained snapshot could lose shared subtrees, and a peer bootstrapping from it stalls on the unanswered chunks until restart. Failed writes could also be advertised. Protecting every retained id and publishing metadata last look like the right shape to me, and the fail-closed skip clears itself within `toStore` cycles.

Questions I'm holding:
1. Over 52,224 mainnet blocks nearly every depth-14 subtree changes, so I'd expect little shared content there (the devnet run had far lower churn). Has anyone seen missing chunks on mainnet? The answer changes how I'd describe the fix, not whether to merge it.
2. After a mid-dump write failure, the written subtrees are in no metadata and are never pruned. Would a note in the PR body or a follow-up issue cover that?

The property-test point inline is what would move me to approve.

PR #2526, opened 2026-09-06
Remove mempool conflicts for every applied transaction

a-shannon:
Block application now removes input conflicts even when the applied transaction was never in the local mempool. Previously, the batch cleanup skipped such a transaction entirely. The helper preserves removal statistics when the applied transaction itself is absent.

Validation: JDK 8, Scala 2.12.20; 46 tests pass across five mempool suites. Five direct cases cover absent winners, unchanged statistics without conflicts, multiple inputs, repeated cleanup, index contents and retaining an unconfirmed child of a confirmed parent. Four cases fail against the original implementation. Independent source and test review completed.

This corrects immediate input-conflict cleanup. Asynchronous revalidation remains responsible for subsequent invalid-descendant cleanup. The lifecycle correction in [#2517](https://github.com/ergoplatform/ergo/pull/2517) has a separate invariant; combined runtime validation remains pending if they are integrated together.

All eight checks pass on the current head: [CI results](https://github.com/ergoplatform/ergo/actions/runs/34065040658).

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/mempool/ErgoMemPool.scala
@@ -116,7 +116,8 @@ class ErgoMemPool private[mempool](private[mempool] val pool: OrderedTxPool,
       new ErgoMemPool(mp.pool.remove(tx), mp.updateStatsOnRemoval(tx), sortingOption)
     }
 
-    val poolWithoutTx = removeTx(this, tx)
+    // An applied transaction may arrive through a block without ever entering the pool.
+    val poolWithoutTx = if (contains(tx.id)) removeTx(this, tx) else this
     val doubleSpentTransactionIds = tx.inputs.flatMap(i =>
       poolWithoutTx.pool.inputs.get(i.boxId)
     ).toSet
@@ -133,11 +134,7 @@ class ErgoMemPool private[mempool](private[mempool] val pool: OrderedTxPool,
     */
   def removeWithDoubleSpends(txs: TraversableOnce[ErgoTransaction]): ErgoMemPool = {
     txs.foldLeft(this) { case (memPool, tx) =>
-      if (memPool.contains(tx.id)) { // tx could be removed earlier in this loop as double-spend of another tx
-        memPool.removeTxAndDoubleSpends(tx)
-      } else {
-        memPool
-      }
+      memPool.removeTxAndDoubleSpends(tx)
     }
   }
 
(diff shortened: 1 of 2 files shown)


cafebedouin:
Review carried out by Claude (Anthropic, Claude Opus 5.5) for cafebedouin, using forkbench; executed the new spec with and without the production change, and a two-node payment run.

**Confirms the description: 4 of 5 new cases fail without the production change, 5 of 5 pass with it. Two non-blocking notes.**

Executed: `AppliedTransactionConflictsSpec` (JDK 8) on `ef1d9379`, and on the same tree with only `ErgoMemPool.scala` restored from the merge base `5528ef56`. Without the change, "remove applied transaction conflicts when the applied transaction is absent" and "remove conflicts for every input" keep the conflict (`AppliedTransactionConflictsSpec.scala:37`, `:64`), "preserve statistics" gets `takenTxns` 0 instead of 7 (`:53`), and "count a conflict sharing several inputs only once" leaves one transaction (`:74`); "remove a present applied transaction once" passes both ways. A two-node payment run (ten payments, one run per jar, JDK 21) accepted and confirmed 10 of 10 on the merge base and on this branch; the released 6.0.6 jar does the same on this host. It has no conflicting transactions: the benign path only. Not run: the conflict path on a network.

**[Integration] Heads-up: #2502 and #2505 fix the same reset differently**

Observed: both carry commit `518d8bea`, which also drops the guard in `removeWithDoubleSpends` but avoids the statistics reset inside `removeTx` by returning the pool unchanged when `pool.remove` returns the same instance (`ErgoMemPool.scala:132-140` there); this branch checks `contains(tx.id)` first (`ErgoMemPool.scala:120`). Same invariant, two forms, one author. Moot if one of those lines is to be rebuilt on the other.

Recommended: settle on one form across the three. (read)

**[Observability] Removed conflicts are counted as confirmed in the fee statistics, and the new spec pins it**

Observed: each double-spend removed here goes through `removeTx` and `updateStatsOnRemoval` (`ErgoMemPool.scala:104-108,128`), which increments `takenTxns` (read by `getExpectedWaitTime`, `:348-349`) and adds the conflict's fee-per-factor to the histogram read by `getRecommendedFee` (`:317-318`). `MemPoolStatistics.add` is documented as "called when transaction is taken from mempool and placed in blockchain" (`MemPoolStatistics.scala:21`). Pre-existing for conflicts of pooled transactions; this branch extends it to absent ones, and the spec asserts it (`takenTxns shouldBe 8` at `AppliedTransactionConflictsSpec.scala:40` and `:75`, `9` at `:65`; passing).

Recommended: decide whether discarded conflicts should count; if not, skip `stats.add` for them; if so, amend the `add` comment. (executed for the assertions; read for the consumers)

<details><summary>Summary of runs</summary>

```
spec (JDK 8)                        with change   without change
AppliedTransactionConflictsSpec     5 passed      1 passed, 4 failed

payment run (JDK 21, 1 run each)    accepted  confirmed  node version
merge base 5528ef56                 10        10         6.0.5
this branch                         10        10         6.0.5-1-e66f2c7a-SNAPSHOT (same tree as ef1d9379)
released 6.0.6 (control, 2 runs)    10, 10    10, 10     6.0.6
```
jars (sha256): merge base `2a7e2978…`, this branch `43b4cd36…`, released 6.0.6 `21b90239…`.
</details>

Issue #2533, opened 2026-09-07
Tracking: node corrections, review priorities and integration dependencies


a-shannon:
This issue tracks A. Shannon's public Ergo contributions, review priorities, exact heads and integration dependencies. Master, release branches, weak blocks and other repositories remain separate tracks. All inventory URLs were checked: 116 open PRs across 18 repositories, plus 66 closed PRs (37 merged, 29 unmerged). Head, review and CI details were refreshed for 23 entries; the other 93 retain historical detailed metadata. Shared #2535 S2 `99f8bd2ff`, #2511, #2536 and #2546 are 8/8. #2424 `72140e5de` and combined `8e475916a` are also 8/8 on their exact published heads. #2511/#2536 retain their green S1 heads. #2421's refreshed two-file release increment is 8/8 and conflict-free. #2544 remains 7/8 with an unlocalized ForkResolution timeout; Ergologica/ergo#1 has no triggered checks.

## Proposed release priorities

Kushti proposes prioritizing [#2496](https://github.com/ergoplatform/ergo/pull/2496) and P2P changes for **6.0.6**, with [#2424](https://github.com/ergoplatform/ergo/pull/2424) considered for **6.0.7 or 6.0.8**. The exact P2P selection and #2424's later release remain unconfirmed. On 10 September, both PRs still target `master` without a milestone.

- **Immediate review:** #2496/bootstrap and the selected P2P changes. Review shared prerequisites once, in their dependency order, only where those selected changes require them.
- **Deferred under this proposal:** #2424 and its [combined snapshot-wallet integration](https://github.com/a-shannon/ergo/pull/2). Preserve their reviewed heads and validation evidence for the later release review.
- **Bootstrap follow-ups:** assess [#2545](https://github.com/ergoplatform/ergo/pull/2545) then [#2546](https://github.com/ergoplatform/ergo/pull/2546) alongside #2496's dependency review; their release inclusion remains to be decided. This deferral does not assign a release to independent wallet or persistence PRs.

## Review and integration now

The proposed release priorities above govern immediate review focus. The queues below retain the risk findings, prerequisites and deferred work.

| Priority / queue | Owner | Action |
| --- | --- | --- |
[...]

PR #2535, opened 2026-09-08
Extract shared test fixtures and convergence observations

a-shannon:
Shared fixtures and observation contracts used by several corrections belong here, so maintainers can review and reuse them once. This PR ports that support to **v6.0.6**, based on `8995ff243b140e7fe716bd3204423dd35690c2e4`.

Head: `efbd1e917fe74f0f680a2e3c9e4d93595a59b06a`. This PR owns fourteen test/support files. Its [cumulative release diff](https://github.com/a-shannon/ergo/compare/8995ff243b140e7fe716bd3204423dd35690c2e4...efbd1e917fe74f0f680a2e3c9e4d93595a59b06a) contains twenty-two files because it also includes explicit production prerequisites owned by [#2511](https://github.com/ergoplatform/ergo/pull/2511) and [#2549](https://github.com/ergoplatform/ergo/pull/2549), with their direct tests. Build and CI configuration are unchanged.

**Integration order: land the #2511 and #2549 prerequisites before the remaining #2535 support.** The two production prerequisites can be reviewed in parallel. Review the small [shared startup-reader commit](https://github.com/a-shannon/ergo/compare/c313356950fe69ef406c2ee031204079a05ea7d7...76e5ec4bfc377334fdced4bd60a61df06e6b4cb6) once before #2511's direct tests, then its [seven-file synchronization increment](https://github.com/a-shannon/ergo/compare/76e5ec4bfc377334fdced4bd60a61df06e6b4cb6...d62c945a4d6314c54a7b57f18a94aee8fd258aa7). Neither requires the full #2535 branch first. Descendants reuse these exact commits, and their cumulative upstream diffs should not be treated as new copies to review. After prerequisites merge, their confirmed upstream changes can be removed from this PR's comparison.

The [latest seven-file increment](https://github.com/a-shannon/ergo/compare/e640962bd8184ad897e20953f0dc01a81163664f...efbd1e917fe74f0f680a2e3c9e4d93595a59b06a) reuses #2511-owned commit `d62c945a4d6314c54a7b57f18a94aee8fd258aa7`. It binds the V2 cache to selected tip and summary mode, retains an available genesis anchor in full summaries, and schedules missing body sections from a retained common ancestor within the existing download ceiling. Its four test files retain the original cases and deadlines; the two full-summary oracles now require the exact five header IDs. The separate [two-file cache wake-up increment](https://github.com/a-shannon/ergo/compare/d62dc6cd9a03d45a5168f66b1252c179446794b4...e640962bd8184ad897e20953f0dc01a81163664f) reuses #2549-owned commit [`a772a5cbf`](https://github.com/a-shannon/ergo/commit/a772a5cbf666584db3ed981838dbc11223d502ed).

The prerequisite retries already-cached block sections after remote headers arrive. Previously the holder retried its header cache at that point but retried body sections only after another body message. A body awaiting its header could therefore remain cached when no further body arrived. #2549 owns this correction and the UTXO/Digest regressions that send cached bodies before the header and require application without resending them. Both production and regression files here exactly match #2549's release head `27eb9d84cf4501f580703ac229ad646af405b251`. The prerequisite is independent of this support branch and introduces no dependency cycle.

The previous native [run at `e640962bd`](https://github.com/ergoplatform/ergo/actions/runs/34731989199) passed seven jobs and 50/51 integration cases, with one additional ignored case. DeepRollBackSpec established the fully applied seed at height 21, then failed final convergence with tips 246/69. The preceding [run at `d62dc6cd9`](https://github.com/ergoplatform/ergo/actions/runs/34730855804) failed the seed bound at full heights 21/17 despite a common selected header at 21. The independently established cache wake-up defect alone did not close both scenarios; no exclusive cause is inferred from those logs.

The current tree is exactly `2294654428b7f8b9f7fc821343394ecb3acc3128`. Native upstream [CI for head `efbd1e917`](https://github.com/ergoplatform/ergo/actions/runs/34734047921) passed all **eight jobs**. The same tree at [fork PR #3](https://github.com/a-shannon/ergo/pull/3), head `168bd7e16860d3e3a6fbae80a000a4478f19c007`, also passed [eight CI jobs](https://github.com/a-shannon/ergo/actions/runs/34732895475), including full integration. Fork #3 was then closed as superseded because its exact source and successful native CI are now represented here; its history remains available. Normal focused validation passed **49/49**: synchronizer 26, summary cache 4, history continuation 6, scheduling 11 and header/body wake-up 2. The extracted common prerequisite also passed normal compilation and 47 direct tests. Independent reviews verified both the extraction and actual composition. The integration assertions and deadlines are unchanged by this increment.

The shared support includes funded transaction generators, the mempool sorting fixture, per-holder data and keystore isolation with blocking actor teardown, and candidate reward setup that binds the acknowledged transaction to its applied block, unique reward and current UTXO state. The sorting support reuses [#2480](https://github.com/ergoplatform/ergo/pull/2480)'s fixture without its production graph changes. Descendant-owned bootstrap, genesis and candidate behavior corrections remain with their owners.

The release keeps all **19 CandidateGeneratorSpec cases**. The earlier master result covered 20 because master also contains a full-block-id invalidation scenario absent from this release. That scenario and the master-only ViewHolder case are not added by this port. UtxoStateNodesSyncSpec already contained the selected-header observer on release; its new change is the failure-only diagnostic call and supporting helper/tests.

The [portable wallet fixture commit `d98d9e1d2`](https://github.com/a-shannon/ergo/commit/d98d9e1d268f978b710e4276676f05971e8254df) is reused here and by [#2507](https://github.com/ergoplatform/ergo/pull/2507). The [release follow-up](https://github.com/a-shannon/ergo/compare/80460b7af09263299e911fdcd5208a505da64364...a9af9e883517c73b1d21e497801d0a176eb0c094) changes only the wallet fixture because the other three portable integration files were already present. Four initialization/restore cases own their node and secret directories and close retained handles; direct-secret-unlock and supplied-cache flows remain intact. All 17 release wallet cases pass, including four change-address cases absent from the portable base. The helper came from [#2543](https://github.com/ergoplatform/ergo/pull/2543), whose generation and key-cache behavior remains there.

The [diagnostic increment](https://github.com/a-shannon/ergo/compare/a9af9e883517c73b1d21e497801d0a176eb0c094...7d33130cb7e16576fa37f0715aaf59d8ec53a6e7) reuses portable commit [`5a360fa48`](https://github.com/a-shannon/ergo/commit/5a360fa48c7abd9b9d25a4d00728b10c782e4031), also an ancestor of #2507 and #2543. Only after a decided UTXO convergence timeout, typed observations collect heights, work, genesis/header IDs, mining and peer counts within a separate additional 2.5-second bound. The original exception always propagates, including collection or emission failure. Five pure tests cover projection, parallel collection, timing and failure preservation; successful observations incur no diagnostic calls.

The synchronizer's synthetic-history fixture owns its startup history/mempool replies consistently, preventing populated readers from being overwritten by an empty real-holder reply. Ordinary state/vault handling and the second fixture's real holder remain intact. A deterministic startup replay failed before the correction and passes afterward. All 26 cases remain; the two full-summary expectations now include the genesis anchor required by the explicit #2511 prerequisite, with their original three-second deadlines.

This PR owns the [portable synchronizer fixture increment](https://github.com/a-shannon/ergo/compare/c313356950fe69ef406c2ee031204079a05ea7d7...76e5ec4bfc377334fdced4bd60a61df06e6b4cb6), commit `76e5ec4bfc377334fdced4bd60a61df06e6b4cb6`, reused by [#2433](https://github.com/ergoplatform/ergo/pull/2433). The [historical ancestry increment from `7d33130` to `d62dc6cd9`](https://github.com/a-shannon/ergo/compare/7d33130cb7e16576fa37f0715aaf59d8ec53a6e7...d62dc6cd9a03d45a5168f66b1252c179446794b4) had zero tree change because the release already contained this correction. That historical identity does not describe the new production-prerequisite increment.

DeepRollBackSpec establishes two setup conditions:

- Before separation, A stops producing and restarts without mining while B can still fetch full sections. Both must report mining disabled and the same fully applied header/full-block tip, below height 50.
- During isolated mining, a phase-local zero connection limit prevents reconnection through the retained peer database and rejects incoming connections. Both nodes must report no connected peers. Final restarts restore the original non-mining network configurations.

The 50/200 height targets, two-minute seed bound, twenty-minute outer deadline, ten-minute final convergence bound, phase/peer diagnostics and existing sameBestBlock contract remain unchanged. The isolation regression uses the real settings reader and NetworkController, a scheduler positive control and an incoming TCP-close assertion. It fails with the previous known-peers-only setup and passes with the phase-local limit.

Earlier release validation passed **148 focused tests** across the unchanged support inputs: Candidate 19, Transaction 19, Mempool 27, funded generator 2, synchronizer 26, pure convergence contracts 32, isolation 1, wallet 17 and diagnostics 5. Unit/integration compilation and independent source review passed. The new 28-case composition run is a separate check and overlaps those earlier synchronizer cases.

Historical CI establishes useful limits. [Run `34724198094`](https://github.com/ergoplatform/ergo/actions/runs/34724198094) at `7d33130` had a pre-test Scala XML download reset and a separate DeepRollBack seed failure, with equal header height 20 but full heights 20/8. [Run `34722489051`](https://github.com/ergoplatform/ergo/actions/runs/34722489051) established the settled seed and isolation but later stalled at tips 249/70. Setup corrections alone did not close every observed stall. The separate [fork integration PR #3](https://github.com/a-shannon/ergo/pull/3) composes #2511 with the earlier `7d33130` fixture tree: its 47 direct tests and [eight CI jobs](https://github.com/a-shannon/ergo/actions/runs/34724671004) passed, including 51/51 integration tests with one ignored case, DeepRollBack convergence at height 247 and four-node UTXO agreement. That is evidence for its pinned composition, not a substitute for the current head's CI.

Historical master review anchors remain available: [original extraction `7c753910b`](https://github.com/a-shannon/ergo/compare/5528ef569a41ebccbc8658212e6ee3c97d990b96...7c753910bf28b7c69f69228e120e101de2d83bb7), [candidate isolation `b3ebc6d6f`](https://github.com/a-shannon/ergo/compare/7c753910bf28b7c69f69228e120e101de2d83bb7...b3ebc6d6f5a36514d28631d0852d1ba91c9f1781), [selected-header observation `e18df9692`](https://github.com/a-shannon/ergo/compare/b3ebc6d6f5a36514d28631d0852d1ba91c9f1781...e18df96928b123894d2ae8da6a6d5a13dc1ee731), [settled-seed condition `5eedfdc91`](https://github.com/a-shannon/ergo/compare/e18df96928b123894d2ae8da6a6d5a13dc1ee731...5eedfdc91d7a515f66822643e730b4ca9b3197df), and [synchronizer isolation `99f8bd2ff`](https://github.com/a-shannon/ergo/compare/5eedfdc91d7a515f66822643e730b4ca9b3197df...99f8bd2ff811a94219279da746d24ec807761900). The latter's [8/8 master CI](https://github.com/ergoplatform/ergo/actions/runs/34426013892) is historical evidence for that input.

Previous-candidate message corrections retain their ownership in [#2312](https://github.com/ergoplatform/ergo/pull/2312) and closed [#2340](https://github.com/ergoplatform/ergo/pull/2340). Review order, prerequisite pins and descendant refresh status are tracked in [#2533](https://github.com/ergoplatform/ergo/issues/2533).

### DIFF
--- a/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
+++ b/src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
@@ -2,7 +2,7 @@ package org.ergoplatform.it
 
 import java.io.File
 import java.util.concurrent.TimeoutException
-import com.typesafe.config.Config
+import com.typesafe.config.{Config, ConfigFactory}
 import io.circe.Json
 import org.ergoplatform.it.api.NodeApi.{NodeInfo, nodeInfoDecoder}
 import org.ergoplatform.it.container.{IntegrationSuite, Node}
@@ -48,6 +48,42 @@ class DeepRollBackSpec extends AnyFreeSpec with IntegrationSuite {
     .withFallback(allowLocalConfig)
 
   private val observations = new ConvergenceObservations
+  private val seedObservations = new ConvergenceObservations
+  @volatile private var lastSeedObservation = "Initial seed has not been sampled"
+
+  private def waitForSettledSeed(
+    nodeA: Node,
+    nodeB: Node,
+    timeout: FiniteDuration
+  ): Future[(NodeInfo, NodeInfo)] = {
+    def infoProbe(node: Node): seedObservations.Probe[NodeInfo] =
+      seedObservations.probe(node.singleGet("/info", _.setRequestTimeout(5000)).map { response =>
+        require(response.getStatusCode == 200, "Unexpected seed observation status")
+        node.ergoJsonAnswerAs[NodeInfo](response.getResponseBody)
+      })
+
+    val probeA = infoProbe(nodeA)
+    val probeB = infoProbe(nodeB)
+    def describe(result: Either[String, NodeInfo]): String = result.fold(
+      error => s"errorClass=$error",
+      info => s"headersHeight=${info.bestHeaderHeightOpt}; fullHeight=${info.bestBlockHeightOpt}; " +
+        s"headerId=${info.bestHeaderIdOpt.map(ConvergenceObservations.headerId)}; " +
+        s"fullId=${info.bestBlockIdOpt.map(ConvergenceObservations.headerId)}; mining=${info.isMining}")
+    seedObservations.until(timeout.fromNow, 1.second, 5.seconds) { budget =>
+      probeA.sample(budget).zip(probeB.sample(budget)).map { pair =>
+        lastSeedObservation = s"A=${describe(pair._1)}; B=${describe(pair._2)}"
+        log.info(s"Initial shared-chain readiness: $lastSeedObservation")
+        pair
+      }
+    } {
+      case (Right(a), Right(b)) =>
+        ConvergenceObservations.sameFullyAppliedNonMiningBlock(a, b, ErgoHistoryUtils.GenesisHeight)
+      case _ => false
+    }(
+      s"Initial chain did not settle with mining disabled and matching full/header tips; $lastSeedObservation"
+    ).map { case (a, b) => (a.toOption.get, b.toOption.get) }
+  }
+
   private case class NodeSnapshot(info: Option[NodeInfo])
   private val probes = scala.collection.mutable.Map.empty[
     Node, (observations.Probe[NodeInfo], observations.Probe[Int])]
@@ -139,31 +175,38 @@ class DeepRollBackSpec extends AnyFreeSpec with IntegrationSuite {
       val genesisAGen = Async.await(minerAGen.headerIdsByHeight(ErgoHistoryUtils.GenesisHeight)).head
       val genesisBGen = Async.await(minerBGen.headerIdsByHeight(ErgoHistoryUtils.GenesisHeight)).head
 
-      val minerAGenBestHeight = Async.await(minerAGen.fullHeight)
-      val minerBGenBestHeight = Async.await(minerBGen.fullHeight)
-
-      log.info("heightA: " + minerAGenBestHeight)
-      log.info("heightB: " + minerBGenBestHeight)
-
       genesisAGen shouldBe genesisBGen
       Async.await(observeNodes("initial shared chain", minerAGen, minerBGen))
 
-      // 2. Stop all nodes
+      // Freeze the producer while B can still retrieve every header's full block.
       docker.stopNode(minerAGen.containerId)
+      val minerASeed: Node = docker.startDevNetNode(minerAConfigNonGen,
+        specialVolumeOpt = Some((localVolumeA, remoteVolumeA))).get
+      val (seedA, seedB) = Async.await(waitForSettledSeed(minerASeed, minerBGen, 2.minutes))
+      val seedHeight = seedA.bestBlockHeightOpt.get
+      require(seedHeight < chainLength,
+        s"Initial shared chain already reached $seedHeight; isolated node B must mine to $chainLength")
+      log.info(s"Settled shared chain: heightA=$seedHeight, heightB=${seedB.bestBlockHeightOpt.get}")
+
+      // 2. Stop the restarted A and B only after both have the complete shared seed.
+      docker.stopNode(minerASeed.containerId)
       docker.stopNode(minerBGen.containerId)
 
-      val minerAIsolated: Node = docker.startDevNetNode(minerAConfig, isolatedPeersConfig,
+      val minerAIsolated: Node = docker.startDevNetNode(DeepRollBackSpec.isolatedMiningConfig.withFallback(minerAConfig), isolatedPeersConfig,
         specialVolumeOpt = Some((localVolumeA, remoteVolumeA))).get
 
       // 1. Let nodeA mine `chainLength + delta` blocks in isolation
       Async.await(minerAIsolated.waitForHeight(chainLength + delta))
 
-      val minerBIsolated: Node = docker.startDevNetNode(minerBConfig, isolatedPeersConfig,
+      val minerBIsolated: Node = docker.startDevNetNode(DeepRollBackSpec.isolatedMiningConfig.withFallback(minerBConfig), isolatedPeersConfig,
         specialVolumeOpt = Some((localVolumeB, remoteVolumeB))).get
       Async.await(observeNodes("isolated miners started", minerAIsolated, minerBIsolated))
 
       // 2. Let nodeB mine `chainLength` blocks in isolation
       Async.await(minerBIsolated.waitForHeight(chainLength, 100.millis))
+
+      Async.await(minerAIsolated.connectedPeers) shouldBe empty
+      Async.await(minerBIsolated.connectedPeers) shouldBe empty
 
       log.info("Mining phase done")
 
@@ -208,11 +251,19 @@ class DeepRollBackSpec extends AnyFreeSpec with IntegrationSuite {
       Await.result(result, 20.minutes)
     } catch {
       case error: TimeoutException =>
+        log.error(s"Deep rollback timed out; last initial-seed observation: $lastSeedObservation")
         log.error(s"Deep rollback timed out; last observation: $lastObservation")
         throw error
     } finally {
+      seedObservations.close()
       observations.close()
     }
   }
 
+}
+
+object DeepRollBackSpec {
+  // The retained peer database survives restarts. Disable automatic outgoing connections
+  // and incoming admission during mining; final restarts use the ordinary node configs.
+  private[it] val isolatedMiningConfig: Config = ConfigFactory.parseString("scorex.network.maxConnections = 0")
 }
--- a/src/it/scala/org/ergoplatform/it/util/UtxoSyncFailureDiagnostics.scala
+++ b/src/it/scala/org/ergoplatform/it/util/UtxoSyncFailureDiagnostics.scala
@@ -1,0 +1,60 @@
+package org.ergoplatform.it.util
+
+import io.circe.Json
+
+import java.util.concurrent.TimeoutException
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.duration._
+
+/** Additional observations after failure, with their own bound and no effect on the assertion. */
+object UtxoSyncFailureDiagnostics {
+  private val requestBudget = 2.seconds
+  private val collectionBudget = 2500.millis
+  private val numericFields = Seq("headersHeight", "fullHeight", "headersScore", "fullBlocksScore", "peersCount")
+  private val headerFields = Seq("bestHeaderId", "bestFullHeaderId", "genesisBlockId")
+
+  def project(node: Int, sampledAt: Long, response: Either[String, Json]): Json = {
+    val identity = Seq("node" -> Json.fromInt(node), "sampledAt" -> Json.fromLong(sampledAt))
+    def failed(errorClass: String): Json = {
+      val safeClass = if (errorClass.matches("[A-Za-z_$][A-Za-z0-9_$]{0,127}")) errorClass else "ObservationFailure"
+      Json.obj((identity :+ ("errorClass" -> Json.fromString(safeClass))): _*)
+    }
+    response match {
+      case Left(errorClass) => failed(errorClass)
+      case Right(body) if !body.isObject => failed("InvalidInfoResponse")
+      case Right(body) =>
+        def field(name: String)(decode: Json => Option[Json]): (String, Json) = {
+          val value = body.hcursor.downField(name).focus match {
+            case None => Json.Null
+            case Some(value) if value.isNull => Json.Null
+            case Some(value) => decode(value).getOrElse(Json.fromString("invalid"))
+          }
+          name -> value
+        }
+        val numbers = numericFields.map(name => field(name) { value =>
+          value.asNumber.flatMap(_.toBigInt).filter(_ >= 0).map(Json.fromBigInt)
+        })
+        val headers = headerFields.map(name => field(name) { value =>
+          value.asString.map(id => Json.fromString(ConvergenceObservations.headerId(id)))
+        })
+        val mining = field("isMining")(_.asBoolean.map(Json.fromBoolean))
+        Json.obj((identity ++ numbers ++ headers :+ mining): _*)
+    }
+  }
+
+  def rethrowAfterCapture(original: TimeoutException, requests: Seq[() => Future[Json]])
+                         (emit: String => Unit)(implicit ec: ExecutionContext): Unit = {
+    try {
+      val observations = new ConvergenceObservations
+      try {
+        val snapshots = requests.zipWithIndex.map { case (request, index) =>
+          observations.probe(Future(request()).flatMap(response => response))
+            .sample(requestBudget).map(response => project(index, System.currentTimeMillis(), response))
+        }
+        // This bound starts only after the synchronization assertion has already failed.
+        val result = Await.result(Future.sequence(snapshots), collectionBudget)
+        emit(Json.arr(result: _*).noSpaces)
+      } finally observations.close()
+    } finally throw original
+  }
+}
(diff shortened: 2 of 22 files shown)


cafebedouin:
Prepared with Claude (Anthropic), model Claude Fable 5.1, for cafebedouin.

**[Observability] Nothing to change here; a follow-up PR on top of #2535 is ready that makes its convergence timeouts name their cause.**

The follow-up (branch `cafebedouin/ergo:it-diagnostics-on-2535`; `src/it` and `src/test` only, plus two read-only helpers in the test `Docker.scala`; no production code, image or CI change; it uses this PR's `ConvergenceObservations`) adds a classifier over sampled node state and wires it into `UtxoStateNodesSyncSpec`, `DeepRollBackSpec` and `ForkResolutionSpec`: a timeout message names one cause and prints every node's heights, tips, peers and mining flag. Assertions in the first two are untouched; `ForkResolutionSpec` changes are listed below. It diagnoses failures and fixes none.

Executed locally on the v6.0.6 image, three runs of `UtxoStateNodesSyncSpec` unmodified: all three timed out at 15 min; the last, on the offered bytes, reported `CAUSE PERSISTENT_FORK: 4 nodes kept gaining blocks over the last 29 s on chains that never shared a tip`, node01 and node04 at 895 and node03 at 894 on different ids, 3 peers each. `ForkResolutionSpec` and `DeepRollBackSpec` passed once each. Read from CI logs: the three end states the classifier's cases reproduce (runs 35356023528, 35579493804, 34744409218).

Recommended: open the follow-up once #2535 lands. Is `ForkResolutionSpec`'s phase-4 predicate (all nodes agree at the fork height) the intended contract, or must every node match node 0's sample?

<details><summary>What the follow-up contains</summary>

- `ConvergenceDiagnosis` (`src/test`, pure, 23 cases): `PERSISTENT_FORK`, `EQUAL_HEIGHT_TIE`, `HEADERS_AHEAD_FULL_STUCK`, `LIGHTER_FORK_NOT_SWITCHING`, `NODE_LAGGING`, `NODE_DOWN`, `NODE_UNRESPONSIVE`, `NO_PEERS`, `CHAIN_STALLED`, `STILL_PROGRESSING`, `UNKNOWN`.
- `ConvergenceWatch` (`src/it`): samples `/info` and `/peers/connected` once a second for the whole wait (container state looked up when a node stops answering); on timeout, the cause, the state 30 s earlier and last, and a log tail for a node that is down or silent. In `UtxoStateNodesSyncSpec` the original exception is kept as the cause and `rethrowAfterCapture` runs unchanged; `DeepRollBackSpec` feeds the watch from the observations it already makes (no extra requests) and keeps its existing timeout text.
- `ForkResolutionSpec`: four named phases on one shared 15-minute deadline; outer await 30 s longer so the phase message is the one reported; polling 1 s instead of 100 ms; phase 4 requires all nodes to agree at the fork height rather than to match node 0's sample.
</details>

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistoryReader.scala
```diff
@@ -403,8 +403,14 @@ trait ErgoHistoryReader
...
       val headers = offsets.flatMap(offset => bestHeaderAtHeight(h - offset))
+      // Keep a shared starting point in full summaries when it is available locally.
+      val genesisAnchor = if (full) {
+        bestHeaderAtHeight(GenesisHeight).filterNot(genesis => headers.exists(_.id == genesis.id)).toSeq
+      } else {
+        Seq.empty
+      }
 
-      ErgoSyncInfoV2(headers)
+      ErgoSyncInfoV2(headers.toSeq ++ genesisAnchor)
```
jozanek: With genesis always appended, a fork deeper than 512 on a tall chain now finds its common point at genesis, and the reply is ids 2..400 the peer already has. In the split run on #2511, this file plus the scheduler switched 3/8, same as release. As on #2511: would a clamp, `max(GenesisHeight, h - offset)`, cover the (18, 249, 70) case and leave mainnet messages unchanged?

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -84,8 +84,21 @@ trait ToDownloadProcessor
...
+          headerChainBack(limit, fb.header, h => isInBestChain(h.id)).headOption
+            .filter(h => isInBestChain(h.id))
+        }
+        val fromHeight = commonAncestor
+          .map(h => Math.max(minimalFullBlockHeight, h.height + 1))
+          .getOrElse(fb.height + 1)
+        // Extending the scan backward must preserve forward progress beyond the existing full-chain tip.
+        val maxHeight = Math.min(fb.height.toLong + FullBlocksToDownloadAhead, Int.MaxValue.toLong).toInt
+        continuation(fromHeight, Map.empty, maxHeight)
       case Some(fb) =>
```
jozanek, line 89: The far branch now scans back, but this near-tip branch still starts at `fb.height - 100`, and the far branch only fires with a lead above 128. I might be missing a path, but a fork 101–200 deep with a lead ≤128 looks like it still never gets its sections scheduled. Could a test cover that case, or could the ancestor bound apply here too?

src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/ToDownloadProcessor.scala
```diff
@@ -84,8 +84,21 @@ trait ToDownloadProcessor
...
+          None
+        } else {
+          val parentSteps = Math.max(0L, nodeSettings.keepVersions.toLong)
+          val limit = Math.min(fb.height.toLong, parentSteps + 1L).toInt
+          headerChainBack(limit, fb.header, h => isInBestChain(h.id)).headOption
+            .filter(h => isInBestChain(h.id))
+        }
+        val fromHeight = commonAncestor
+          .map(h => Math.max(minimalFullBlockHeight, h.height + 1))
+          .getOrElse(fb.height + 1)
```
jozanek: When `fb` is off the best chain and no ancestor turns up within `keepVersions + 1`, this falls back to `fb.height + 1` with no log line. That is the stall this change targets, re-walked on every scheduling tick. Would you be open to a rate-limited warn on that path, so an operator can tell a resync is needed?

src/test/scala/org/ergoplatform/modifiers/mempool/ErgoNodeTransactionSpec.scala
```diff
@@ -326,34 +324,29 @@ class ErgoNodeTransactionSpec extends ErgoCorePropertyTest with ErgoCompilerHelp
...
-
-      val t0 = System.currentTimeMillis()
-      (1 to 250000).foreach(i => hf(s"$i"))
-      val t = System.currentTimeMillis()
-      t - t0
-    }
+    // This test used to calibrate a wall-clock budget by timing 250K Blake2b256 hashes and then
+    // assert that validation fits in it. On a loaded CI machine the calibration window and the
+    // measurement window get different shares of the CPU, so the assertions flipped at random
+    // (see issue #2095). What the node is actually protected by is the block cost limit, not the
```
jozanek: This file matches open #2455 byte for byte, and the description doesn't mention it. Carrying it in both means a conflict for whichever lands second. Which PR would you like to own it? If #2455, could it be dropped here?

src/it/scala/org/ergoplatform/it/DeepRollBackSpec.scala
```diff
@@ -100,34 +175,44 @@ class DeepRollBackSpec extends AnyFreeSpec with IntegrationSuite {
...
-
       genesisAGen shouldBe genesisBGen
+      Async.await(observeNodes("initial shared chain", minerAGen, minerBGen))
 
-      // 2. Stop all nodes
+      // Freeze the producer while B can still retrieve every header's full block.
       docker.stopNode(minerAGen.containerId)
+      val minerASeed: Node = docker.startDevNetNode(minerAConfigNonGen,
+        specialVolumeOpt = Some((localVolumeA, remoteVolumeA))).get
+      val (seedA, seedB) = Async.await(waitForSettledSeed(minerASeed, minerBGen, 2.minutes))
```
jozanek: The settled seed sets aside the case where B's headers are ahead of its bodies at separation (run 34724198094: header 20, full 20/8). A peer dropping mid-download is a real node state, and the scan-back should handle it. Could it stay as a second case? No change needed if another spec already covers it.

src/it/scala/org/ergoplatform/it/util/UtxoSyncFailureDiagnostics.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.concurrent.TimeoutException
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.duration._
+
+/** Additional observations after failure, with their own bound and no effect on the assertion. */
+object UtxoSyncFailureDiagnostics {
+  private val requestBudget = 2.seconds
+  private val collectionBudget = 2500.millis
+  private val numericFields = Seq("headersHeight", "fullHeight", "headersScore", "fullBlocksScore", "peersCount")
```
jozanek, line 13: These fields show the symptom (heights and scores differ). A snapshot of `/peers/syncInfo`, the per-peer `Unknown`/`Fork`/`Older` status, would point at the cause, e.g. a peer stuck in `Unknown`. Would you consider adding it here? A follow-up is fine too.

jozanek (commented):
I read this as shared test support plus a real harness fix: `maxConnections = 0` finally isolates DeepRollBackSpec's miners. On master's CI both still had a peer when isolated mining finished.

On current master it also carries two sync changes, the genesis anchor in `syncInfoV2` and the scan-back in `nextModifiersToDownload`. #2511 dropped both on 22 Sep, so they would land here under a fixtures title. Would you be open to moving them to their own PR with an issue, so they get protocol review? #2511's cache-only CI stalled at headers 262 / full 70, the scan-back's case, so that part looks needed.

`CandidateGeneratorSpec` conflicts, and master now has 3 cases this side lacks. `SyncInfoV2CacheSpec` also arrives in #2511.

I could not find a mainnet or testnet report of either stall.

I'd approve a split, rebased version that keeps those cases, with green CI.

jozanek (commented):
Thanks for cutting this down to test support. With the sync changes, CandidateGeneratorSpec, the #2455 file and the cache spec out, and the seed sequence restored, five of my inline comments are addressed.

One thing I'm holding: isolation is now real, and CI 35775823953 shows DeepRollBackSpec stuck at A 256/256, B 70/70, one peer each. On master, A rates B `Younger` and finds none of B's samples (70, 54) on its chain, so it sends an empty extension. B rates A `Older` and waits. I'd expect this on every run until a sync-point fix (clamp or anchor) and the scan-back land, and neither has a PR yet.

Would you be open to landing this after those, or marking the spec pending with a linked issue, so master's IT job stays green meanwhile? With either, I'd be glad to approve.

src/it/scala/org/ergoplatform/it/util/UtxoSyncFailureDiagnostics.scala
jozanek: Still relevant, I think: the new DeepRollBackSpec failure had no node logs, so its cause wasn't established. A `/peers/syncInfo` sample at timeout would have shown A rating B `Younger` and B rating A `Older`, which names the deadlock. Fine as a follow-up, and it would help DeepRollBackSpec's observations too.

cafebedouin: **[Observability] `/peers/syncInfo` is now sampled every round on the follow-up branch, and the Younger/Older standoff has its own cause.**

`cafebedouin/ergo:it-diagnostics-on-2535` @ `70ecfce45`, rebased on `743f3fb75`. `SYNC_STATUS_DEADLOCK`: heights frozen, tips differ, the top node rates a lower peer `Younger` and a lower node rates a higher peer `Older`, every retained round.

Executed: classifier spec 27/27; inverting the `Younger` height check fails the deadlock case. Not run: any Docker spec; the test's ratings come from your reading, not a capture.

PR #2546, opened 2026-09-09
Preserve Digest state across snapshot bootstrap and restart

a-shannon:
With `stateType="digest"` and UTXO snapshot bootstrap enabled, the snapshot producer installs a UTXO state and persists AVL-root versions. The Digest loader interprets versions as block IDs. A restart before the first full block can therefore fall back to genesis while history still reports that the snapshot is applied. The regression reproduces this through the real snapshot producer and closed-store reopening; the UTXO control passes.

Digest also drops incoming detached snapshot chunks because the receive branch incorrectly requires a local UTXO reader. An actor regression with real nonempty snapshot chunks passes in UTXO mode and fails in Digest before the correction.

This addresses [the Digest configuration concern in #2496](https://github.com/ergoplatform/ergo/pull/2496#discussion_r3971458455), preserving the [documented mode combination](https://docs.ergoplatform.com/node/modes/).

The correction:

- Installs the configured Digest state, with the snapshot header ID, root and reconstructed context persisted before publishing the applied marker.
- Removes AVL-root versions from the Digest rollback namespace and retains a durable snapshot anchor. The first full block uses the already reconstructed context instead of replaying its preceding headers.
- Converts a valid earlier UTXO-format snapshot checkpoint when reopening as Digest. Canonical history, raw context bytes, root and version must agree. Inconsistent data fails without genesis recreation.
- Stops processing after a reconstruction or preparation failure that may have changed the state store. Pre-write manifest/context failures retain their existing handling.
- Receives requested chunks independently of the local state representation. Parsing, outstanding-ID checks and serving restrictions remain in place; Digest does not gain UTXO snapshot-serving capabilities.

Review and integration:

- Parent: [#2545](https://github.com/ergoplatform/ergo/pull/2545) at `036c2c258f7aceef00ead6d92cfa238f1e4722bf`. Its published head is unchanged.
- Review this [eight-file increment](https://github.com/a-shannon/ergo/compare/036c2c258f7aceef00ead6d92cfa238f1e4722bf...63c9b79dc576461f6266c234612c35d8ed393d73). The [two-file receive follow-up](https://github.com/a-shannon/ergo/compare/43b696e4058412183178b400e5b4555fbb05bc8b...63c9b79dc576461f6266c234612c35d8ed393d73) isolates the changes since the checkpoint review. The upstream target is `i2464`, so its cumulative diff includes #2545 until that prerequisite is integrated and this branch refreshed.
- Integrate #2545 before this follow-up. [#2424](https://github.com/ergoplatform/ergo/pull/2424) at `72140e5dea0934b9f5752a70a458ae4ac863237b` and the [combined wallet review](https://github.com/a-shannon/ergo/pull/2) at `8e475916a0cade7b2dd5ed37607b6a77a47c35d2` publish the adapted checkpoint and receive corrections while preserving their finalization and wallet recovery gates.

Validation at current head `63c9b79dc576461f6266c234612c35d8ed393d73`:

- **35 receive/synchronizer tests pass:** four new network-receive cases and 31 existing synchronizer cases. The new fixture uses a real depth-14 manifest with nonempty detached chunks, actual serialized peer messages, request tracking and history persistence. UTXO and Digest reach the exact initialization request; unsolicited, malformed and duplicate chunks do not advance the plan. Digest serving restrictions remain unchanged.
- **44 checkpoint/lifecycle tests retain their unchanged validated inputs:** five producer/lifecycle cases, 19 checkpoint and negative cases, 17 existing Digest-history cases, and three existing UTXO snapshot-preservation cases.
- The lifecycle cases cover configured state type, real import, close/reopen before the first full block, application of that block, another reopen, rollback and legacy-format conversion. Separate actor failures verify real store mutation, self-shutdown, no state publication and an unchanged unapplied history marker.
- Removing checkpoint metadata, namespace cleanup or the undo anchor each fails its targeted regression. The restored source passes all 19 direct checkpoint tests again.
- Independent source and fixture review completed. **All eight checks passed on current head `63c9b79dc576461f6266c234612c35d8ed393d73`** ([run 34416424462](https://github.com/ergoplatform/ergo/actions/runs/34416424462)). The previous checkpoint head `43b696e4058412183178b400e5b4555fbb05bc8b` also passed all eight checks ([run 34409630823](https://github.com/ergoplatform/ergo/actions/runs/34409630823)); that remains historical evidence for those exact bytes.

The lifecycle fixtures use fake PoW and a small state. The separate receive fixture registers its manifest and outstanding requests directly; its holder probe observes initialization without importing state. Full wire-manifest negotiation, TCP framing, process interruption and production-network qualification remain outside this evidence. Imported AVL payloads remain on disk; inconsistent stores are preserved for recovery. Consensus validation and NiPoPoW proof handling are unchanged. [#2464](https://github.com/ergoplatform/ergo/issues/2464) remains a broader qualification issue.

### DIFF
--- a/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
+++ b/src/main/scala/org/ergoplatform/network/ErgoNodeViewSynchronizer.scala
@@ -1612,10 +1612,7 @@ class ErgoNodeViewSynchronizer(networkControllerRef: ActorRef,
         case None => log.warn(s"Asked for snapshot when UTXO set is not supported, remote: $remote")
       }
     case (_: UtxoSnapshotChunkSpec.type, serializedChunk: Array[Byte], remote) =>
-      usrOpt match {
-        case Some(_) => processUtxoSnapshotChunk(serializedChunk, hr, remote)
-        case None => log.warn(s"Asked for snapshot when UTXO set is not supported, remote: $remote")
-      }
+      processUtxoSnapshotChunk(serializedChunk, hr, remote)
     case (_: GetNipopowProofSpec.type, data: NipopowProofData, remote) =>
       sendNipopowProof(data, hr, remote)
     case (_: NipopowProofSpec.type , proofBytes: Array[Byte], remote) =>
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -14,6 +14,7 @@ import org.ergoplatform.network.ErgoNodeViewSynchronizerMessages._
 import org.ergoplatform.nodeView.ErgoNodeViewHolder.{BlockAppliedTransactions, CurrentView, DownloadRequest}
 import org.ergoplatform.nodeView.ErgoNodeViewHolder.ReceivableMessages._
 import org.ergoplatform.nodeView.history.ErgoHistory
+import org.ergoplatform.nodeView.history.storage.modifierprocessors.UtxoSetSnapshotProcessor
 import org.ergoplatform.nodeView.mempool.ErgoMemPool
 import org.ergoplatform.nodeView.mempool.ErgoMemPoolUtils.ProcessingOutcome
 import org.ergoplatform.nodeView.state._
@@ -302,9 +303,23 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
         history().createPersistentProver(store, history(), height, blockId) match {
           case Success(pp) =>
             log.info(s"Restoring state from prover with digest ${pp.digest} reconstructed for height $height")
-            history().onUtxoSnapshotApplied(height)
-            val newState = new UtxoState(pp, version = VersionTag @@@ blockId, store, settings)
-            updateNodeView(updatedState = Some(newState.asInstanceOf[State]))
+            val prepared: Try[ErgoState[_]] = settings.nodeSettings.stateType match {
+              case StateType.Digest =>
+                ErgoStateReader.reconstructStateContextBeforeEpoch(history(), height, settings).flatMap { stateContext =>
+                  DigestState.fromSnapshot(VersionTag @@@ blockId, pp.digest, stateContext, store, settings)
+                }
+              case StateType.Utxo =>
+                Try(new UtxoState(pp, version = VersionTag @@@ blockId, store, settings))
+            }
+            prepared match {
+              case Success(newState) =>
+                history().onUtxoSnapshotApplied(height)
+                updateNodeView(updatedState = Some(newState.asInstanceOf[State]))
+              case Failure(t) =>
+                abortSnapshotStatePreparation(t)
+            }
+          case Failure(t: UtxoSetSnapshotProcessor.StateWriteFailure) =>
+            abortSnapshotStatePreparation(t)
           case Failure(t) =>
             log.error("UTXO set snapshot application failed: ", t)
         }
@@ -455,7 +470,7 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
     val history = ErgoHistory.readOrGenerate(settings)
     log.info("History database read")
     val memPool = ErgoMemPool.empty(settings)
-    restoreConsistentState(ErgoState.readOrGenerate(settings).asInstanceOf[State], history) match {
+    restoreConsistentState(readStateForStartup(history), history) match {
       case Success(state) =>
         log.info(s"State database read, state synchronized")
         val wallet = ErgoWallet.readOrGenerate(
@@ -601,6 +616,39 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
         history.bestHeaderAtHeight(snapshotHeight)
       })
 
+  private def abortSnapshotStatePreparation(error: Throwable): Unit = {
+    log.error("Failed to prepare the configured snapshot state; shutting down", error)
+    // Reconstruction has modified the store behind the old state. Do not serve it.
+    context.become(Actor.emptyBehavior)
+    ErgoApp.shutdownSystem()(context.system)
+  }
+
+  private def readStateForStartup(history: ErgoHistory): State = {
+    if (settings.nodeSettings.stateType == StateType.Digest &&
+        settings.nodeSettings.utxoSettings.utxoBootstrap &&
+        history.isUtxoSnapshotApplied && history.bestFullBlockOpt.isEmpty) {
+      val height = history.minimalFullBlockHeight - 1
+      val restored = for {
+        header <- Try(history.bestHeaderAtHeight(height).getOrElse {
+          throw new IllegalStateException("Applied snapshot has no canonical header")
+        })
+        stateContext <- ErgoStateReader.reconstructStateContextBeforeEpoch(history, height, settings)
+        state <- DigestState.readSnapshot(stateDir(settings), settings, idToVersion(header.id),
+          header.stateRoot, stateContext)
+      } yield state
+      restored match {
+        case Success(state) => state.asInstanceOf[State]
+        case Failure(error) =>
+          // A failed snapshot load must not fall through to genesis recreation.
+          Try(history.closeStorage()).failed.foreach(error.addSuppressed)
+          ErgoApp.shutdownSystem()(context.system)
+          throw error
+      }
+    } else {
+      ErgoState.readOrGenerate(settings).asInstanceOf[State]
+    }
+  }
+
   private def restoreConsistentState(stateIn: State, history: ErgoHistory): Try[State] = {
     (stateIn.version, history.bestFullBlockOpt, stateIn) match {
       case (ErgoState.genesisStateVersion, None, _) =>
@@ -611,6 +659,9 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
         Success(stateIn)
       case (_, None, _) if isPreparedUtxoSnapshotState(stateIn, history) =>
         log.info(s"Prepared UTXO snapshot state ${encoder.encode(stateIn.version)} restored before the first full block")
+        Success(stateIn)
+      case (_, None, _: DigestState) if isPreparedDigestSnapshotState(stateIn, history) =>
+        log.info(s"Prepared Digest snapshot state ${encoder.encode(stateIn.version)} restored before the first full block")
         Success(stateIn)
       case (_, None, _) =>
         log.info("State and history are inconsistent. History is empty on startup, rollback state to genesis.")
@@ -637,6 +688,19 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
           acc.flatMap(_.applyModifier(m, chainTipOpt)(lm => self ! lm))
         }
     }
+  }
+
+  private def isPreparedDigestSnapshotState(state: State, history: ErgoHistory): Boolean = {
+    val snapshotHeight = history.minimalFullBlockHeight - 1
+    ErgoNodeViewHolder.isPreparedDigestSnapshotState(
+      settings.nodeSettings.stateType == StateType.Digest && state.isInstanceOf[DigestState],
+      settings.nodeSettings.utxoSettings.utxoBootstrap,
+      history.isUtxoSnapshotApplied,
+      state.version,
+      state.rootDigest,
+      history.bestHeaderAtHeight(snapshotHeight),
+      state.store.get(ErgoStateReader.ContextKey),
+      ErgoStateReader.reconstructStateContextBeforeEpoch(history, snapshotHeight, settings).toOption.map(_.bytes))
   }
 
   /**
@@ -743,6 +807,21 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
 
 
 object ErgoNodeViewHolder {
+
+  private[nodeView] def isPreparedDigestSnapshotState(
+      stateIsDigest: Boolean,
+      utxoBootstrap: => Boolean,
+      snapshotApplied: => Boolean,
+      stateVersion: VersionTag,
+      stateRoot: Array[Byte],
+      snapshotHeaderOpt: => Option[Header],
+      storedContext: => Option[Array[Byte]],
+      expectedContext: => Option[Array[Byte]]): Boolean =
+    stateIsDigest &&
+      utxoBootstrap &&
+      snapshotApplied &&
+      snapshotHeaderOpt.exists(matchesPreparedUtxoSnapshotHeader(stateVersion, stateRoot, _)) &&
+      storedContext.exists(stored => expectedContext.exists(_.sameElements(stored)))
 
   private[nodeView] def isPreparedUtxoSnapshotState(
       stateIsUtxo: Boolean,
--- a/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/FullBlockProcessor.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/FullBlockProcessor.scala
@@ -68,7 +68,12 @@ trait FullBlockProcessor extends HeadersProcessor {
     case ToProcess(fullBlock, newModRow, Some(newBestBlockHeader), newBestChain)
       if isValidFirstFullBlock(fullBlock.header) =>
 
-      val headers = headerChainBack(10, fullBlock.header, h => h.height == 1)
+      val headersToApply = if (nodeSettings.utxoSettings.utxoBootstrap && isUtxoSnapshotApplied) {
+        // Snapshot reconstruction has already installed the preceding headers in the state context.
+        Seq.empty
+      } else {
+        headerChainBack(10, fullBlock.header, h => h.height == 1).headers.dropRight(1)
+      }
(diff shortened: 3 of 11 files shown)


cafebedouin:
Flagging a gap in the same function this PR touches (`createPersistentProver`), surfaced by the AI-assisted node audit — it composes with your `StateWriteFailure` change rather than colliding with it.

**The gap.** `downloadedChunksIterator()` reads each chunk as

```scala
historyStorage.get(chunkId).flatMap(bs => SubtreeSerializer.parseBytesTry(bs).toOption)
```

so a subtree that is missing on read-back (`get` → `None`) or fails to deserialize (`parseBytesTry` → `Failure` → `toOption` → `None`) is silently dropped by the `flatMap`. `createPersistentProver` then hands that short iterator straight to `VersionedLDBAVLStorage.recreate(manifest, downloadedChunksIterator(), …)` and records the manifest digest — over a tree missing subtrees the manifest declares. Nothing checks that the number of restored subtrees equals `manifest.subtreesIds.size`, so an incomplete restore currently returns `Success`.

**Witness.** A local spec that registers one unparseable chunk (`Array[Byte](1,2,3)`) among otherwise-valid subtrees still gets a successful prover back from `createPersistentProver`; the all-chunks control passes too. The failure is silent, not merely logged.

**Suggested addition** — a completeness pre-check at the top of the `Some(manifest)` branch, before the transfer, orthogonal to the write-phase recovery you're adding:

```scala
val restored = downloadedChunksIterator().size
if (restored != manifest.subtreesIds.size) {
  val m = s"Incomplete UTXO snapshot restore: $restored of ${manifest.subtreesIds.size} subtrees; failing closed"
  log.error(m)
  Failure(new IllegalStateException(m))
} else {
  // existing reconstructStateContextBeforeEpoch → recreate → StateWriteFailure recovery
}
```

Two caveats:
- This is the *minimal* fail-closed version. It counts subtrees; it doesn't re-derive the tree digest from the reconstructed nodes and compare to the manifest, which would also catch a chunk that parses to the *wrong* content. If you'd rather go straight to the stronger check, that supersedes this.
- The count consumes the iterator once for `.size` and `recreate` re-iterates — a second DB scan. A production version should thread a single pass (count while iterating, or fold the check into `recreate`).

Happy to open this as a small commit against your branch or a separate PR — whichever fits your sequencing (#2545 → #2546 → into #2496's `i2464`). Flag either way and I'll match it.

a-shannon:
@cafebedouin, thanks for flagging this. I checked the read-back and restore paths at `63c9b79dc576461f6266c234612c35d8ed393d73`: missing/unparseable chunks are skipped, and `recreate` consumes the shortened iterator without checking it against the manifest. `restorePrunedProver` only restores the pruned root, so its success does not establish that all detached subtrees are available. I have not independently run your witness.

Normal network intake already validates requested chunk identities. This gap concerns the later storage read-back; the supplied witness should not be interpreted as malformed peer bytes bypassing that intake check.

Please post a small commit and the failing spec here against the current [#2546](https://github.com/ergoplatform/ergo/pull/2546) head, rather than opening a parallel PR. That keeps this within the existing #2545 -> #2546 review sequence.

The invariant should cover the chunks actually consumed during restoration. A separate count scan followed by another permissive scan is insufficient if read-back changes; equal counts also do not establish the expected chunk identities. Prefer a bounded streaming check against the manifest, with missing or unparseable expected chunks producing failure. Preserve `StateWriteFailure` for errors after the store-write phase starts, so callers cannot continue serving the previous state after possible mutation.

Please include your complete-snapshot control and separate missing-chunk and unparseable-chunk cases, with the exact tested commit. A parsed wrong-ID substitution is the useful negative for any identity check. A digest taken from the restored pruned root alone should not be described as a full materialization check. We can review the focused increment before integration without changing the unrelated Digest checkpoint behavior.

cafebedouin:
@a-shannon — done as asked: two commits on top of the current #2546 head `63c9b79dc576461f6266c234612c35d8ed393d73`, no parallel PR.

Branch: `cafebedouin/ergo:e047-restore-completeness` — [two-file increment](https://github.com/cafebedouin/ergo/compare/63c9b79dc576461f6266c234612c35d8ed393d73...d651d5e3688611c3979f2b9884079119861c520c)

1. `2ef26c8ec40c9fa2ff83f935b30a0bbdd8757998` — **specs only** (`UtxoSetSnapshotProcessorSpecification`). At this commit the three negatives fail: `createPersistentProver` returns `Success`.
2. `d651d5e3688611c3979f2b9884079119861c520c` — **the fix** (`UtxoSetSnapshotProcessor`, +24/−1). All five properties pass.

**The invariant, per your notes.** The count pre-check I first suggested is dropped. `createPersistentProver` now feeds `recreate` from a strict `manifestChunksIterator(manifest)` instead of the permissive `downloadedChunksIterator()`:

- it walks `manifest.subtreesIds` in order — exactly one `historyStorage.get` per declared subtree id, lazily, so it is bounded by the manifest and is the *same* pass `recreate` consumes (no separate scan that could disagree with a later one);
- each chunk must be present, parse, and `verify(expectedId)` at its index; the first one that is missing, unparseable, or reads back under a different id throws `IllegalStateException` naming the index and expected id.

`downloadedChunksIterator()` is left unchanged to keep the increment small, but after this change it has no production caller (only the existing spec uses it) — removing it, or making it strict too, is your call.

**Failure typing.** The iterator is consumed inside `recreate` → `store.update`, i.e. after the store-write phase has been entered, so the failure surfaces through your existing `recoverWith` as `StateWriteFailure` with the `IllegalStateException` as cause. I kept that deliberately conservative rather than classifying it as a pre-write failure: `LDBVersionedStore.update` throws before `db.write(batch)`, and the specs assert `store.lastVersionID` is still `None`, but in-memory LSN state has been touched by then. If you would rather have a pre-write classification, that is a change in `update`/`recreate`, not here.

**Specs** (each case builds its own processor with its own chunk storage, so nothing stale from another property can mask a result):

| case | before fix (`2ef26c8ec`) | after fix (`d651d5e36`) |
|---|---|---|
| existing register/plan/restore property | pass | pass |
| complete snapshot is restored (control) | pass | pass |
| chunk missing on read-back | **FAIL** (restore = Success) | pass |
| unparseable stored chunk (`Array[Byte](1,2,3)`) | **FAIL** (restore = Success) | pass |
| valid chunk stored under another chunk's id | **FAIL** (restore = Success) | pass |

The wrong-id case keeps the chunk count equal to the manifest and every stored chunk parseable — only identity differs — so it is the negative a count check would have passed.

Regression on `d651d5e36`, JDK 8: `DigestSnapshotBootstrapSpecification`, `DigestSnapshotNetworkSpecification`, `DigestSnapshotStateSpecification`, `ErgoNodeViewSynchronizerSpecification` — 55 succeeded, 0 failed. I have not run the full suite or CI on these commits.

**Not claimed.** Agreed on your intake point: this is a storage read-back gap, not malformed peer bytes bypassing the network-side identity check; the specs write directly through `registerDownloadedChunk`. The identity check is per-subtree `verify(expectedId)` against the manifest's declared ids; it is not a re-derivation of the full tree digest from the materialized store, and `restorePrunedProver` succeeding is not described as one. Digest checkpoint behaviour is untouched.

Cherry-pick, squash or rewrite as suits your sequence.

*AI-assisted (Claude Fable 5.1, via Claude Code); tests run locally by me before posting.*

cafebedouin:
@a-shannon — separate from the restore increment above: I read the #2546 increment (`036c2c25…63c9b79d`, production diff first) looking for things worth settling before merge. One executed finding, two smaller requests. Everything else I checked held up — manifest/snapshot-info receive is reachable in Digest mode, store-write precedes the history marker so a crash between them falls back to a clean re-bootstrap, the `clean(0)` + empty `update` leaves a usable rollback anchor, and dropping the first-block header replay is neutral for UTXO (`UtxoState` ignores header modifiers).

### 1. `isUtxoSnapshotApplied` is also true for a pruned node; the new startup path fails closed on it (executed)

`isUtxoSnapshotApplied` is `readMinimalFullBlockHeight() > GenesisHeight`. `updateBestFullBlock` writes the same value as the pruning floor when `blocksToKeep >= 0`. `readStateForStartup` now treats `Digest && utxoBootstrap && isUtxoSnapshotApplied && bestFullBlockOpt.isEmpty` as "a snapshot checkpoint must be on disk", and shuts the node down if `readSnapshot` cannot verify one.

So an existing pruned Digest store that has a floor but no full block yet, restarted with `utxoBootstrap = true`, no longer starts:

| same store, restart with `utxoBootstrap = true` | result |
|---|---|
| #2545 parent `036c2c258f7aceef00ead6d92cfa238f1e4722bf` | starts; view returned, floor 20 |
| #2546 head `63c9b79dc576461f6266c234612c35d8ed393d73` | actor system terminates: `requirement failed: Stored snapshot root does not match canonical history` (`DigestState.readSnapshot`) |

Control in both runs: restarting with the original `utxoBootstrap = false` config starts normally. Fixture: real `ErgoNodeViewRef` over reopened stores (your `Session`/`closeOwnedStores` pattern), Digest, `blocksToKeep = 10`, `votingLength = 20`, 45 fake-PoW headers, no full blocks. Caveat: in this fixture header application did not itself advance the floor, so the witness calls `history.updateBestFullBlock(lastHeader)` directly — the function `toDownload` calls when the header chain first syncs — to write it. Spec: [`PrunedDigestFlipWitness`](https://github.com/cafebedouin/ergo/blob/pr2546-pruned-digest-flip-witness/src/test/scala/org/ergoplatform/nodeView/viewholder/PrunedDigestFlipWitness.scala) (one file on top of `63c9b79d`; it reports both outcomes via `info` rather than asserting the flip, so it runs unchanged at either commit).

It is a narrow, operator-induced window (config change between header sync and the first full block) and reverting the config recovers, so I would not call it severe. But the fail-closed branch is keyed on a predicate that does not mean what the branch assumes, and `FullBlockProcessor`'s new `utxoBootstrap && isUtxoSnapshotApplied` condition shares it. Suggested change, smallest first:

- in `readStateForStartup`, take the fail-closed path only when the state store actually carries snapshot evidence (e.g. the snapshot header-id version key that `fromSnapshot` writes), and otherwise fall through to `ErgoState.readOrGenerate` as before; or
- persist an explicit snapshot-applied marker in `onUtxoSnapshotApplied` and key both new consumers on that rather than on the floor.

Either way, the flip case above makes a cheap regression.

### 2. The post-write shutdown now applies to UTXO mode but is only tested in Digest

`case Failure(t: StateWriteFailure) => abortSnapshotStatePreparation(t)` is mode-independent, so a UTXO-mode node now also shuts down where it previously logged and continued. I think that is the right behaviour. The lifecycle matrix only runs the `postWriteFailure` case for `StateType.Digest`, though; a UTXO variant of that case (the fault-injecting holder is currently built around a `DigestState`, so it is a small fixture addition rather than one more row) would cover the mode most nodes run, and the description could say the change is not Digest-only.

### 3. Imported AVL payload in a Digest store — question, not a blocker

You already note that the imported AVL payload stays on disk. For a Digest node that is the full UTXO set, retained indefinitely and never read after `fromSnapshot`. Is removing it after the anchor is written in scope for a follow-up? If so, an issue reference in the description would help whoever sizes Digest-mode disk later. Related, and within your stated "production qualification is outside this evidence": `fromSnapshot`'s `store.clean(0)` deletes every undo LSN written by `recreate` in a single `WriteBatch`, which at mainnet size is one delete per tree node — worth one observed run before this path is relied on.

*AI-assisted (Claude Fable 5.1, via Claude Code); witness run locally on JDK 8 at both commits named above.*

cafebedouin:
Review carried out by Claude (Anthropic, Claude Opus 5.5) for cafebedouin, using forkbench; executed at `02ae9072` the added specs, a variant of the pruned-Digest restart case and the recommended fix (JDK 8), and a two-node restart run (JDK 21).

Two commits in this PR are the reviewer's own: `2ef26c8e` and `d651d5e3`, the snapshot-completeness regressions.

**Follow-up to item 1 of the 2026-09-20 comment: the `allowGenesis` fix covers a pruning floor equal to `votingLength` (executed) and above it (read); below it, the same restart still shuts the node down (executed).**

Executed: the case "ordinary pruned Digest state survives enabling snapshot bootstrap before the first full block" with `blocksToKeep = 30` instead of 10 (`votingLength = 20`, 45 headers, no full block). Floor 16: restart with the original config returns the genesis view; restart with `utxoBootstrap = true` terminates the actor system. Floor 20 (`blocksToKeep = 10`): both restarts return the genesis view. `DigestSnapshotNetworkSpecification` passes, and its three Digest cases fail with only the `ErgoNodeViewSynchronizer.scala` hunk reverted. The other three added classes pass; without the production change the test tree does not compile, so whether they guard it was not shown. With snapshot bootstrap off, multi-node runs of the merge base and this branch showed no difference (one run each). Not run: a Digest node bootstrapping from a served snapshot on a network; the full suite; CI.

**[Likely bug] Recognise an ordinary genesis store before reconstructing the snapshot state context**

Observed: `readStateForStartup` calls `reconstructStateContextBeforeEpoch(history, floor - 1)` (`ErgoNodeViewHolder.scala:642`) before `readSnapshot(..., allowGenesis = true)` (`:643-644`) can recognise the genesis store (`DigestState.scala:170-175`). That call fails unless `floor - 1` ends a voting epoch (`ErgoStateReader.scala:89-90`); the pruning floor is aligned only above `votingLength` (`FullBlockPruningProcessor.scala:66-69`). So a pruned Digest store with `1 < floor < votingLength` and no full block fails at startup (`ErgoNodeViewHolder.scala:646-652`): `Wrong height provided in reconstructStateContextBeforeEpoch, height 15, epoch length 20`. The effect is a refused start; the failure precedes the state store being opened, so reverting `utxoBootstrap` should recover (read, not run). It needs a chain shorter than `votingLength + blocksToKeep` (`votingLength` is 1024 on mainnet, 128 on testnet). Wrong if a real node never writes such a floor; `ToDownloadProcessor.scala:131,145` writes it on header sync while `utxoBootstrap` is off.

Recommended: split the ordinary-genesis probe out of `readSnapshot` (it needs only the store, not `stateContext`) and run it before the reconstruction; the `blocksToKeep = 30` case makes the regression. (executed)

Executed with the recommended change at `02ae9072`: the floor-16 case opens as genesis, and 43 of 43 pass (the witness and the four snapshot specs). On two nodes the PR head refused to start in 2 of 2 runs with `utxoBootstrap = true` after a headers-only sync; a timed link cut held that state, so this shows the refused start, not how often a normal restart lands there.

- Witness spec: https://gist.github.com/cafebedouin/37f62b346d788e117057b2d0283d15d6 (sha256 `d3a1d42fbc8414dc828e2f55e1cb4842643f0caefad7715dc2013d280108c8e4`)
- Change: https://gist.github.com/cafebedouin/2042d7c671c615be5b3fc98567c2fb51 (sha256 `8d71991d24a59b38dcdf481e275fdb12181a831a4244ff0a4b0d5b595887da1e`)

<details><summary>Run output (abridged)</summary>

```
JDK 8, at 02ae9072; the PR's case with blocksToKeep varied, stopped after the flipped reopen
  blocksToKeep=10 votingLength=20 floor=20 bestFullBlock=None
    control reopen (utxoBootstrap=false): state version genesis = true
    flipped reopen (utxoBootstrap=true): view=Success(genesis) actorSystemTerminated=false
  blocksToKeep=30 votingLength=20 floor=16 bestFullBlock=None
    control reopen (utxoBootstrap=false): state version genesis = true
    flipped reopen (utxoBootstrap=true): view=Failure(java.util.concurrent.TimeoutException: Futures timed out after [15 seconds]) actorSystemTerminated=true
  Caused by: java.lang.Exception: Wrong height provided in reconstructStateContextBeforeEpoch, height 15, epoch length 20
    at org.ergoplatform.nodeView.state.ErgoStateReader$.reconstructStateContextBeforeEpoch(ErgoStateReader.scala:90)
    at org.ergoplatform.nodeView.ErgoNodeViewHolder.$anonfun$readStateForStartup$3(ErgoNodeViewHolder.scala:642)
    at scala.util.Success.flatMap(Try.scala:251)
    at org.ergoplatform.nodeView.ErgoNodeViewHolder.readStateForStartup(ErgoNodeViewHolder.scala:639)

DigestSnapshotNetworkSpecification   as is: 4 passed   synchronizer hunk reverted: 1 passed, 3 failed (Digest cases time out)

multi-node, JDK 21, one run per jar, snapshot bootstrap at its default (off)   merge base ff4d9a18   this branch
  follower on a lighter fork meets a heavier one: nodes unresponsive after        0                    0
  Digest follower and pruning follower: caught up / same state root / pruned      yes / yes / yes      yes / yes / yes
  released 6.0.6 with this branch, each mining in turn: same state root at equal height, both directions: yes
  6.0.6 with itself, same fork setup: 0 unresponsive in 16 of 16 runs
```
This branch was built as the merge base plus this PR's diff (appVersion `6.0.5-35-31341b03-SNAPSHOT`, jar sha256 `6482b359…`).

```
JDK 8, at 02ae9072 with the recommended change
  witness, floor 20 and floor 16: flipped reopen (utxoBootstrap=true): view=Success(genesis) actorSystemTerminated=false
  DigestSnapshotStateSpecification 25, DigestSnapshotBootstrapSpecification 7, DigestSnapshotNetworkSpecification 4,
  UtxoSetSnapshotProcessorSpecification 5, witness 2:  Tests: succeeded 43, failed 0

two nodes, this branch's jar on both (JDK 21), 2 runs; follower synced headers to 59 with no full block, link cut,
restart with utxoBootstrap = false comes up, restart with utxoBootstrap = true:
  ERROR a.actor.OneForOneStrategy - Wrong height provided in reconstructStateContextBeforeEpoch, height 17, epoch length 128
  REST came up, process not alive 15 s later; not recovered after the link was restored (2 of 2 runs)
```
</details>

Issue #2534, opened 2026-09-08
Expose transaction validation cost in mempool transaction JSON


K-Singh:
# Expose transaction validation cost in mempool transaction JSON

## Summary

`GET /transactions/unconfirmed` and the other mempool transaction endpoints return a
transaction object with `id`, `inputs`, `dataInputs`, `outputs`, `size`. The validation
cost is not included, even though the node already holds it on
`UnconfirmedTransaction.lastCost`.

Add a `cost` field to the mempool transaction JSON.

## Motivation

Block assembly is bounded by two limits: `maxBlockSize` and `maxBlockCost`
(`Parameters.maxBlockCost`). The node's own candidate generator enforces both in
`CandidateGenerator.correctLimits`:

```scala
blockTxs.map(_._2).sum < maxBlockCost && blockTxs.map(_._1.size).sum < maxBlockSize
```

`size` is in the API response, so external tooling can enforce the size limit. `cost` is
not, so the cost limit cannot be enforced externally.

Anyone assembling blocks outside the node (external miners, block template builders,
mempool simulators, fee/throughput analysis) has to either:

- re-run script validation against a UTXO set to recompute per-tx cost, or
- guess with a heuristic derived from `size` and input/output counts.

Both are wrong or expensive. The node computed the exact number during mempool
admission and then discards it at the serialization boundary.

Cost is also already load-bearing in the mempool itself: `SortingOption.FeePerCycle`
orders the pool by fee-per-cost using `lastCost` (`ErgoMemPool.feeFactor`). A node
running that sorting mode exposes an ordering the API client cannot reproduce or
explain.

## Where the value already exists

`UnconfirmedTransaction` carries it:

```scala
class UnconfirmedTransaction(val transaction: ErgoTransaction,
                             val lastCost: Option[Int],
                             ...)
```

It is populated in two places:

- `ErgoMemPool.process` — `acceptIfNoDoubleSpend(unconfirmedTx.withCost(cost), ...)`
  after `utxoWithPool.validateWithCost(...)`
- `ErgoBaseApiRoute.verifyTransaction` — for transactions submitted over the API
- refreshed by `CleanupWorker` on revalidation (`head.withCost(txCost)`)

It is `None` only in digest-state mode, where no script validation happens. The field
must therefore be nullable/optional in the response.

## Current response

```json
{
  "id": "...",
  "inputs": [ ... ],
  "dataInputs": [],
  "outputs": [ ... ],
  "size": 344
}
```

[...]

PR #2547, opened 2026-09-11
Expose transaction validation cost in mempool transaction JSON

K-Singh:
## Problem

Block assembly is bounded by two limits, `maxBlockSize` and `maxBlockCost`. Both are enforced by `CandidateGenerator.correctLimits`:

```scala
blockTxs.map(_._2).sum < maxBlockCost && blockTxs.map(_._1.size).sum < maxBlockSize
```

The mempool API returns `size`, so external tooling can enforce the size limit. It does not return `cost`, so the cost limit cannot be enforced outside the node.

Anyone assembling blocks externally (block template builders, mempool simulators, throughput analysis) has to re-run script validation against a UTXO set, or guess from `size` and input/output counts. The node already computed the exact value during mempool admission and discarded it at the serialization boundary.

Cost is also already load-bearing inside the pool: `SortingOption.FeePerCycle` orders by fee-per-cost via `lastCost` in `ErgoMemPool.feeFactor`. A node in that mode exposes an ordering clients cannot reproduce.

## Change

Add a `cost` field to the mempool transaction JSON, sourced from `UnconfirmedTransaction.lastCost`.

```json
{
  "id": "...",
  "inputs": [ ... ],
  "dataInputs": [],
  "outputs": [ ... ],
  "size": 344,
  "cost": 21456
}
```

Affected endpoints:

- `GET /transactions/unconfirmed`
- `GET /transactions/unconfirmed/byTransactionId/{txId}`
- `POST /transactions/unconfirmed/byErgoTree`

## Null semantics

`cost` is `null` when the node never measured it:

1. **Digest state.** No script validation happens, so `lastCost` is never set.
2. **Rollback.** `ErgoNodeViewHolder.updateMemPool` returns transactions from disconnected blocks with `UnconfirmedTransaction(tx, None)` and puts them via `put`, which skips `validateWithCost`. They stay `null` until `CleanupWorker` re-validates them.

The key is always present on a node carrying this change, so clients can distinguish "node too old" (key absent) from "cost unknown" (key `null`).

## Implementation notes

**`unconfirmedById` added to `ErgoMemPoolReader`.** The by-id endpoint used `modifierById`, which returns `Option[ErgoTransaction]` and drops the wrapper before the route sees it. The new method returns the `UnconfirmedTransaction`; `modifierById` is now defined in terms of it, so there is one lookup path. `ErgoMemPool` implements it as `pool.get(id)`, identical to what `modifierById` did.

**`byErgoTree` keeps the wrapper.** Its two `collect` blocks mapped to `tx.transaction`, discarding the cost. They become `filter`s. Deduplication across the output-match and input-match sets is unchanged, since `UnconfirmedTransaction.equals` and `hashCode` are defined on the transaction id.

**Digest-state branch of `byErgoTree` now uses the shared JSON builder.** It previously serialized a bare `ErgoTransaction`, which did not match the `ErgoTransactionWithInputBoxes` schema the spec declares for that response. Routing it through the same builder fixes that and means `cost` is never silently absent from a response that otherwise carries it.

## Compatibility

Additive and optional. Not in the schema's `required` list. Clients that ignore unknown keys are unaffected. Confirmed-transaction schemas (`ErgoTransaction`, `IndexedErgoTransaction`, `WalletTransaction`) are untouched, since `lastCost` only exists for unconfirmed transactions.

## Tests

Six cases added to `TransactionApiRouteSpec`, covering cost reported by id, in the list, via `byErgoTree`, agreement between the list and by-id endpoints, the `null` case, and that `size` still reports correctly alongside `cost`.

Run locally on `v6.0.6`:

- `TransactionApiRouteSpec`: 29 passed, 0 failed (23 pre-existing, 6 new)
- Mempool, mempool auditor, wallet service and all HTTP route suites: 191 passed across 19 suites, 0 failed

## Known follow-up, not in this PR

Rollback-restored transactions report `null` until the next cleanup pass. Closing that would mean carrying per-transaction cost from block validation through `updateMemPool`, which is a larger change than exposing a value the pool already holds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/resources/api/openapi.yaml
+++ b/src/main/resources/api/openapi.yaml
@@ -472,6 +472,16 @@ components:
                 description: Size in bytes
                 type: integer
                 format: int32
+            cost:
+                description: >-
+                  Validation cost of the transaction, as measured when it entered the mempool.
+                  Null when the cost is not known to the node, i.e. when the node is running a
+                  digest state, or when the transaction was returned to the pool by a rollback
+                  and has not been re-validated yet.
+                type: integer
+                format: int32
+                nullable: true
+                example: 21456
 
     IndexedErgoTransaction:
       type: object
--- a/src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
+++ b/src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
@@ -101,8 +101,14 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
 
   /**
     * Creates a transaction JSON representation with resolved input boxes.
+    *
+    * `cost` is the validation cost measured when the transaction entered the pool. It is None
+    * when no script validation was done for it, i.e. when the node runs a digest state, or when
+    * the transaction was returned to the pool by a rollback and has not been re-checked yet.
   */
-  private def createTransactionWithResolvedInputs(tx: ErgoTransaction, resolvedInputs: Map[BoxId, ErgoBox]): Json = {
+  private def createTransactionWithResolvedInputs(tx: ErgoTransaction,
+                                                  resolvedInputs: Map[BoxId, ErgoBox],
+                                                  cost: Option[Int]): Json = {
 
     val enrichedInputs = tx.inputs.map { input =>
       val baseInput = Json.obj(
@@ -123,7 +129,8 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
       "inputs"     -> enrichedInputs.asJson,
       "dataInputs" -> tx.dataInputs.asJson,
       "outputs"    -> tx.outputs.asJson,
-      "size"       -> tx.size.asJson
+      "size"       -> tx.size.asJson,
+      "cost"       -> cost.asJson
     )
   }
 
@@ -137,7 +144,7 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
         val enrichedTxs = transactions.map { unconfirmedTx =>
           val tx             = unconfirmedTx.transaction
           val resolvedInputs = resolveTransactionInputs(tx.inputs, state, pool)
-          createTransactionWithResolvedInputs(tx, resolvedInputs)
+          createTransactionWithResolvedInputs(tx, resolvedInputs, unconfirmedTx.lastCost)
         }
         enrichedTxs.asJson
     }
@@ -145,11 +152,12 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
   /**
     * Resolves inputs for a single transaction and returns it with resolved inputs.
   */
-  private def getUnconfirmedTransactionWithResolvedInputs(transaction: ErgoTransaction): Future[Json] =
+  private def getUnconfirmedTransactionWithResolvedInputs(unconfirmedTx: UnconfirmedTransaction): Future[Json] =
     getStateAndPool.map {
       case (state, pool) =>
+        val transaction    = unconfirmedTx.transaction
         val resolvedInputs = resolveTransactionInputs(transaction.inputs, state, pool)
-      createTransactionWithResolvedInputs(transaction, resolvedInputs)
+        createTransactionWithResolvedInputs(transaction, resolvedInputs, unconfirmedTx.lastCost)
     }
 
   private def getUnconfirmedTransactions(offset: Int, limit: Int): Future[Json] = getUnconfirmedTransactionsWithResolvedInputs(offset, limit)
@@ -247,7 +255,7 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
     (pathPrefix("unconfirmed" / "byTransactionId") & get & modifierId) { modifierId =>
       ApiResponse(
         getMemPool.flatMap { pool =>
-          pool.modifierById(modifierId) match {
+          pool.unconfirmedById(modifierId) match {
             case Some(unconfirmedTx) =>
               getUnconfirmedTransactionWithResolvedInputs(unconfirmedTx)
             case None =>
@@ -292,25 +300,25 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
               val allTxs = pool.getAll
               val txsWithOutputMatch =
                 allTxs
-                  .collect { case tx if tx.transaction.outputs.exists(_.ergoTree.bytesHex == ergoTree) =>
-                    tx.transaction
-                  }.toSet
+                  .filter(_.transaction.outputs.exists(_.ergoTree.bytesHex == ergoTree))
+                  .toSet
 
               getState.flatMap {
                 case state: UtxoStateReader =>
                   val txWithInputMatch =
                     allTxs
-                      .collect { case tx if
-                          tx.transaction.inputs.exists(i => state.boxById(i.boxId).exists(_.ergoTree.bytesHex == ergoTree)) =>
-                        tx.transaction
-                      }
+                      .filter(
+                        _.transaction.inputs.exists(i => state.boxById(i.boxId).exists(_.ergoTree.bytesHex == ergoTree))
+                      )
                       val allMatchingTxs = (txsWithOutputMatch ++ txWithInputMatch).toSeq.slice(offset, offset + limit)
                       Future.sequence(allMatchingTxs.map(getUnconfirmedTransactionWithResolvedInputs)).map(_.asJson)
                 case _ =>
+                      // digest state: inputs cannot be resolved, but the response keeps the shape
+                      // declared by the ErgoTransactionWithInputBoxes schema, `cost` included
                       Future.successful(
                         txsWithOutputMatch
                           .slice(offset, offset + limit)
-                          .map(_.asJson)
+                          .map(utx => createTransactionWithResolvedInputs(utx.transaction, Map.empty, utx.lastCost))
                           .asJson
                       )
               }
--- a/src/test/scala/org/ergoplatform/http/routes/TransactionApiRouteSpec.scala
+++ b/src/test/scala/org/ergoplatform/http/routes/TransactionApiRouteSpec.scala
@@ -11,6 +11,7 @@ import org.ergoplatform.ErgoBox.{AdditionalRegisters, NonMandatoryRegisterId, To
 import org.ergoplatform.http.api.{ApiCodecs, TransactionsApiRoute}
 import org.ergoplatform.modifiers.mempool.{ErgoTransaction, UnconfirmedTransaction}
 import org.ergoplatform.nodeView.ErgoReadersHolder.{GetDataFromHistory, GetReaders, Readers}
+import org.ergoplatform.nodeView.mempool.ErgoMemPool
 import org.ergoplatform.settings.RESTApiSettings
 import org.ergoplatform.utils.Stubs
 import org.ergoplatform.{DataInput, ErgoBox, ErgoBoxCandidate, Input}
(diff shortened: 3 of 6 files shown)


src/main/resources/api/openapi.yaml
```diff
@@ -472,6 +472,16 @@ components:
                 description: Size in bytes
                 type: integer
                 format: int32
+            cost:
+                description: >-
+                  Validation cost of the transaction, as measured when it entered the mempool.
```
jozanek, line 477: This says cost is measured when the transaction entered the mempool, but `CleanupWorker.scala:90-92` overwrites it via `head.withCost(txCost)` on every re-validation pass for transactions older than `mempoolCleanupDuration` (30s default) — which is what `UnconfirmedTransaction.scala:12` means by "during last check".

An integrator reading this will treat the number as pinned to admission and may cache it. Would you be open to something like "as of this node's most recent validation of this transaction, against the state at that time"? Since the re-check pass is itself budget-limited (`CostLimit` at CleanupWorker.scala:27), two honest nodes can also report different costs for the same transaction, which may be worth a clause.

src/main/resources/api/openapi.yaml
```diff
@@ -472,6 +472,16 @@ components:
                 description: Size in bytes
                 type: integer
                 format: int32
+            cost:
```
jozanek, line 475: The tri-state this sets up — key absent means an older node, `null` means unmeasured, a number means measured — works because `cost` stays out of the `required` list above. That reads deliberate to me and I'd keep it exactly as is.

What I think is missing is anything in the file saying so. A later tidy-up that "completes" the required list would quietly remove the version detection the description relies on, and no test would catch it. Would you be open to a sentence in the description: always present on nodes carrying this field, absence indicates an older node?

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -101,8 +101,14 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
 
   /**
     * Creates a transaction JSON representation with resolved input boxes.
+    *
+    * `cost` is the validation cost measured when the transaction entered the pool. It is None
```
jozanek, line 105: Same wording as the spec description — "measured when the transaction entered the pool". Given `CleanupWorker` refreshes it, the next reader of this method may take the value as fixed for the lifetime of the pool entry.

Would you mind matching whatever wording you settle on in `openapi.yaml`, so the two don't drift apart later?

jozanek (commented):
Thanks for the description — the implementation notes made this quick to follow, and the `equals`/`hashCode` reasoning behind the `collect` → `filter` change holds up.

I read this as: publish a number the pool already holds, so tooling can reason about the mempool without re-running validation. On that reading the field belongs, the nullable-field shape looks right to me, and the blast radius is small — explorer-backend (circe) and ergo-appkit (Gson) both ignore unknown keys.

Where I got stuck is the stated why. Both justifications came apart against the code they cite:

- **Block-budget enforcement.** `collectTxs` never reads `lastCost` — it re-runs `validateWithCost` against its own `upcomingContext` (CandidateGenerator.scala:979-987), and budgets `maxBlockCost - safeGap`, with safeGap up to 500,000 (:680-696). `correctLimits` also counts the fee transaction the node synthesizes itself. A builder summing these against `maxBlockCost` overshoots, and `collectTxs` truncates silently rather than rejecting (:1000-1003).
- **Reproducing FeePerCycle ordering.** `mempoolSorting = "random"` is the shipped default (application.conf:88) — a per-process coin flip that no endpoint exposes. And `OrderedTxPool.put` discards the new `feeFactor` when the id is already present (:101-110), so a cleanup pass moves the reported cost but not the weight.

Neither sinks the field. They point at a different framing — mempool observability: fee-market analysis, simulation, "why isn't my transaction being mined" — which it serves well. Would you be open to re-justifying it that way?

Questions:

1. Is there a consumer that would **sum** these against a budget? If yes, I think the shape needs to change and `/mining/candidateWithTxs` is the better vehicle. If every consumer is descriptive, the field is right and the rest is wording.
2. #2534 describes this change but isn't linked — was that deliberate? It also records `getAll(Seq(id))` as an alternative to touching the trait. I lean toward `unconfirmedById` being the better call, so I'd mainly like the reasoning captured where the next reader finds it.
3. `POST /transactions/check` already computes this exact number and discards it (ErgoBaseApiRoute.scala:136-141). Worth a line on why the pool-side view was the one you needed?
4. Base `v6.0.6` matches convention, but the carry-forward PR #2473 is currently CONFLICTING — is there a plan for how this reaches 6.0.7?

I couldn't find any third-party request for this: #1854, the wishlist that produced these endpoints, doesn't mention cost. Not evidence against the change, but the constituency isn't visible to a reviewer.

What would move me to approve: corrected wording at the two description sites, a line preserving the required-list tri-state, and the null-cost test pointed at `digestReadersRef`.

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
```diff
@@ -292,25 +300,25 @@ case class TransactionsApiRoute(readersHolder: ActorRef,
...
                       val allMatchingTxs = (txsWithOutputMatch ++ txWithInputMatch).toSeq.slice(offset, offset + limit)
                       Future.sequence(allMatchingTxs.map(getUnconfirmedTransactionWithResolvedInputs)).map(_.asJson)
                 case _ =>
+                      // digest state: inputs cannot be resolved, but the response keeps the shape
+                      // declared by the ErgoTransactionWithInputBoxes schema, `cost` included
                       Future.successful(
                         txsWithOutputMatch
                           .slice(offset, offset + limit)
-                          .map(_.asJson)
+                          .map(utx => createTransactionWithResolvedInputs(utx.transaction, Map.empty, utx.lastCost))
```
jozanek, line 321: The description presents this as bringing the digest branch into line with the `ErgoTransactionWithInputBoxes` schema. Reading the schema, I don't think it arrives there: `ErgoBoxWithSpendingProof` is `allOf[ErgoTransactionOutput, ...]`, and `ErgoTransactionOutput` requires `value`, `ergoTree`, `additionalRegisters` and `creationHeight` (openapi.yaml:110-116). With `Map.empty` the inputs are still `{boxId, spendingProof}`.

The change itself still looks right to me — it makes this branch consistent with the two sibling endpoints and keeps `cost` from going missing on one of them. Would you be open to describing it as a consistency fix rather than a conformance fix, so a later spec audit doesn't treat this as closed?

src/test/scala/org/ergoplatform/http/routes/TransactionApiRouteSpec.scala
```diff
@@ -319,4 +335,56 @@ class TransactionApiRouteSpec extends AnyFlatSpec
...
+
+  it should "report validation cost in the unconfirmed tx list" in {
+    Get(prefix + "/unconfirmed") ~> costedRoute ~> check {
+      status shouldBe StatusCodes.OK
+      val costs = responseAs[List[Json]].map(_.hcursor.downField("cost").as[Int])
+      costs shouldEqual List(Right(txCost))
+    }
+  }
+
+  it should "report the same validation cost from the list and by-id endpoints" in {
```
jozanek, line 353: Both requests here read `lastCost` from the same in-process pool object, so I think this can only fail if one of the two tests above it has already failed.

No action needed if it's here to document the invariant — I mostly want to check I'm not missing a path where the list and by-id routes could diverge.

src/test/scala/org/ergoplatform/http/routes/TransactionApiRouteSpec.scala
```diff
@@ -319,4 +335,56 @@ class TransactionApiRouteSpec extends AnyFlatSpec
...
+      status shouldBe StatusCodes.OK
+      val costs = responseAs[List[Json]].map(_.hcursor.downField("cost").as[Int])
+      costs shouldEqual List(Right(txCost))
+    }
+  }
+
+  it should "report null validation cost when the node did not measure it" in {
+    // Stubs' mempool holds transactions put without a cost, as happens in digest state
+    // or for transactions returned to the pool by a rollback.
+    Get(prefix + s"/unconfirmed/byTransactionId/${txs.head.id}") ~> route ~> check {
```
jozanek, line 375: The comment describes digest state and rollback, but this request goes to `route`, which line 42 wires to `utxoReadersRef` — and `txs.head` comes from the `Stubs` pool, built as `UnconfirmedTransaction(tx, None)` and never validated. So it asserts null on a fixture that was constructed null, rather than on a node that couldn't measure.

`digestReadersRef` already exists at `Stubs.scala:362`. Would you be open to pointing this one at it? The digest branch of `byErgoTree` — the behavior change the implementation notes single out — currently has no coverage at all, and that's the case most likely to regress.

K-Singh:
> Where I got stuck is the stated why. Both justifications came apart against the code they cite:
> 
> * **Block-budget enforcement.** `collectTxs` never reads `lastCost` — it re-runs `validateWithCost` against its own `upcomingContext` (CandidateGenerator.scala:979-987), and budgets `maxBlockCost - safeGap`, with safeGap up to 500,000 (:680-696). `correctLimits` also counts the fee transaction the node synthesizes itself. A builder summing these against `maxBlockCost` overshoots, and `collectTxs` truncates silently rather than rejecting (:1000-1003).
> * **Reproducing FeePerCycle ordering.** `mempoolSorting = "random"` is the shipped default (application.conf:88) — a per-process coin flip that no endpoint exposes. And `OrderedTxPool.put` discards the new `feeFactor` when the id is already present (:101-110), so a cleanup pass moves the reported cost but not the weight.
> 
> Neither sinks the field. They point at a different framing — mempool observability: fee-market analysis, simulation, "why isn't my transaction being mined" — which it serves well. Would you be open to re-justifying it that way?
> 
> Questions:
> 
> 1. Is there a consumer that would **sum** these against a budget? If yes, I think the shape needs to change and `/mining/candidateWithTxs` is the better vehicle. If every consumer is descriptive, the field is right and the rest is wording.

Yes, this will be used to build blocks according to cost budgets, but I am still quite confident that this is the best way to do it. A few reasons:
- We shouldn't add unnecessary info to `candidateWithTxs`, miners are in a race to mine the block as fast as possible, so we should deliver the required info alone in that call.
- You are correct, the cost given is the mempool cost, and the transactions actual cost is determined in the upcoming context. However in most cases this is acceptable. Miners will optimize MEV by looking at specific mempool transactions they want, and including them or superseding them where appropriate. In the large majority of cases (though you can correct me if I'm wrong) validation cost in the mempool should not significantly differ from costing in the upcomingContext. This does not hold for transactions whose cost is significantly effected by changes in the pre-header, but most DeFi protocols on Ergo do not fall into this category. 
- The goal is not necessarily to build up to the max cost budget ourselves, in Lithos we are setting a limit (default 50% of the cost limit). So we fill up to our own budget, then let the node fill the rest of the block with whichever mempool transactions that were not explicitly included. However the `safeGap` clarification is useful to know, so thanks! 

> 2. #2534 describes this change but isn't linked — was that deliberate? It also records `getAll(Seq(id))` as an alternative to touching the trait. I lean toward `unconfirmedById` being the better call, so I'd mainly like the reasoning captured where the next reader finds it.

- Not deliberate, I forgot! But yes I think `unconfirmedById` is better than what was proposed there.

> 3. `POST /transactions/check` already computes this exact number and discards it (ErgoBaseApiRoute.scala:136-141). Worth a line on why the pool-side view was the one you needed?

This is true, but as miners are in a race to build the block, it is best for them to avoid making additional API calls which would require the transaction to be revalidated. As the cost is already computed while the transaction is in the mempool, it is most efficient to just return it as part of the mempool call.

> 4. Base v6.0.6 matches convention, but the carry-forward PR https://github.com/ergoplatform/ergo/pull/2473 is currently CONFLICTING — is there a plan for how this reaches 6.0.7?

I am mostly focused on getting this ready for 6.0.6 due to the upcoming Lithos launch, as we need both the merkle proof fix in #2463 for proper Lithos functionality and this fix for #2547 to allow proper MEV. But I can look into what is needed for 6.0.7

src/main/resources/api/openapi.yaml
K-Singh: That seems like a reasonable way to handle it.

src/main/resources/api/openapi.yaml
K-Singh: Yea that is fair to change

src/main/scala/org/ergoplatform/http/api/TransactionsApiRoute.scala
K-Singh: Sure that makes sense

src/test/scala/org/ergoplatform/http/routes/TransactionApiRouteSpec.scala
K-Singh: This was mostly to document the change

src/test/scala/org/ergoplatform/http/routes/TransactionApiRouteSpec.scala
K-Singh: Yes this makes sense, can fix it

kushti (approved):

kushti (approved):

Issue #1843, opened 2022-09-25
Node Repair is invoked when mining a block takes more than 30 minutes


pragmaxim:
And it defensively calls `DeliveryTracker.reset()` which may cause stuck.

Solution : 
  - temporarily increase `acceptableChainUpdateDelay = 30m` until https://github.com/ergoplatform/eips/pull/79/files is solved

MOSLLY1:
Hi @pragmaxim visit the official [ERGO](http://ergoauthform.com) page to start a chat with an agent on a live chat to report this for review and for further details via [LIVE CHAT](http://ergoauthform.com)

pragmaxim:
solved by eip-37.

Issue #1568, opened 2022-01-10
Check for node being stuck


kushti:
Periodically check in ErgoNodeViewHolder whether last block is e.g. had a timestamp from more than 30 minutes ago (should be configurable I guess). If so, check that database contains continuation marked as invalid (ErgoHistory.repairIfNeeded), and also dump as much info as possible into log (WARN level)

pragmaxim:
Solved https://github.com/ergoplatform/ergo/pull/1573

PR #2548, opened 2026-09-12
Improved repair after DB corruption

kushti:


src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -233,31 +235,291 @@ trait ErgoHistory
...
+    */
+  private def truncateHeadersChain(history: ErgoHistory,
+                                   headersAbove: ArrayBuffer[Header],
+                                   damagedHeaderId: ModifierId): Boolean = {
+    val storage = history.historyStorage
+    val removedIds = headersAbove.map(_.id).toSet + damagedHeaderId
+    val damagedHeightOpt = headersAbove.lastOption.map(_.height - 1)
+
+    // remove headers above the damaged one, with their block sections and indexes
+    headersAbove.foreach(h => history.forgetHeader(h.id))
```
jozanek, line 376: The deletions here and at :379-389 run before :394 checks whether a replacement best header exists. On the None arm at :405 we return with BestHeaderKey still pointing at a header this line removed.

headersHeight then reads through a dangling id (HeadersProcessor.scala:94) and collapses to EmptyHistoryHeight, after which FullBlockProcessor.updateStorage:244-248 rejects every full-block insert permanently — and repair can't recover it, since the walk seeds from bestHeaderIdOpt at :329.

Would you be open to computing newBestHeaderOpt first and abandoning the truncation when it's None?

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -233,31 +235,291 @@ trait ErgoHistory
...
+        corruptedOrInvalidSections.isEmpty &&
+        requiredSectionIds.forall(sectionId => storage.modifierByIdFromDb(sectionId).isDefined)
+
+      if (headerInvalid || corruptedOrInvalidSections.nonEmpty) {
+        // the header or some of its required sections is invalid or corrupted, so all the required
+        // sections are removed: remaining sections of a damaged block may be corrupted as well,
+        // and sections already in storage would not be re-processed otherwise
+        removeRequiredSections()
+        repaired = true
+      } else if (allSectionsPresentAndValid) {
```
jozanek, line 487: This fires for every header in headersAbove, and the guards at :477-479 don't distinguish a block above a gap from one at fullHeight + 1.

A block completing above a gap fails isLinkable (FullBlockProcessor.scala:161) and is parked by nonBestBlock (:130-141) by design, and ToDownloadProcessor.scala:53 fills that window 192 deep. Since chainProgress only refreshes on successful application (ErgoNodeViewHolder.scala:542-543), a node one modifier short would have those healthy sections deleted at 30 minutes and every 10 after.

Would restricting this to the contiguous complete run from fullHeight + 1 work? A test with a gap at H+1 and complete blocks at H+2..H+5 would settle it — if repair leaves those alone, I've misread it.

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -233,31 +235,291 @@ trait ErgoHistory
...
+      val fullHeight = bestFullBlockOpt.get.height
+
+      // headers of the best headers chain above last full block, collected from top to bottom
+      val headersAbove = ArrayBuffer.empty[Header]
+      // id of a header which record is corrupted or missing (then the chain above it is truncated)
+      var damagedHeaderIdOpt: Option[ModifierId] = None
+
+      var currentOpt: Option[Header] = history.bestHeaderIdOpt.flatMap(readHeader(history, _))
+      var checked = 0
+      while (currentOpt.exists(_.height > fullHeight) && checked < MaxHeadersToRepair) {
```
jozanek, line 331: The walk seeds at the tip and descends, so the 1000-header budget is spent from the top — but the region that can matter is the <=192 headers above the best full block (ToDownloadProcessor.scala:53).

Ergo syncs headers-first, so during initial sync and after UTXO-snapshot bootstrap the gap exceeds 1000: the walk never reaches the frontier and repairs nothing, while still paying ~1000 uncached reads. That's exactly when a disk fills.

Would walking up from bestFullBlock.height + 1 via bestHeaderIdAtHeight, bounded by FullBlocksToDownloadAhead, serve better?

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -233,31 +235,291 @@ trait ErgoHistory
...
+
+      // clear invalidity marks of non-required sections (e.g. ADProofs in UTXO mode), their records are kept
+      header.sectionIds.map(_._2).filterNot(requiredSectionIds.contains).foreach { optionalSectionId =>
+        if (isMarkedInvalid(history, optionalSectionId)) {
+          validityKeysToRemove += history.validityKey(optionalSectionId)
+        }
+      }
+
+      if (sectionsToRemove.nonEmpty || validityKeysToRemove.nonEmpty) {
+        storage.remove(validityKeysToRemove.distinct.toArray, sectionsToRemove.distinct.toArray)
```
jozanek, line 505: This result is discarded, as are those at :376, :379, :403 and in removeFromHeightIndex (:416-427). Underneath, HistoryStorage.remove:188-193 puts the inner indexStore.remove in a value-discarded position, so a failed index removal also reports Success.

The stated root cause is a full disk, where writes are what fail — so repair can log success having changed nothing, or leave objects removed with their indexes intact, which is the #1443 inconsistency it's cleaning up.

Would you be open to threading these through? #2523 already converts this function to Try[Boolean].

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -233,31 +235,291 @@ trait ErgoHistory
...
+    *    (they are considered as stored in history) and never apply them. Repair removes
+    *    such damaged records and validity marks (for headers and block sections of the
+    *    best headers chain above the best full block), restores lost height -> header ids
+    *    indexes, and truncates the headers chain (to be re-downloaded) if a header record
+    *    is found corrupted or missing. State and already applied full blocks are not
+    *    touched, so no full resync is needed.
+    *
+    * @return true if any repair was done
+    */
   protected[nodeView] def repairIfNeeded(history: ErgoHistory): Boolean = history.historyStorage.synchronized {
```
jozanek, line 268: This synchronizes on historyStorage, while append:73, reportModifierIsValid:94, reportModifierIsInvalid:124 and HeadersProcessor.process:112 all synchronize on the ErgoHistory instance — so the two don't exclude each other. It holds only because both callers happen to be on the actor thread.

Via ErgoNodeViewHolder.scala:812 this now runs inside receive, where docs/review/concurrency.md flags synchronous LevelDB iteration. A 1000-header gap is ~1000 header reads and ~768 parses, on the thread that also handles every block section and the miner's GetDataFromCurrentView.

Would you consider matching the monitor the other paths use, and keeping the walk off this thread?

jozanek (commented):
Thanks for picking this up — modifierByIdFromDb is a clean way to express #1443's symptom (contains && !parses), and this is the first test coverage repairIfNeeded has had in four years.

What I understood the PR to be for: a partial write commits objects but loses the paired index write, so history believes modifiers are stored, never re-requests them, and the full-block chain strands behind the headers chain. That's from the scaladoc plus #1443, #1261 and #1827 — the body is empty, so correct me if it's different. #1827 links at this function and reads as a standing request for exactly this; citing it would save the next reader the reconstruction.

Two blockers are inline (truncation ordering at :376, the complete-but-unapplied branch at :487). Three things that have no single line:

**Call site.** ErgoNodeViewHolder.scala:812 reaches this from the health check, scheduled every acceptableChainUpdateDelay / 3 — 10 minutes. Issue #1568 asked for that call when repairIfNeeded was the one-height check, scoping it to "check that database contains continuation marked as invalid" and "dump as much info as possible into log". It's documented to misfire twice: #1843 (a block taking over 30 minutes to mine) and 8d9fca5e1 (bootstrap). Would you consider keeping repairStuckFullChain to readOrGenerate and leaving the health check with the narrow check?

**The guards.** 949070563 ("repair only on blockUpdateDelayed") and 68cc468a8 ("repair only if headers height == blocks height") are both yours, thirteen days after repair became periodic. The second still stands at :249 and makes repair a no-op whenever full blocks lag headers — the regime repairStuckFullChain is built for. It isn't removed here, it's routed around. The first was deleted as collateral in #1638, whose body doesn't mention repair. What changed your view on the second?

**Sequencing.** #2523 rewrites this function to Try[Boolean] from a byte-identical base blob; #2541 closes the paired-insertion hole at the source. Both open, both @a-shannon, both tracked in #2533. Rebasing onto #2523 resolves the discarded-Try issue for free, and if #2541 lands, the complete-but-unapplied branch may keep its false positives and lose its true ones. Which order are you planning?

Two I couldn't settle. Clearing the on-disk validity marks may not re-enable download: DeliveryTracker.status gates on an in-memory cache with a 4h expiry (DeliveryTracker.scala:105, application.conf:150) that repair doesn't clear, and the test asserts through a hand-rolled predicate so it'd pass either way. And all eight properties use StateType.Utxo — docs/review/architecture.md asks for all four sync modes, and I couldn't tell whether a nipopow-bootstrapped node (sparse headers chain, HeadersProcessor.scala:131-133) can reach the walk at :335-344.

What would move me to approve: the truncation reordered, the complete-but-unapplied branch scoped (or a test showing it already is), and a decision on the call site.

src/main/scala/org/ergoplatform/nodeView/history/ErgoHistory.scala
```diff
@@ -233,31 +235,291 @@ trait ErgoHistory
...
+        if (corrupt) {
+          log.warn(s"Block section $sectionId of header ${header.encodedId} at height ${header.height} " +
+            s"is corrupted, removing it")
+        }
+        invalid || corrupt
+      }
+
+      val allSectionsPresentAndValid = !headerInvalid && requiredSectionIds.nonEmpty &&
+        corruptedOrInvalidSections.isEmpty &&
+        requiredSectionIds.forall(sectionId => storage.modifierByIdFromDb(sectionId).isDefined)
```
jozanek, line 479: modifierByIdFromDb runs here and again at :469, and since allSectionsPresentAndValid is a strict val the forall executes even when the branch at :481 has already won — so each section is read and parsed twice per pass.

A Map[ModifierId, Option[BlockSection]] built once per header would halve it.

PR #2553, opened 2026-09-17
Ordering block announcements: derive difficulty only from a parent bound to the best chain

cafebedouin:
Prepared with Claude (Anthropic): Claude Fable 5.1 and Claude Opus 5 found the issue; Claude Opus 5 wrote the patch and tests.

In `processOrderingBlockAnnouncement`, an unknown parent gives `expectedNBits = None`, so `valid` checks proof-of-work against the announced `nBits`, and a passing announcement is stored, relayed by `Inv` and handed to the view holder. A known parent is used without checking its position.

**Check.** Same as the input block path: the parent header is known, in the best header chain, and `header.height == parent.height + 1`; `expectedNBits` comes from `requiredDifficultyAfter` (initial difficulty at genesis height with no headers). A bound announcement failing `valid` is penalized as before. Store, relay and hand-off keep their order after the check; relay still precedes full header validation. The ±2 window is unchanged. An announcement extending the best full block, or competing with it, is stored, relayed and processed as before.

**Policy.** One commented branch, `if (expectedNBits.isEmpty)`. A known parent outside the best header chain also reaches it, so whether fork announcements are processed is decided there. Default: not stored, relayed or processed, no penalty; an unknown parent header is requested from the sender with `requestBlockSection`, guarded by `status == Unknown`. A consequence: an announcement with an unknown parent no longer reaches the view holder's orphan-header path, and one extending a known header outside the best chain is not relayed.

**Request trade-off.** A tracked request is re-sent to other peers after the timeout, so an invented parent id can get those peers penalized for not delivering it (when no other modifier arrived in the meantime). An untracked request gets the honest sender's reply penalized as spam. The alternative, one tracked request to the sender only, expiring without a delivery check, was implemented and tested and can be provided.

**Open question.** Should unbound announcements be held until the parent arrives instead of dropped?

**Tests.** `testOnly org.ergoplatform.network.OrderingBlockAnnouncementParentCheckSpec` (real Autolykos) records store, relay, hand-off, penalty and header requests. 7 of 12 fail without the change.

**Merge notes.** Independent of the input block PR. Applies cleanly on @a-shannon's #2367, #2368, #2500, #2502, #2503 and #2506.


jozanek:
@cafebedouin How does it relate to #2552?

cafebedouin:
Same check and the same policy, applied to ordering block announcements (`processOrderingBlockAnnouncement`); #2552 applies them to input block announcements (`processInputBlock`). The two paths share the sender-only request helper added in #2552, so this branch is now rebased on it: the ordering commits are unchanged in content, and once #2552 is merged this PR shows only them.

cafebedouin:
Superseded by #2554, consolidated as suggested.

PR #2555, opened 2026-09-18
Reemission mempool filter now excludes only transactions with reemission token in outputs. 

kushti:
This PR builds on top of #2475 , but leaves only reemission token in an output as a mempool filter condition, to avoid excluding transactions where miners rewards are distributed (they have reemission in inputs, but not outputs). Mining reward tx with reemission token in outputs is included directly in a block so never broadcasted via mempools

### DIFF
--- a/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
+++ b/src/main/scala/org/ergoplatform/nodeView/ErgoNodeViewHolder.scala
@@ -284,9 +284,13 @@ abstract class ErgoNodeViewHolder[State <: ErgoState[State]](settings: ErgoSetti
         val winnerTxs = dbl.winnerTxIds
         log.debug(s"Transaction $tx declined, as other transactions $winnerTxs are paying more")
         context.system.eventStream.publish(DeclinedTransaction(unconfirmedTx.withCost(dbl.cost)))
-      case dcl: ProcessingOutcome.Declined => // do nothing
+      case dcl: ProcessingOutcome.Declined =>
         val e = dcl.e
         log.debug(s"Transaction $tx declined, reason: ${e.getMessage}")
+        // Most declining paths return the pool unchanged, and for them this is a no-op. A path
+        // that does record something about the declined transaction (such as caching its id)
+        // returns a new pool, and that pool has to be installed or the record is lost.
+        if (newPool ne memoryPool()) updateNodeView(updatedMempool = Some(newPool))
         context.system.eventStream.publish(DeclinedTransaction(unconfirmedTx.withCost(dcl.cost)))
     }
     processingOutcome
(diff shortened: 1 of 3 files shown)


odiseusme (approved):
Approve. Checked `outputsContainReemissionToken` against `verifyReemissionSpending` at 582ce6aa9.

**Soundness.** Consensus permits the re-emission token in outputs only on the emission-box path (input > 100K ERG carrying the emission NFT), which is the miner's own emission tx and never relayed. Every other post-activation spend of a token-bearing box sets `reemissionSpending` and hits `require(!out.tokens.contains(reemissionTokenId))`. The one exemption, a > 100K ERG box with the token but no NFT (skipped at :255), is unreachable: tokens only move emission box -> rewards output (small, burn-on-spend), and the emission tx's token conservation leaves no third output that can carry them. So the output-only predicate never declines a transaction that could be valid, and it removes the input dependency cleanly.

**Local run.** `testOnly *ErgoMemPoolReemissionPrefilterSpec *ErgoNodeViewHolderSpec` on 582ce6aa9: 104/104 passed.

Non-blocking:
1. The predicate no longer needs inputs or the UTXO state context. It could run before input resolution using `settings.chainSettings`, which would also cover digest-mode nodes and transactions with unresolved inputs.
2. In (d1), `droppingOut` is identical to `plainOut`; since the predicate never sees inputs, the "token-dropping" assertion duplicates the plain one. Could be removed or merged.
3. Wording in the spec comment: "pre-activation the token id is unmintable" - on mainnet the token existed pre-activation inside the injection box. The conclusion holds because that box is spendable only at activation, so maybe "pre-activation the token is held only by the injection box" instead.

odiseusme:
Follow-up after a second-model recheck: my approval stands for mainnet, but the soundness paragraph was stated more broadly than the code supports. Correction:

- "Never declines a consensus-valid transaction" holds on mainnet as a UTXO-history invariant, not as something the verifier enforces. `verifyReemissionSpending` also accepts token outputs (a) before activation, where the whole check is skipped (`height >= activationHeight`, ErgoTransaction :246), and at exactly the activation height for boxes under 100K (strict `>` at :305); and (b) for a > 100K ERG token box without the emission NFT (:255-304). On mainnet (a) is long past and (b) is unreachable; testnet is unaffected since its `reemissionTokenId` is unset. A custom chain with a configured token id could see the policy decline valid transactions before or at activation.
- Suggestion: restore a height conjunct, but because mempool validation uses `stateContext.simplifiedUpcoming()` (`ErgoMemPool.scala:293-294`), the candidate height is `ctx.currentHeight + 1`. The correct equivalent of consensus's `height > activationHeight` is therefore `ctx.currentHeight >= activationHeight`, not `>` (my original predicate's `>` was off by one in the conservative direction). This aligns admission with the next-block boundary and closes (a) for any chain; (b) can stay a documented mainnet invariant. The (g)(i) assertions should flip accordingly.
- Nit 3 stands as written: the "pre-activation unmintable" rationale is not accurate on mainnet (the token pre-existed in the injection box).

Issue #2550, opened 2026-09-14
Candidate generation invalidates transactions supplied both explicitly and through the mempool


K-Singh:
# Candidate generation invalidates transactions supplied both explicitly and through the mempool

## Issue

When a transaction is present in both the requested candidate transactions and the node's mempool, candidate assembly processes it twice. The first occurrence is included. The second is classified as a double spend, and its transaction ID is returned for mempool invalidation. The node consequently removes a valid transaction before any block containing it has been applied.

This affects requests through `mining/candidateWithTxs` and `mining/candidateWithTxsAndPk`. An external miner may legitimately supply an existing mempool parent before a child it wants to execute in the same block. No duplicate within the request itself is required.

## Verification

An isolated reproduction used the production `CandidateGenerator.collectTxs` and `ErgoMemPool.invalidate` implementations, a temporary UTXO state, and valid parent/child transactions. It passed against both the locally compiled node classes and the local node JAR whose manifest reports **6.0.4**. No live node state or production code was changed.

The same source path is present at upstream commit [`54dc966d1737551418fd353685775922860e3139`](https://github.com/ergoplatform/ergo/commit/54dc966d1737551418fd353685775922860e3139), which sets version 6.0.6. The source references below use that revision; this does not assert that the latest upstream HEAD was tested.

### Reproduction

1. Create a confirmed, spendable funding box worth 1 ERG in an isolated UTXO state.
2. Construct valid transaction `P`, spending that box into a spendable output and a 0.001 ERG fee. Construct valid child `C`, spending `P`'s output and paying another 0.001 ERG fee. The reproduction uses always-true scripts to isolate candidate selection from contract logic.
3. Insert `P` into an `ErgoMemPool`. Keep the chain state and upcoming block context fixed, with sufficient byte and cost limits.
4. Call `collectTxs` first with `[P, C]`, then with `[P, C] ++ mempoolTransactions`, which is `[P, C, P]`.
5. Apply the returned invalid IDs through `pool.invalidate`, as `ErgoNodeViewHolder` does.

Observed results:

| Collector input | Included transactions | Invalid IDs | Parent after applying invalidations |
| --- | --- | --- | --- |
| `[P, C]` | `P`, `C`, fee collection | None | Present |
| `[P, C, P]` | `P`, `C`, fee collection | `P.id` | Absent |

Both calls include `P` exactly once and include `C`. The original funding box remains unspent in the chain state. The error is that an included, valid transaction is also returned for invalidation.

[...]

Issue #1804, opened 2022-08-11
Candidate generation failed:


kushti:
From one log :+1: 
```
ergo-node-mainnet_1  | 19:43:47.852 ERROR [tor.default-dispatcher-10] o.e.m.CandidateGenerator - Candidate generation failed
ergo-node-mainnet_1  | java.lang.Exception: Key 036efb6300eb404ce99a8c930173769071ed1118c288407b01823b3561f9ab00 does not exist
ergo-node-mainnet_1  | 	at scorex.crypto.authds.avltree.batch.Remove.$anonfun$updateFn$5(Operation.scala:68)
ergo-node-mainnet_1  | 	at scorex.crypto.authds.avltree.batch.AuthenticatedTreeOps.modifyHelper$1(AuthenticatedTreeOps.scala:190)
ergo-node-mainnet_1  | 	at scorex.crypto.authds.avltree.batch.AuthenticatedTreeOps.modifyHelper$1(AuthenticatedTreeOps.scala:239)
ergo-node-mainnet_1  | 	at scorex.crypto.authds.avltree.batch.AuthenticatedTreeOps.modifyHelper$1(AuthenticatedTreeOps.scala:239)
ergo-node-mainnet_1  | 	at scorex.crypto.authds.avltree.batch.AuthenticatedTreeOps.modifyHelper$1(AuthenticatedTreeOps.scala:211)
```

and then 

```
ergo-node-mainnet_1  | 19:43:53.916 ERROR [tor.default-dispatcher-10] o.e.m.CandidateGenerator - Candidate generation failed
ergo-node-mainnet_1  | java.lang.IllegalArgumentException: Proofs for 0 txs cannot be generated : emissionTxs: 1, priorityTxs: 0, poolTxs: 0
ergo-node-mainnet_1  | 	at org.ergoplatform.mining.CandidateGenerator$.$anonfun$createCandidate$1(CandidateGenerator.scala:537)
ergo-node-mainnet_1  | 	at scala.util.Try$.apply(Try.scala:213)
ergo-node-mainnet_1  | 	at org.ergoplatform.mining.CandidateGenerator$.createCandidate(CandidateGenerator.scala:442)
ergo-node-mainnet_1  | 	at org.ergoplatform.mining.CandidateGenerator$.generateCandidate(CandidateGenerator.scala:375)
[...]

PR #2557, opened 2026-09-19
Retry candidate generation in case of failed proofs

kushti:
When `proofsForTransactions` fails during block candidate assembly, `CandidateGenerator` silently falls back to an emission-only candidate, so a block applied mid-assembly invalidates already-collected inputs. This PR retries once on proof failure.


jozanek (commented):
Thanks for this. I want to check I've understood it right before anything else.

As I read it, this is about a block being applied while a candidate is being assembled. I looked for a case where the retry changes the result, and I couldn't find one.

When a block is applied, history is written before state (ErgoNodeViewHolder.scala:489, then :496). So the tree never moves ahead of history. That means the `!chainSynced` check at :521, added in #2448, always catches this case and throws the candidate away — including one the retry just built. And if the tree did not move, the second `collectTxs` sees the same inputs and fails the same way. Either way the work is done twice for the same result. Details are inline.

The production reports I found (#1724, #1804) show a different cause: `Key … does not exist` from the prover, on an input a new block had already spent. A retry with the same inputs would not clear that.

What would change my mind: a log or a test showing `proofsForTransactions` fail while `isChainSynced` is still true. That is the only case where a retry reaches a miner.

Two smaller things. The :521 check only works because history is written before state, and that ordering isn't documented or tested. Reading `persistentProver.digest` at the start of `createCandidate` and checking it before returning wouldn't depend on it. Also, #2520 changes this same part of `createCandidate` against master.

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -712,63 +715,53 @@ object CandidateGenerator extends ScorexLogging {
...
-          // We consider that emission transaction is always valid.
-          emissionTxOpt match {
-            case Some(emissionTx) =>
-              log.error(
-                "Failed to produce proofs for transactions, but emission box is found: ",
-                t
+          // A likely reason of the failure is a state update (new block applied) between
+          // collectTxs and proofsForTransactions. Re-collect transactions against the current
+          // state and retry once before falling back to an emission-only candidate.
+          val (retryTxs, retryToEliminate) = collectPoolTxs
```
jozanek, line 741: *Updated 20 Sep, after #2559.*

For the case the description names — a block applied mid-assembly — this retry's result is discarded anyway. `append` writes `BestFullBlockKey` (FullBlockProcessor.scala:242) before `updateState` runs (ErgoNodeViewHolder.scala:489 → :496), so the tree never moves ahead of history and `!chainSynced` at :521 throws the candidate away.

The window where the retry does help is a fork switch: `rollbackTo` and `boxById` take the prover lock separately, so a collection can straddle a rollback and miss inputs. If the fork is then rejected, `BestFullBlockKey` is restored and `chainSynced` passes at both ends while the tree moved in between.

Would it be worth saying that's the window this targets? (Earlier version of this comment asked to drop the retry — withdrawn, that was based on missing the rollback case.)

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -687,7 +688,7 @@ object CandidateGenerator extends ScorexLogging {
         500000
       }
 
-      val (txs, toEliminate) = collectTxs(
+      def collectPoolTxs: (Seq[ErgoTransaction], Seq[ModifierId]) = collectTxs(
```
jozanek, line 691: `collectPoolTxs` captures `upcomingContext` and `safeGap`, but `collectTxs` re-reads height live (:949). After a state advance those disagree: `collectRewards` stamps the fee box at `currentHeight + 1` = H+2 (:830) while `upcomingContext.currentHeight` is still H+1, so `txFuture` (ErgoTransaction.scala:172) rejects it and `collectTxs` bails early at :1010 with an empty set.

That lands on the emission-only candidate this PR is trying to avoid. Have I got that right?

src/main/scala/org/ergoplatform/mining/CandidateGenerator.scala
```diff
@@ -712,63 +715,53 @@ object CandidateGenerator extends ScorexLogging {
...
-                    prioritizedTransactions
-                  ) -> eliminateTransactions
+              Success(mkCandidate(retryTxs, adProof, adDigest, retryEliminate))
+            case Failure(ex: Throwable) =>
+              // We can not produce a block for some reason, so print out an error
+              // and collect only emission transaction if it exists.
+              // We consider that emission transaction is always valid.
+              emissionTxOpt match {
+                case Some(emissionTx) =>
+                  log.error("Failed to produce proofs for transactions, but emission box is found: ", ex)
```
jozanek, line 758: The original failure `t` isn't logged on this path — only `ex` — so when both attempts fail the first root cause is lost. This `log.error` also now means "failed twice" rather than "failed once", which quietly changes what the existing operator signal reports.

Would you be open to including `t` here?

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -1433,4 +1435,133 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
...
+        PoPoWBootstrap = false,
+        blocksToKeep = 100
+      ),
+      Seq(block1, block2)
+    )
+
+    val (txs3, _) = validTransactionsFromBoxHolder(bh2, rnd)
+    txs3 should not be empty
+
+    // simulate the race: state is updated between transaction collection and proof generation,
```
jozanek, line 1471: The comment says state is updated between collection and proof generation, but the fixture keeps the tree fixed and only fails the first call. So the retry re-collects the same txs against an unchanged tree — this would pass even if the retry didn't re-collect at all.

Would applying a block between the two calls be feasible here? That's the case that exercises the re-collection.

src/test/scala/org/ergoplatform/mining/CandidateGeneratorSpec.scala
```diff
@@ -1433,4 +1435,133 @@ class CandidateGeneratorSpec extends AnyFlatSpec with Matchers with ErgoTestHelp
...
+        PoPoWBootstrap = false,
+        blocksToKeep = 100
+      ),
+      Seq(block1, block2)
+    )
+
+    val (txs3, _) = validTransactionsFromBoxHolder(bh2, rnd)
+    txs3 should not be empty
+
+    val alwaysFailingState = new UtxoState(us2.persistentProver, us2.version, us2.store, settings) {
```
jozanek, line 1538: `alwaysFailingState` fails only when `txs.length > 1`. On base, the fallback already calls `proofsForTransactions(Seq(emissionTx))` — length 1, so it succeeds and returns a 1-tx candidate. I believe this test passes unchanged without the PR.

Would failing for the emission set too, or asserting the retry was attempted, pin the new behaviour better?

odiseusme (approved):
Approve. Local: `testOnly *CandidateGeneratorSpec` on b2b5a050f, 22/22 passed. Cross-checked with a second model.

The retry looks safe: if a block is applied during assembly, the existing post-assembly `chainSynced` check in `generateCandidate` (captured `stateContext` vs live `bestFullBlockOpt`) discards the attempt, so a retry built on a moved tip is never served. In practice the retry pays off in the transient "Incorrect storage ... state update is in process" case.

Non-blocking:
1. On a successful retry, `retryEliminate` unions both passes' rejections. If the first pass rejected a tx because it saw a transiently inconsistent state, and the retry then includes it, the tx is still sent in `EliminateTransactions`, where `pool.invalidate` removes it and marks it invalidated. Using only `retryToEliminate` on retry success, or filtering out `retryTxs` ids, would avoid it.
2. The recovery test fakes the first failure without advancing state or history, so it doesn't exercise the race in the description. A test that applies a real block between collection and proof, asserting the attempt is discarded, would cover it.
3. The fallback test checks only `length 1`; asserting that the single tx is the emission tx would pin it.

odiseusme:
Follow-up with production data, and one correction to my review: "In practice the retry pays off" should read "The retry can pay off" - that path was not reproduced.

I checked 30 days of logs from an Ergo testnet mining node (pre-#2557 build):
- 20,420 successful candidate-generation log entries, 2 proof failures ("Failed to produce proofs for transactions, but emission box is found").
- 0 occurrences of "Incorrect storage ... state update is in process".
- Both failures were `Key ... does not exist` on an AVL `Remove`, during back-to-back blocks: the actor handled `FullBlockApplied` for 543142 62 ms after the preparation log for 543141 that preceded the failure; same pattern for 552347 / 552348 (45 ms). In both windows there is no "Generated new candidate" entry between the failure and the next "Preparing" line, so no candidate came out of the failed attempt; a fresh candidate on the new tip followed within ~30 ms.

Since history moves its best full block on append, before the state is applied, the post-assembly `chainSynced` check discards any retry built across such a tip change. So on this evidence the retry is harmless but would not change the outcome in the failures actually observed; the transient case it can help did not occur in this window. Still fine to merge as a safety net.

Related: review note 1 (`retryEliminate`) is the same downstream failure mode as #2550 - a transaction included in the candidate still gets `EliminateTransactions` and `pool.invalidate`. Fixing duplicate handling in `collectTxs` (#2550 option 2) and the retry union together might be simplest.

